PR #4, opened 2025-02-08
feat: replace pytesseract

Goldziher:
This PR replaces pytesseract with "inhouse" 

It also updates testing to hit 100% test coverage


kreuzberg/_string.py
```diff
@@ -4,6 +4,8 @@
 
 from charset_normalizer import detect
 
+from kreuzberg import ParsingError
```
Goldziher:
```suggestion
from kreuzberg.exceptions import ParsingError
```

Issue #39, opened 2025-04-10
Error when working with EasyOCR


mlinmg:
Hi, 
Awesome project, however when working with EasyOCR,  the conversion to bytes make the image unusable for easyocr and during         img = cv2.imdecode(nparr, cv2.IMREAD_COLOR) it leads to a null imager and throws error. 
an MRE is:
```pytthon
from kreuzberg import extract_file, ExtractionConfig

await extract_file(
                file_or_path_of_pdf_file,
                config=ExtractionConfig(
                    extract_tables=True,
                    ocr_backend='easyocr'
                ))).content
```
My env is 
```python
easyocr==1.7.2
kreuzberg==3.1.2
```
II'll open a small PR to fix this shortly

Goldziher:
Thanks! Can you  you suggest a reproduction?

mlinmg:
For me this mre 
```python
from kreuzberg import extract_file, ExtractionConfig

await extract_file(
                file_or_path_of_pdf_file,
                config=ExtractionConfig(
                    extract_tables=True,
                    ocr_backend='easyocr'
                ))).content
``` 
with any pdf that needs ocr yields an error

Goldziher:
Great work, thanks! I am releasing a patch version now

PR #40, opened 2025-04-10
fix: easyocr byte string issues

mlinmg:
Fixed image passing to the easyocr backend by using an nparray instead of bytes


mlinmg:
fixes  #39

kreuzberg/_ocr/_easyocr.py
```diff
@@ -180,7 +181,7 @@ async def process_image(self, image: Image.Image, **kwargs: Unpack[EasyOCRConfig
         try:
             result = await run_sync(
                 self._reader.readtext,
-                image.tobytes(),
+                np.array(image),
```
Goldziher, line 185: I see, please import numpy locally inside the function - reason is, its a dependency of easyocr but not Kreuzberg.

PR #42, opened 2025-04-10
Fix validating pandoc version on Windows

arnoudplantinga:
When running pandoc --version on windows, the result is something like ` pandoc.exe 3.6.4`. Currently, the version check does not take into account the ".exe".


Goldziher:
thanks, but i think this is no longer requireed.

PR #67, opened 2025-06-11
feat(backend): Allow enabling GPU acceleration for OCR

naderalexan:


kreuzberg/_ocr/_paddleocr.py
```diff
@@ -96,10 +97,95 @@ class PaddleOCRConfig:
     """Whether to recognize spaces."""
     use_zero_copy_run: bool = False
     """Whether to enable zero_copy_run for inference optimization."""
-
-
-class PaddleBackend(OCRBackend[PaddleOCRConfig]):
-    _paddle_ocr: ClassVar[Any] = None
+    gpu_memory_limit: float | None = None
```
Goldziher, line 102: if we are doing this, we should define "__slots__", read up on this - it make instantiation faster. slots should be defined on the base class as well (OCRBackend) to be inherited correctly. 

You use slots when you defined class attributes that are dynamically initialized in the constructor.

kreuzberg/_ocr/_paddleocr.py
```diff
@@ -96,10 +97,95 @@ class PaddleOCRConfig:
...
+                use_space_char=self.config.use_space_char,
+                drop_score=self.config.drop_score,
+                use_mp=self.config.use_mp,
+                total_process_num=self.config.total_process_num,
+                enable_mkldnn=self.config.enable_mkldnn,
+                use_zero_copy_run=self.config.use_zero_copy_run,
+                device=self._device
+            )
+            
+    def _process_image(self, image: Image.Image) -> str:
```
Goldziher, line 195: when is this called?

kreuzberg/_ocr/_easyocr.py
```diff
@@ -153,10 +154,78 @@ class EasyOCRConfig:
...
+            except ImportError as e:
+                raise MissingDependencyError("easyocr", e) from e
+                
+            self._reader = easyocr.Reader(
+                self.config.language,
+                gpu=self.config.use_gpu,
+                device=self._device
+            )
+            
+    def _process_image(self, image: Image.Image) -> str:
```
Goldziher: where is this used?

kreuzberg/_ocr/_easyocr.py
```diff
@@ -153,10 +154,78 @@ class EasyOCRConfig:
...
+    """Device to use for computation. Can be 'cpu', 'cuda', 'mps', or 'auto'."""
 
 
 class EasyOCRBackend(OCRBackend[EasyOCRConfig]):
-    _reader: ClassVar[Any] = None
+    """EasyOCR backend implementation."""
+    
+    def __init__(self, config: EasyOCRConfig) -> None:
+        self.config = config
+        self._reader = None
```
Goldziher, line 171: this is now untyped, it should be properly typed, same for paddle ocr. 

Please install pre-commit (pre-commit install) and run it (pre-commit run --all-files)

kreuzberg/_ocr/_paddleocr.py
naderalexan: my bad, this is old code that is not needed, removed it

kreuzberg/_chunker.py
```diff
@@ -36,7 +38,7 @@ def get_chunker(
     if key not in _chunkers:
         try:
             if mime_type == MARKDOWN_MIME_TYPE:
-                from semantic_text_splitter import MarkdownSplitter
+                from semantic_text_splitter import MarkdownSplitter  # type: ignore[import-not-found]
```
Goldziher, line 41: ok, the issue you had and why you needed t do all this changes is that you did not install the extras.

Look under pyproject.toml to understand how optional dependencies work. Read in the uv sync docs about how to install all extras. 

Most of the typing changes you did are unnecessary, you simply didnt have the extras in place.

kreuzberg/_ocr/_paddleocr.py
naderalexan: I implemented it, can you check if it is correct now?

kreuzberg/_chunker.py
naderalexan: Should I revert this commit then?

PR #69, opened 2025-06-13
chore(docs): fix contribution guidelines and pyproject.toml

naderalexan:


docs/contributing.md
```diff
@@ -19,53 +19,26 @@ Thank you for considering contributing to Kreuzberg! This document provides guid
...
     ```
 
-## Running Tests
+    Then install dependencies:
 
-```bash
-pytest
-```
+    ```bash
+    uv pip install -e ".[all,dev,docs,extras]"
```
Goldziher: uv sync --all-extras --dev

pyproject.toml
```diff
@@ -49,35 +49,17 @@ dependencies = [
...
-  "paddleocr>=3.0.1",
-  "paddlepaddle>=3.0.0",
-  "setuptools>=80.9.0",
-]
-urls.homepage = "https://github.com/Goldziher/kreuzberg"
-
-[dependency-groups]
-dev = [
+optional-dependencies.dev = [
   "covdefaults>=2.3.0",
```
Goldziher: these are not optional- its dev deps.

docs/contributing.md
naderalexan: doing so leads mypy to complain, can you try it out and recommend how to best do this?

PR #76, opened 2025-07-02
feat(backend): add entity and keyword extraction capabilities

naderalexan:
@Goldziher I still need to update the documentation to explain how to use entity and keyword extraction but I first want you to review the code incase there are changes that would affect the documentation at the end


kreuzberg/_types.py
```diff
@@ -149,12 +167,26 @@ class ExtractionConfig:
     """Configuration to pass to the OCR backend."""
     gmft_config: GMFTConfig | None = None
     """GMFT configuration."""
-    post_processing_hooks: list[PostProcessingHook] | None = None
+    post_processing_hooks: tuple[PostProcessingHook, ...] | None = None
```
Goldziher: Why this change?

kreuzberg/_types.py
```diff
@@ -149,12 +167,26 @@ class ExtractionConfig:
...
     """Post processing hooks to call after processing is done and before the final result is returned."""
-    validators: list[ValidationHook] | None = None
+    validators: tuple[ValidationHook, ...] | None = None
```
Goldziher: Sane

kreuzberg/extraction.py
```diff
@@ -34,6 +37,26 @@ async def _validate_and_post_process_async(result: ExtractionResult, config: Ext
             content=result.content,
         )
 
+    if config.extract_entities:
```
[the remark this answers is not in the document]
Goldziher: I agree with copilot

kreuzberg/_entity_extraction.py
```diff
@@ -0,0 +1,91 @@
...
+    from gliner import GLiNER
+except ImportError:
+    GLiNER = None
+
+try:
+    from keybert import KeyBERT
+except ImportError:
+    KeyBERT = None
+
+logger = logging.getLogger(__name__)
```
Goldziher: We don't use logger elsewhere if iirc

kreuzberg/_entity_extraction.py
```diff
@@ -0,0 +1,91 @@
...
+
+try:
+    from gliner import GLiNER
+except ImportError:
+    GLiNER = None
+
+try:
+    from keybert import KeyBERT
+except ImportError:
+    KeyBERT = None
```
Goldziher: If these are optional dependencies, we should use the same pattern we use elsewhere. Use a class, rather than a function, do validation.

kreuzberg/extraction.py
```diff
@@ -34,6 +37,26 @@ async def _validate_and_post_process_async(result: ExtractionResult, config: Ext
...
+            logger.warning("Entity extraction failed: %s", e)
+            result.entities = None
+
+    if config.extract_keywords:
+        try:
+            result.keywords = extract_keywords(
+                result.content,
+                keyword_count=config.keyword_count,
+            )
+        except RuntimeError as e:
```
Goldziher: Why runtime error?

naderalexan: Because it will catch most potential exceptions coming out of KeyBERT or GliNER without using a generic Exception. 
What would be your recommendation here?

kreuzberg/_types.py
naderalexan: That's not needed, was done by gemini-cli while fixing another hashing issue elsewhere and I forgot to revert it, will revert it now

kreuzberg/_types.py
naderalexan: That's not needed, was done by gemini-cli while fixing another hashing issue elsewhere and I forgot to revert it, will revert it now

kreuzberg/_entity_extraction.py
naderalexan: That's true, it is not used in the rest of the code, but shouldn't we use a logger?

kreuzberg/_entity_extraction.py
naderalexan: I can't find in the rest of the code where we do this, but I did convert it to a class -- not sure if my solution is what you are looking for so I wont resolve this comments until you confirm that the change I made is what you mean

kreuzberg/_entity_extraction.py
Goldziher: [blocker] perhaps we should, but its a seperate issue id say. Feel free to open a ticket. This is out of scope for now please.

kreuzberg/_entity_extraction.py
Goldziher: See these as an example: 

kreuzberg/_ocr/_paddleocr.py
kreuzberg/_ocr/_tesseract.py

These are classes.

You dont have to use them. Functions are fine as well, but the imports should happen inside the functions - and they should raise validation errors. The reason is that the code will run on import.

You can abstract this to a helper. 

Another possible pattern is this:

kreuzberg/_api/main.py

See the try/except block there.

kreuzberg/extraction.py
Goldziher: i would check what these libraries raise in the potential code path and use the correct exceptions, or use Exception.

kreuzberg/_entity_extraction.py
```diff
@@ -75,29 +56,36 @@ def extract_entities(
...
     """
-    if KeyBERT is None:
-        return []
+    kw_model = model
+    if not kw_model:
+        try:
+            from keybert import KeyBERT
+
+            kw_model = KeyBERT()
+        except (ImportError, RuntimeError):
```
Goldziher: see how we use MissingDependencyError.create_for_package -> this is what should be raised if someone tries to use this feature without the dependencies.

Goldziher (approved):

PR #88, opened 2025-07-12
feat: automatic document type detection

naderalexan:


kreuzberg/extraction.py
```diff
@@ -58,30 +60,54 @@ def _validate_and_post_process_helper(result: ExtractionResult, config: Extracti
...
+        if config.document_classification_mode == "vision" and file_path:
+            import anyio
+
+            from kreuzberg._document_classification import classify_document_from_layout
+            from kreuzberg._ocr import get_ocr_backend
+
+            async def _get_layout() -> ExtractionResult:
+                return await get_ocr_backend("tesseract").process_file(file_path, **config.get_config_dict())
+
+            layout_result = anyio.run(_get_layout)
```
naderalexan: [blocking] @Goldziher is this okay or should we convert the whole function to async?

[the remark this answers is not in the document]
Goldziher: thiis is indeed a problem. we have a fully async and a fully sync path, both need to remain pure

Goldziher: it not ok.  this is indeed blocking. we have fully sync and fully async paths. We cannot mix them like this, or performance is screwed.

naderalexan: @Goldziher What if I isolate the async operation in a separate process?

Goldziher: So, we can't use a different process..

Sync already does multi threading.

kreuzberg/_document_classification.py
```diff
@@ -0,0 +1,114 @@
+from __future__ import annotations
+
+import re
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from kreuzberg._types import ExtractionConfig, ExtractionResult
+
+
+DOCUMENT_CLASSIFIERS = {
```
Goldziher, line 15: This is a pretty rudimentary approach to the problem, that will work only with english docs

kreuzberg/_types.py
```diff
@@ -135,10 +135,18 @@ class ExtractionResult:
     """Extracted keywords and their scores, if keyword extraction is enabled."""
     detected_languages: list[str] | None = None
     """Languages detected in the extracted content, if language detection is enabled."""
+    document_type: str | None = None
+    """Detected document type, if document type detection is enabled."""
+    type_confidence: float | None = None
```
Goldziher: Should be document_type_confidence

kreuzberg/_types.py
```diff
@@ -135,10 +135,18 @@ class ExtractionResult:
...
+    """Confidence of the detected document type."""
+    layout: DataFrame | None = field(default=None, repr=False, hash=False)
+    """Internal layout data from OCR, not for public use."""
 
     def to_dict(self) -> dict[str, Any]:
         """Converts the ExtractionResult to a dictionary."""
-        return asdict(self)
+        data = asdict(self)
+        data.pop("layout", None)  # Exclude layout from the dictionary representation
```
Goldziher: Why?

kreuzberg/_types.py
```diff
@@ -193,6 +201,12 @@ class ExtractionConfig:
     """Configuration for language detection. If None, uses default settings."""
     spacy_entity_extraction_config: SpacyEntityExtractionConfig | None = None
     """Configuration for spaCy entity extraction. If None, uses default settings."""
+    auto_detect_document_type: bool = False
+    """Whether to automatically detect the document type."""
+    type_confidence_threshold: float = 0.7
```
Goldziher: document_type_confidence_threshold

kreuzberg/extraction.py
```diff
@@ -58,30 +60,54 @@ def _validate_and_post_process_helper(result: ExtractionResult, config: Extracti
             config=config.language_detection_config,
         )
 
+    if config.auto_detect_document_type:
+        if config.document_classification_mode == "vision" and file_path:
+            import anyio
```
Goldziher: Should be top level import

kreuzberg/extraction.py
```diff
@@ -58,30 +60,54 @@ def _validate_and_post_process_helper(result: ExtractionResult, config: Extracti
             config=config.language_detection_config,
         )
 
+    if config.auto_detect_document_type:
+        if config.document_classification_mode == "vision" and file_path:
```
Goldziher: Extract the logic here and below to functions please

kreuzberg/extraction.py
```diff
@@ -58,30 +60,54 @@ def _validate_and_post_process_helper(result: ExtractionResult, config: Extracti
...
     for validator in config.validators or []:
         run_sync_only(validator, result)
 
-    result = _validate_and_post_process_helper(result, config)
+    result = _validate_and_post_process_helper(result, config, file_path)
 
     for post_processor in config.post_processing_hooks or []:
         result = run_sync_only(post_processor, result)
 
+    if hasattr(result, "layout"):
```
Goldziher: Why do we need this?

kreuzberg/extraction.py
```diff
@@ -58,30 +60,54 @@ def _validate_and_post_process_helper(result: ExtractionResult, config: Extracti
...
+        result.layout = None
```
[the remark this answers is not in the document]
Goldziher: Please address

pyproject.toml
```diff
@@ -52,6 +52,7 @@ dependencies = [
   "html-to-markdown[lxml]>=1.6.0",
   "mcp>=1.11.0",
   "msgspec>=0.18.0",
+  "pandas>=2.2.0",
```
Goldziher: This can't be a none optional dependency, it's too large

Goldziher:
please rebase on main. ou will see the updated structure of the `_ocr` module, which shows a clear seperateion between sync and async

kreuzberg/_types.py
naderalexan: so the result would not have layout as it is irrelevent to the result. Not doing this would not really have an "negative" effect, but I think it provides a cleaner result this way

kreuzberg/extraction.py
naderalexan: in case of vision model document categorization, removing layout from result. This is to provide a clean result at the end. Not a big deal, can remove it if you are okay with the result having the attr layout (which is useless for the result)

kreuzberg/_document_classification.py
naderalexan: You are right, let me think of a better solution

kreuzberg/_document_classification.py
```diff
@@ -50,7 +56,13 @@ def classify_document(result: ExtractionResult, config: ExtractionConfig) -> tup
         A tuple containing the detected document type and the confidence score,
         or (None, None) if no type is detected with sufficient confidence.
     """
-    content_lower = result.content.lower()
+    if not GoogleTranslator:
```
Goldziher: I would do the import inside the function simply, or abstract it into its own function-

`get_google_translator` etc.

PR #93, opened 2025-07-22
feat: add crypto extra for more PDF coverage

chbndrhnns:
Some pdfs on my disk cause an error when trying them with Kreuzberg, mentioning the `crypto` extra for playa-pdf. Is it a bad idea to add it?


Goldziher:
well, lets start by fixing ci

chbndrhnns:
> well, lets start by fixing ci

I am a bit lost how adding an extra for a dependency causes a drop in coverage. Do you have an idea?

Goldziher:
crypto was added as an optional depenedencies in the latest version

PR #251, opened 2025-12-24
fix(docker): install ONNX Runtime from GitHub releases

dbhowmick:
## Summary
  - Install ONNX Runtime 1.23.2 from GitHub releases instead of apt
  - libonnxruntime1.23 package is unavailable in Debian repositories
  - Download in builder stage, copy lib/include files to runtime stage

  ## Test plan
  - [x] Build Dockerfile.core for amd64
  - [x] Build Dockerfile.full for amd64
  - [x] Verify running /extract with embedding enabled with the REST endpoint.


Goldziher:
Thank you for the PR. I'm afraid though the issue is that ort doesnt support the newest onnxruntimes.

PR #263, opened 2026-01-04
feat: add Djot format support

### LINKED ISSUE #262: feat: add Djot format support
Is there any interest in also supporting [Djot](https://djot.net/)?

I build the PHP version for it:
https://github.com/php-collective/djot-php

This would require a ruby clone? or how would one be able to support that format?

I use that nowadays for everything instead of markdown, for reasons stated in the docs.

dereuromark:
## Summary

- Add new `DjotExtractor` for parsing Djot markup documents
- Uses the `jotdown` crate (pull-parser similar to `pulldown-cmark`)
- Supports YAML frontmatter metadata extraction
- Supports table extraction as structured data
- Registered under the `office` feature flag

## Implementation Details

Djot is a modern markup language designed by John MacFarlane (creator of Pandoc and CommonMark). It has simpler parsing rules than CommonMark while supporting similar features.

The implementation follows the same pattern as the existing `MarkdownExtractor`:
- YAML frontmatter parsing for metadata (title, author, date, keywords, etc.)
- Event-based text extraction from the AST
- Table extraction with markdown output format
- Title extraction from first heading if not in frontmatter

MIME types: `text/djot`, `text/x-djot`

## Test plan

- [ ] Unit tests included in the PR
- [ ] `cargo test -p kreuzberg --features office` passes
- [ ] Manual testing with sample `.djot` files

Closes #262

### DIFF
--- a/crates/kreuzberg/src/extractors/djot.rs
+++ b/crates/kreuzberg/src/extractors/djot.rs
@@ -10,21 +10,21 @@
 //! Djot is a modern markup language with simpler parsing rules than CommonMark.
 //! See https://djot.net for the specification.
 //!
-//! Requires the `office` feature.
+//! Requires the `djot` feature.
 
-#[cfg(feature = "office")]
+#[cfg(feature = "djot")]
 use crate::Result;
-#[cfg(feature = "office")]
+#[cfg(feature = "djot")]
 use crate::core::config::ExtractionConfig;
-#[cfg(feature = "office")]
+#[cfg(feature = "djot")]
 use crate::plugins::{DocumentExtractor, Plugin};
-#[cfg(feature = "office")]
+#[cfg(feature = "djot")]
 use crate::types::{ExtractionResult, Metadata, Table};
-#[cfg(feature = "office")]
+#[cfg(feature = "djot")]
 use async_trait::async_trait;
-#[cfg(feature = "office")]
+#[cfg(feature = "djot")]
 use jotdown::{Container, Event, Parser};
-#[cfg(feature = "office")]
+#[cfg(feature = "djot")]
 use serde_yaml_ng::Value as YamlValue;
 
 /// Djot markup extractor with metadata and table support.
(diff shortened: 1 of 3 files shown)


crates/kreuzberg/src/extractors/djot.rs
```diff
@@ -0,0 +1,637 @@
...
+//! - Djot parsing using the jotdown crate
+//! - YAML frontmatter metadata extraction (same as Markdown)
+//! - Table extraction as structured data
+//! - Heading structure preservation
+//! - Code block and link extraction
+//!
+//! Djot is a modern markup language with simpler parsing rules than CommonMark.
+//! See https://djot.net for the specification.
+//!
+//! Requires the `office` feature.
```
Goldziher: Q: why is this needed?

dereuromark: This doc comment follows the pattern used in other extractors (e.g. `markdown.rs` line 13). It documents that the extractor is only compiled when the `office` feature is enabled, which is useful for users reading the docs.

That said, it's not strictly necessary - I can remove it if you prefer a cleaner module doc.

Goldziher: no, i meant - why do we need the office features? 

This is a seperate format, and as far as i saw (cursory glance) you dont really need anything form that group.

dereuromark: Good point - refactored to a separate `djot` feature. It now only pulls in `jotdown` + `tokio-runtime`, independent of `office`.

Goldziher (commented):
Thanks for this contribution! The implementation is solid overall, but there are several issues that need to be addressed before merge:

## Critical Issues

### 1. Doc domment (line 13)
The doc comment still says `Requires the 'office' feature` but should say `Requires the 'djot' feature` after your refactor.

### 2. Code Duplication
`extract_frontmatter()` and `extract_metadata_from_yaml()` are identical to the markdown extractor (142 lines duplicated). This creates maintenance burden.

### 3. Inefficient Double Parsing (lines 348-363)
The content is parsed twice - once for text extraction, once for tables. You're already collecting events into a `Vec`, so they could be processed in a single pass.

### 4. Test Plan Incomplete
The PR description shows unchecked test boxes, and one still references the wrong feature (`office` instead of `djot`).

### 5. Frontmatter Parser Edge Case
The parser breaks if YAML content contains `---` on its own line or uses `...` as a terminator (both valid YAML). This is also a bug in the markdown extractor, but worth fixing here.

Also -- please update the root readme, and the documentation under docs as required.

Goldziher:
Im merging this -- ill bring this into the next release.

Thank you for contributing! @dereuromark

PR #346, opened 2026-02-02
fix(ocr): pass page count and evaluate per-page for OCR fallback heur…

yiftachashkenazi:
Fix OCR heuristic not triggering on scanned and mixed-content PDFs by passing actual page count to the text quality evaluator (was hardcoded to None, defaulting to 1)
Add per-page OCR evaluation using existing page boundaries — if any single page fails the quality check, the entire document gets OCR'd
Context
Reported in Discord by Mohamed: PDFs containing only OCR text returned no results unless force_ocr=true was explicitly set. Root cause was two bugs in the heuristic layer:

Inflated per-page averages: evaluate_native_text_for_ocr was called with page_count=None (defaults to 1). A 20-page scanned PDF with 64 garbled chars total would calculate avg_non_whitespace=64.0 instead of 3.2, passing the quality check.
No per-page granularity: Text quality was evaluated as a single blob. In mixed-content PDFs, pages with real text masked scanned pages in the aggregate stats.
Changes
extraction.rs: Stop discarding page boundaries, include them in the return tuple
ocr.rs: Add evaluate_per_page_ocr — evaluates each page independently, early-returns on first failing page
mod.rs: Wire per-page evaluation with actual page count into the main OCR decision flow
Test plan
 cargo test --features pdf,ocr -p kreuzberg — 834 passed, 20 failed (identical to clean main, all failures are pre-existing missing libpdfium.dylib)
 cargo clippy --features pdf,ocr — zero warnings
 cargo clippy --features pdf (no ocr) — zero warnings
 cargo fmt -- --check — clean
 Verified same 20 test failures on clean main branch (pre-existing pdfium dep issue)

### DIFF
--- a/crates/kreuzberg/src/extractors/pdf/mod.rs
+++ b/crates/kreuzberg/src/extractors/pdf/mod.rs
@@ -375,6 +375,159 @@ mod tests {
         assert!(ocr::evaluate_native_text_for_ocr(sample, Some(2)).fallback);
     }
 
+    #[cfg(feature = "ocr")]
+    #[test]
+    fn test_per_page_ocr_no_boundaries_falls_back_to_whole_doc() {
+        let text = "This document has enough meaningful words for evaluation purposes here.";
+        let decision = ocr::evaluate_per_page_ocr(text, None, Some(1));
+        assert!(!decision.fallback);
+    }
+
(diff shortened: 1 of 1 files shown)


yiftachashkenazi:
thanks! @Goldziher

Issue #359, opened 2026-02-06
bug: DOCX list items missing whitespace between runs


crishoj:
With the attached sample, `kreuzberg extract` merges text runs without a space separator.

## Input

A DOCX with bullet/dash list items. The last items in the document are:

  - Sermocination ypsiliform
  - Foo bar

## Expected output

```
Sermocination ypsiliform
Foo bar
```

## Actual output

```
Sermocinationypsiliform
Foo bar
```

"Sermocination" and "ypsiliform" are separate text runs in the same list item — kreuzberg concatenates them without a space.

## Version

kreuzberg-cli installed from `releases/latest` (tested 2026-02-06)

[sample.docx](https://github.com/user-attachments/files/25131041/sample.docx)

Goldziher:
Thanks for reporting.

crishoj:
Appears to be an upstream issue in `docx-lite`.

PR submitted: https://github.com/v-lawyer/docx-lite/pull/1

Goldziher:
fix is incoming. new version v4.2.12 includes the fix, and will be released tonight

PR #362, opened 2026-02-06
fix(docx): add regression test for #359 list run whitespace

crishoj:
Adds a failing test for #359 — DOCX list items with multiple text runs lose whitespace between runs ("Sermocinationypsiliform" instead of "Sermocination ypsiliform").

Root cause is `trim_text(true)` in docx-lite's XML parser stripping whitespace from `<w:t>` elements. Upstream fix submitted: https://github.com/v-lawyer/docx-lite/pull/1

This PR is ready to merge once the upstream fix lands and `docx-lite` is updated.


Goldziher:
thank you.

Goldziher:
a fix is already i n place. i merged this.

PR #366, opened 2026-02-07
fix: enable musl CLI build with cross-compiler toolchain

### LINKED ISSUE #184: build: Multiple architecture failures in v4.0.0-rc.1 release build
## Summary

During the v4.0.0-rc.1 release build (workflow run [19371784104](https://github.com/Goldziher/kreuzberg/actions/runs/19371784104)), multiple architecture builds failed unexpectedly:

### 1. CLI Binaries
- ❌ **x86_64-unknown-linux-musl** - Expected (musl target not previously tested in CI)
- ❌ **aarch64-unknown-linux-gnu** - UNEXPECTED (this target succeeded in build-artifacts)

### 2. Package Rust Crates
- ❌ **Package Rust crates job** - Completely unexpected failure

### 3. Node.js Bindings (ARM targets)
- ❌ **aarch64-unknown-linux-gnu** - UNEXPECTED (musl variant succeeded)
- ❌ **armv7-unknown-linux-gnueabihf** - UNEXPECTED

## Context

These failures are separate from the known issues:
- #182 (Windows x64 Ruby gems - ONNX Runtime)
- #183 (Linux aarch64 Python wheels - ring crate)

## Successful Builds

✅ Python wheels: linux-x86_64, macos-universal, windows-amd64
✅ Python sdist
✅ CLI binaries: x86_64-unknown-linux-gnu, aarch64-apple-darwin, x86_64-pc-windows-msvc
✅ Ruby gems: linux, macos-arm64
✅ Node bindings: Most targets including aarch64-unknown-linux-musl, x86_64-unknown-linux-musl

## Investigation Needed

1. **aarch64-unknown-linux-gnu CLI** - Why did this fail in publish but not in build-artifacts?
2. **Package Rust crates** - What dependency or configuration issue is causing this?
3. **ARM Node bindings** - Why do musl variants succeed but gnu/gnueabihf variants fail?

## Logs

Full workflow logs will be available once run 19371784104 completes.

## Workaround

These builds have been configured with `continue-on-error` in the workflow to prevent blocking releases while we investigate.

crishoj:
Closes #184.

- Install [musl.cc](https://musl.cc) cross-compiler in CI (provides musl-native g++/libstdc++)
- Fix build.rs compiler detection: use clang only when explicitly set via CXX/CC env vars, default to g++ (works in Alpine and with musl.cc)
- Fix link flags: drop `dylib=` prefix for musl (static binaries), skip `stdc++fs` (merged into libstdc++ in GCC 9+)
- Re-enable `x86_64-unknown-linux-musl` in publish workflow

## Test plan
- [x] Verified build succeeds in Alpine Docker container (`rust:alpine`)
- [x] Produced binary is statically linked (`static-pie linked, stripped`)
- [ ] CI dry-run on fork (publish workflow needs matching version tag to pass validation)
- [ ] Runtime test on staging server

### DIFF
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -813,7 +813,9 @@ jobs:
           - os: ubuntu-24.04-arm
             target: aarch64-unknown-linux-gnu
             use_cross: false
-          # TODO: Re-enable x86_64-unknown-linux-musl once build issues are resolved
+          - os: ubuntu-latest
+            target: x86_64-unknown-linux-musl
+            use_cross: false
           - os: macos-latest
             target: aarch64-apple-darwin
             use_cross: false
--- a/crates/kreuzberg-tesseract/build.rs
+++ b/crates/kreuzberg-tesseract/build.rs
@@ -447,14 +447,21 @@ mod build_tesseract {
         let target_windows = is_windows_target(&target);
         let target_msvc = is_msvc_target(&target);
         let target_mingw = is_mingw_target(&target);
-        let target_musl = target.contains("musl");
 
         if target_macos {
             cmake_cxx_flags.push_str("-stdlib=libc++ ");
             cmake_cxx_flags.push_str("-std=c++17 ");
         } else if target_linux {
             cmake_cxx_flags.push_str("-std=c++17 ");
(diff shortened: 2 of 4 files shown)


crishoj:
(Not quite there yet, Pdfium fails with "Dynamic loading not supported")

PR #382, opened 2026-02-12
feat: add .mdx support

martinschaer:
This PR adds support to extract MDX files (mime type: text/mdx), which are commonly used for documenting software.

Example MDX file:

```mdx
import {Chart} from './snowfall.js'
export const year = 2013

# Last year’s snowfall

In {year}, the snowfall was above average.
It was followed by a warm spring which caused
flood conditions in many of the nearby rivers.

<Chart year={year} color="#fcb32c" />
```


Goldziher:
mdx support will be released in v4.3.8

PR #419, opened 2026-02-26
feat(ocr): add rapid-ocr backend integration

chikingsley:
Draft for CI validation before review.\n\nScope: rapid-ocr backend integration and Python registration compatibility fixes.\n\nWill mark ready after CI is green.

### DIFF
--- a/.github/workflows/ci-cli.yaml
+++ b/.github/workflows/ci-cli.yaml
@@ -61,7 +61,7 @@ env:
 jobs:
   build-cli:
     name: CLI Build & Test (${{ matrix.os }})
-    if: (github.repository == 'kreuzberg-dev/kreuzberg' || github.repository == 'chikingsley/kreuzberg') && github.actor != 'dependabot[bot]'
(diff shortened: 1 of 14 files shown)


Goldziher:
Hi there @chikingsley. I see you are interested in contributing. 

Its best you open an issue or a discussion first in the future - before writing code, to see that the contribution is something we are interested in. Or you can join our discord server, and write in the pertinent channel. 

RapidOCR for Python -- perhaps, I'm not sure frankly its something that would be useful across the board. I would be happy to see some benchmarks of RapidOCR to justify this inclusion. 

Regarding CI -- i disallowed running our CI on forks. It clogs our own development due to limited resources. You can run CI on your own end, or compile and run locally. Since you are only modifying the Python bindings with this addition, it should be easy enough to test locally on your end.

Goldziher:
Ok, looking at this further. RapidOCR is basically a smarter Python distribution of PaddleOCR. It gives us no benefits -- we already offer PaddleOCR via our Rust core with ONNX optimizations.

chikingsley:
Noted - I will go that route in the future. I'll close this in this case.

PR #447, opened 2026-03-07
fix(extractors/pdf): handle password protected pdfs

saiintbrisson:
Hi, 

I've been playing around with this (awesome) library for my little bot that handles a few bank PDFs for me. One bank sends me password encoded PDFs, and my bot uses the kreuzberg CLI to extract it. Only to fail. Even when adding an inline JSON with the passwords, it still fails. So I came looking for answers.

This is a pretty small patch to handle password protected PDFs in a very naive way. I'm iterating through the configured passwords and trying to load the PDF. The first iteration has no passwords.

I've tried looking for a related issue and PRs, couldn't find any.

I ran the tests with `task` and installed it locally, the passwords are now used. The test though, there's no local `copy_protected` file, so I couldn't test. Guidance is welcomed.


saiintbrisson:
P.S.: Passing the inline config JSON isn't the best UX for a CLI, but I didn't want to extend this PR to beyond what I actually needed. Maybe a `password` CLI flag could help?

PR #472, opened 2026-03-11
chore: migrate publish-pypi and publish-nuget to shared actions

Goldziher:
## Summary

Migrates two publish jobs to use shared `kreuzberg-dev/actions` publish actions:

- **publish-pypi**: Replace inline `pypa/gh-action-pypi-publish` + listing + dry-run with `kreuzberg-dev/actions/publish-pypi@v1`
- **publish-nuget**: Replace inline .NET setup + `dotnet nuget push` + dry-run with `kreuzberg-dev/actions/publish-nuget@v1`

Both shared actions handle: package listing, skip-existing/skip-duplicate, dry-run summaries internally.

### Why only these two?

These are the cleanest migrations where the shared action fully replaces the inline logic. Other publish jobs have gaps:
- **crates.io**: Requires multi-crate dependency ordering with waits between publishes
- **npm**: Requires two-phase publish (platform packages → wait → main package)
- **RubyGems**: Requires OIDC trusted publishing setup that shared action doesn't include
- **Maven**: Requires extensive native resource staging (PDFium, ONNX, Java natives)
- **Hex**: Requires checksum generation from GitHub release assets before publish
- **Homebrew**: No shared action exists

These can be addressed in follow-up PRs by enhancing the shared actions.

## Test plan

- [ ] Verify publish workflow YAML is valid (CI actionlint)
- [ ] Dry-run publish to confirm PyPI and NuGet jobs work with shared actions
- [ ] Verify OIDC `id-token: write` permission still applies for PyPI trusted publishing
- [ ] Verify `NUGET_API_KEY` env var is passed correctly to shared action


Goldziher:
Consolidated into chore/use-shared-actions branch

PR #493, opened 2026-03-13
fix(email): decode MSG PT_STRING8 properties using PR_MESSAGE_CODEPAGE

hoesler:
## Problem

  Outlook .msg files store ANSI string properties (type PT_STRING8 / 0x001E) encoded with the Windows code page
  declared in PR_MESSAGE_CODEPAGE (0x3FFD) or PR_INTERNET_CPID (0x3FDE). The previous implementation decoded these
  bytes with String::from_utf8_lossy, which silently replaces non-ASCII bytes with U+FFFD (�).

  This caused garbled output for any MSG file whose body, subject, or sender fields contain characters outside
  ASCII — including accented Latin characters common in Western European languages (e.g. ö → 0xF6 in CP1252),
  Cyrillic, Greek, Hebrew, Arabic, and CJK encodings.

## Fix

  - Read PR_MESSAGE_CODEPAGE (0x3FFD), falling back to PR_INTERNET_CPID (0x3FDE), from the message's
  __properties_version1.0 stream before decoding any string properties
  - Add encoding_for_windows_codepage() mapping Windows code page integers to encoding_rs encodings (CP1250–1258,
  Shift-JIS, GBK, EUC-KR, Big5, ISO-8859 variants), with a windows-1252 fallback for files that omit the property
  - Add read_msg_int_prop() to read PT_LONG (0x0003) values from the properties stream, following the same pattern
  as the existing read_msg_filetime_prop
  - Thread the resolved code page through all PT_STRING8 reads: body, HTML body, subject, sender name/address,
  recipients, attachment filenames, and transport headers


Goldziher:
nice!

PR #505, opened 2026-03-16
feat(email): make MSG PT_STRING8 fallback codepage configurable

hoesler:
Follow-up to #493, which fixed MSG `PT_STRING8` decoding by reading `PR_MESSAGE_CODEPAGE` / `PR_INTERNET_CPID` and
  using `encoding_rs`. That fix hardcoded windows-1252 as the last-resort fallback for MSG files that declare no
  codepage property.

  This PR makes that fallback configurable for users with non-Western-European MSG archives (e.g. Cyrillic, Greek,
  Central European).

  - Adds `EmailConfig { msg_fallback_codepage: Option<u32> }` to `ExtractionConfig`
  - Re-exports `EmailConfig` from the crate root alongside other config types
  - `parse_msg_content` and `extract_email_content` now accept `fallback_codepage: Option<u32>;` pass `None` for the
  previous windows-1252 default
  - File-declared codepages always take precedence; only the last-resort fallback is affected

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,8 +20,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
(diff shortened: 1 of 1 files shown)


hoesler:
@Goldziher As you already merged the PR, what does that mean for the open tasks from your review feedback? Is there still something for me to do, or will you do it yourself?

PR #511, opened 2026-03-17
fix: PDF image extraction returning raw compressed data

### LINKED ISSUE #500: bug: PDF Image Extraction Returns Raw Compressed Data Instead of Decoded Images
## Problem

Using `ImageExtractionConfig(extract_images=True)` gives me raw FlateDecode compressed bytes instead of actual image data.

```python
config = ExtractionConfig(images=ImageExtractionConfig(extract_images=True))
result = extract_file_sync("paper.pdf", config=config)

img = result.images[0]
print(img.format)           # "FlateDecode" - this is wrong
print(img.data[:4].hex())   # "78daecdd" - zlib header, not PNG/JPEG
```

The `format` field shows "FlateDecode" (compression method) instead of the actual image format. The `data` is still compressed and can't be used directly.

### 1. `crates/kreuzberg/src/pdf/images.rs`

```rust
data: Bytes::from(img.content.to_vec()),  // this is raw compressed stream from lopdf
```

`lopdf::Document::get_page_images()` returns raw PDF stream data, which is compressed. kreuzberg doesn't decode it.

### 2. `crates/kreuzberg/src/extractors/pdf/mod.rs`

```rust
let format = img
    .filters
    .first()
    .cloned()
    // ...
```

The first filter (e.g. "FlateDecode") is used as the image format. That's not correct - FlateDecode is compression, not format.

### 3. `filters` field is lost

`PdfImage` has `filters: Vec<String>` but it's not passed to `ExtractedImage`. So even if I wanted to decode manually, I can't tell what compression was used.

## My workaround

I'm detecting zlib header and decompressing manually, then using PIL to convert raw pixels to PNG:

```python
import zlib
from PIL import Image

def decode_image(data, width, height):
    if data[:2] in (b'\x78\x9c', b'\x78\xda'):
        data = zlib.decompress(data)

    if data[:8] == b'\x89PNG\r\n\x1a\n':
        return data  # already PNG

    # raw pixels -> PNG
    img = Image.frombytes('RGB', (width, height), data)
    buf = BytesIO()
    img.save(buf, format='PNG')
    return buf.getvalue()
```

But this is fragile - I'm guessing the colorspace and don't handle all cases.

## Suggestion

Either:
1. Decode images before returning (FlateDecode -> decompress -> encode as PNG, DCTDecode -> return as-is with format="jpeg")
2. Or expose `filters` field so users can decode properly

kreuzberg 4.4.5, Python 3.12, macOS

kh3rld:
implemented fix for the issue where PDF image extraction returned raw, compressed bytes (e.g., FlateDecode/DCTDecode) directly from the PDF stream instead of usable image data

**Key Changes:**
- **Decoding Engine:** Implemented [decode_image_data](kreuzberg/crates/kreuzberg/src/pdf/images.rs) in [crates/kreuzberg/src/pdf/images.rs](kreuzberg/crates/kreuzberg/src/pdf/images.rs) to handle PDF filters.
    - `DCTDecode` (JPEG) & `JPXDecode` (JPEG2000) are now correctly labeled and passed through.
    - `FlateDecode` (zlib) streams are decompressed and re-encoded as standard PNG files.
- **Metadata Update:** Added `decoded_format` to [PdfImage](kreuzberg/crates/kreuzberg/src/pdf/images.rs) struct to ensure the output `ExtractedImage.format` field reflects standard MIME types (e.g., `"png"`, `"jpeg"`) instead of PDF internal filter names.
- **Dependency Management:** Explicitly added `flate2` to the [pdf](kreuzberg/test_documents/pdf/embedded_images_tables.pdf) feature in [Cargo.toml](kreuzberg/Cargo.toml).
- **Validation:** Added unit tests covering all decoding scenarios, including a round-trip zlib test.

### How to test:
Run the focused unit tests:
```bash
cargo test --lib -p kreuzberg --features pdf pdf::images
```

closes #500

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -85,6 +85,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- **PDF image extraction decoding**: Images returned from PDFs (FlateDecode, DCTDecode, JPXDecode) are now automatically decoded and re-encoded as standard formats (PNG/JPEG/etc.) instead of returning raw compressed bytes. Updated the image format mapping to reflect standard MIME-compatible types.
 - **PDF structure tree heading trust**: Structure tree heading tags (H1–H6) are now trusted as author-intent metadata with only a word-count guard. Previously, font-size validation rejected valid headings when the font was close to body size.
 - **PDF structure tree extraction performance**: MCID text and style maps are now built in a single pass over page objects (was two passes). Uses the pre-loaded `FPDFText` page handle instead of calling `FPDFText_LoadPage` per text object, eliminating multi-second extraction times on complex pages.
 - **OCR layout: Picture regions no longer suppress text**: Layout-detected Picture regions (diagrams, figures) now preserve any embedded text as plain paragraphs instead of silently dropping it. Uses content-aware `is_substantive_text()` heuristic to distinguish real text from OCR artifacts.


Goldziher:
@kh3rld -- after rebasing, also update the changelog. fix -- v4.5.0 (unreleased).

PR #517, opened 2026-03-18
docs: rewrite the comparison page

pratik-mahalle:


pratik-mahalle:
> I think we need to either remove this page, or extend it to cover properly other open source libraries in our category

Hmm... I think we need to keep this but in the extended format covering all other open source libraries. Can you please list down other open source libraries

Goldziher:
> > I think we need to either remove this page, or extend it to cover properly other open source libraries in our category
> 
> Hmm... I think we need to keep this but in the extended format covering all other open source libraries. Can you please list down other open source libraries 

There are quite a few. Start with checking what we benchmark against. 

Apache tika is the classic gold standard. It's a Java lib. 

Also note licenses - all the `mupdf` style libs for example are AGPL, others are GPL etc.

pratik-mahalle:
I have the plan  - 
The overview page (docs/comparisons/index.md or overview.md) would be a single table covering all 12 libraries at a glance -- license, language, format coverage, focus area. people will land here first when evaluating alternatives and want a quick scan before deciding which to read further. This replaces needing to open 12 browser tabs.

Individual pages only for the libraries developers actually choose instead of Kreuzberg in real projects.

What do you think ? @Goldziher

Goldziher:
> I have the plan  - 
> The overview page (docs/comparisons/index.md or overview.md) would be a single table covering all 12 libraries at a glance -- license, language, format coverage, focus area. people will land here first when evaluating alternatives and want a quick scan before deciding which to read further. This replaces needing to open 12 browser tabs.
> 
> Individual pages only for the libraries developers actually choose instead of Kreuzberg in real projects.
> 
> What do you think ? @Goldziher 
> 
> 

Go ahead

PR #522, opened 2026-03-18
fix: resolve type inference failure in docx parser due to quick-xml ambiguity 

### LINKED ISSUE #519: bug: Compile error in docx parser: `Cow::as_ref()` type inference failure with `typed-path` in dependency tree (since 4.4.2)...
- The bug is dependency-tree dependent — typed-path adds extra AsRef impls for Cow<'_, [u8]> which breaks type inference on .as_ref()
   - The fix is a one-liner: attr.value.as_ref() as &[u8] or &*attr.value at both locations
   - This explains why it passes their CI but fails for downstream users with richer dep trees

kh3rld:
I've just resolved a compilation error in the docx parser caused by type inference failure in quick-xml's local_name().as_ref() and name().as_ref() calls

The issue surfaced because zip 8.2.0 introduced typed-path into the dependency tree. Both [u8] and typed_path::Path implement traits that as_ref() can target, creating ambiguity for the compiler when using quick-xml 0.37+

changes:

 - Added explicit type casts (as &[u8]) to as_ref() calls on LocalName and QName objects.
 - Fixed operator precedence in cases where .to_vec() was called on casted results

verification:
 - Verified successful compilation of the kreuzberg crate
 - Confirmed that existing test failures in the repository are unrelated to these changes
 
 
fixes #519

### DIFF
--- a/crates/kreuzberg/src/extraction/docx/math.rs
+++ b/crates/kreuzberg/src/extraction/docx/math.rs
@@ -128,7 +128,7 @@ fn collect_children(reader: &mut Reader<&[u8]>, end_tag: &[u8]) -> Vec<MathNode>
     loop {
         match reader.read_event_into(&mut buf) {
             Ok(Event::Start(ref e)) => {
-                let tag = e.name().(as_ref() as &[u8]).to_vec();
+                let tag = (e.name().as_ref() as &[u8]).to_vec();
                 match tag.as_slice() {
                     b"m:r" => {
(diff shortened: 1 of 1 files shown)


docs/integrations/index.md
```diff
@@ -0,0 +1,10 @@
+# Integrations
```
Goldziher, line 1: You got a mixup here

PR #523, opened 2026-03-18
fix: pass mime_type to extract_file_in_worker binding

NickKoester:
When using the `extractFileInWorker` function in the node bindings, I noticed that the explicit `mime_type` option was not being properly passed through to the Rust binding. This could cause unexpected errors if the files to be extracted did not have an extension present.

This change makes the binding code consistent with the way the Typescript wrapper is expecting it to be used.
https://github.com/kreuzberg-dev/kreuzberg/blob/e7c8e4b7a018e0e339bd73746f88adbb7ab792ef/crates/kreuzberg-node/typescript/extraction/worker-pool.ts#L133-L138

I went ahead and removed a chunk of code that added the old password parameter to the pdf config, as it didn't appear to be used for anything anymore. My assumption was that users could simply pass in the complete list of passwords in their pdf config, but let me know if my assumption is incorrect.

### DIFF
--- a/crates/kreuzberg-node/src/worker_pool_api.rs
+++ b/crates/kreuzberg-node/src/worker_pool_api.rs
@@ -136,7 +136,7 @@ pub fn get_worker_pool_stats(pool: &JsWorkerPool) -> Result<WorkerPoolStats> {
 pub async fn extract_file_in_worker(
     pool: &JsWorkerPool,
     file_path: String,
-    password: Option<String>,
+    mime_type: Option<String>,
     config: Option<JsExtractionConfig>,
 ) -> Result<JsExtractionResult> {
     let pool_id = pool.pool_id;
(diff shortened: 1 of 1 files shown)


crates/kreuzberg-node/src/worker_pool_api.rs
```diff
@@ -136,7 +136,7 @@ pub fn get_worker_pool_stats(pool: &JsWorkerPool) -> Result<WorkerPoolStats> {
 pub async fn extract_file_in_worker(
     pool: &JsWorkerPool,
     file_path: String,
-    password: Option<String>,
```
Goldziher, line 139: Why you removed this?

crates/kreuzberg-node/src/worker_pool_api.rs
```diff
@@ -162,17 +162,8 @@ pub async fn extract_file_in_worker(
 
     let mut rust_config = resolve_config(config)?;
 
-    // Inject password into PDF config if provided
```
Goldziher, line 165: Why remove?

crates/kreuzberg-node/src/worker_pool_api.rs
NickKoester: I didn't remove this, I renamed to to more accurately reflect what was being passed in from the Nodejs calling code.
https://github.com/kreuzberg-dev/kreuzberg/blob/e7c8e4b7a018e0e339bd73746f88adbb7ab792ef/crates/kreuzberg-node/typescript/extraction/worker-pool.ts#L133-L138

The 4 arguments are:
1. pool
2. file_path
3. mime_type
4. config

In the Rust binding signature the 3rd argument (`mime_type`) was called `password`, which I thought was a mistake.

crates/kreuzberg-node/src/worker_pool_api.rs
NickKoester: If `mime_type` is passed into this function instead of `password`, which I think it should be (see above), then this block of code is not doing anything other than adding the Mime type to the list of passwords in the config, which seems incorrect.

Goldziher: gotcha.

Goldziher:
thanks for contributing!

PR #554, opened 2026-03-23
feat: add markdown heading chunker type

tobocop2:
Thanks for this awesome library. I hope you don't mind, but I am creating a PR here because I figured this would be useful to share with others. I am doing similar in one of my projects but would rather delegate to this library.

---

right now kreuzberg returns `chunks: None` for markdown. the existing `Markdown` chunker type handles formatting but doesn't split at heading boundaries or give any section context.

this matters a lot for RAG. 

For example, when chunk just says "The default is 512 tokens" the LLM has no idea what that refers to. but if the chunk says:

```
# Configuration > ## Chunking > ### Chunk Size

The default is 512 tokens.
```

...now it knows exactly where it came from.

[Anthropic's Contextual Retrieval research (2024)](https://www.anthropic.com/news/contextual-retrieval) showed that adding document context to chunks reduces retrieval failures by 49%. their approach uses an extra LLM call per chunk which is expensive. this does a similar thing by just prepending the heading hierarchy, so zero extra cost.

### how it works

reuses the existing heading extraction from `headings.rs` (heading maps + hierarchy resolution already exist). the new chunker splits at heading boundaries and embeds the path in each chunk. oversized sections get sub-split at paragraph boundaries while keeping the heading prefix. all string splitting is UTF-8 safe — tested against CJK, emoji (including 4-byte ZWJ), and mixed scripts.

falls back to `ChunkerType::Markdown` when there are no headings,

```python
config = ChunkingConfig(chunker_type="markdown_heading", max_chars=1000)
```

28 tests — all heading levels H1-H6, hierarchy reset across siblings, multi-byte UTF-8, emoji, large section splitting, prefix > max_chars, fallback behavior, byte offset bounds.

### DIFF
--- a/crates/kreuzberg/src/core/config/processing.rs
+++ b/crates/kreuzberg/src/core/config/processing.rs
@@ -13,12 +13,18 @@ use std::path::PathBuf;
 ///
 /// * `Text` - Generic text splitter, splits on whitespace and punctuation
 /// * `Markdown` - Markdown-aware splitter, preserves formatting and structure
+/// * `MarkdownHeading` - Splits at heading boundaries, prepends heading hierarchy
+///   path to each chunk (e.g., "# Setup > ## Install"). Inspired by Anthropic's
+///   Contextual Retrieval (2024) which showed 49% fewer retrieval failures when
+///   adding document context to chunks.
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Serialize, Deserialize, Default)]
 #[serde(rename_all = "lowercase")]
 pub enum ChunkerType {
     #[default]
     Text,
     Markdown,
+    #[serde(rename = "markdown_heading")]
+    MarkdownHeading,
 }
 
 /// How chunk size is measured.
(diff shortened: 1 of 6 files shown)


crates/kreuzberg/src/core/config/processing.rs
```diff
@@ -13,12 +13,18 @@ use std::path::PathBuf;
...
+///   path to each chunk (e.g., "# Setup > ## Install"). Inspired by Anthropic's
+///   Contextual Retrieval (2024) which showed 49% fewer retrieval failures when
+///   adding document context to chunks.
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Serialize, Deserialize, Default)]
 #[serde(rename_all = "lowercase")]
 pub enum ChunkerType {
     #[default]
     Text,
     Markdown,
+    #[serde(rename = "markdown_heading")]
```
Goldziher, line 26: This is unnecessary

Goldziher:
superseded -- v4.5.4 released with a new chunking config parameter that supports this behavior

PR #555, opened 2026-03-23
feat: add yaml section chunker type

tobocop2:
Thanks for this awesome library. I hope you don't mind, but I am creating a PR here because I figured this would be useful to share with others. I am doing similar in one of my projects but would rather delegate to this library.

---

adds a structured chunker for YAML and JSON that splits by keys with hierarchy path prefixes. nested structures produce leaf-only chunks — parent keys appear in the prefix, not as separate chunks.

```yaml
database:
  primary:
    host: db1
  replica:
    host: db2
```

becomes two chunks: `# database > primary > host` and `# database > replica > host`.

JSON works the same way (minified or pretty-printed). arrays at root fall back to text chunker.

### auto-inference

the chunking processor reads `data_format` from extraction metadata and auto-selects the structured chunker for yaml/json. no explicit `chunker_type` config needed. explicit user choice always wins.

### implementation

both YAML and JSON are parsed into value trees (`serde_yaml_ng` for YAML, `serde_json` for JSON) and walked with a shared BFS function (`VecDeque` work queue, not recursion) to avoid stack overflow on deeply nested input ([RUSTSEC-2024-0012](https://rustsec.org/advisories/RUSTSEC-2024-0012.html)). oversized leaf sections get sub-split with the text splitter, keeping the key prefix. BFS inner loop avoids cloning subtree values by taking ownership, which matters for multi-MB structured data.

41 new tests (34 yaml/json + 4 processor + 3 comparison), including edge cases for invalid YAML, malformed JSON, null values, scalar/list roots, booleans, and missing metadata, including edge cases for invalid YAML, malformed JSON, and null values.

### DIFF
--- a/crates/kreuzberg/src/chunking/core.rs
+++ b/crates/kreuzberg/src/chunking/core.rs
@@ -61,6 +61,11 @@ pub fn chunk_text(
         validate_utf8_boundaries(text, boundaries)?;
     }
 
+    // Yaml creates new content per chunk (key prefix), can't use generic &str splitter.
+    if config.chunker_type == ChunkerType::Yaml {
+        return super::yaml_section::chunk_yaml_by_sections(text, config);
+    }
+
     let text_chunks: Vec<&str> = match &config.sizing {
         #[cfg(feature = "chunking-tokenizers")]
         crate::core::config::ChunkSizing::Tokenizer { model, .. } => {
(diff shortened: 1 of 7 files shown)


Goldziher:
Nice. 

Can we (1) make it the default for yaml / json? That is, it should simply be inferable. (2) How can we deal with nested objects?

tobocop2:
> Nice.
> 
> Can we (1) make it the default for yaml / json? That is, it should simply be inferable. (2) How can we deal with nested objects?

Gave it a shot and just updated the branch to address all your feedback @Goldziher

Thanks!

PR #565, opened 2026-03-24
feat: pst support

### LINKED ISSUE #502: feat: Support for .pst file format (Outlook Personal Folders File Format)
As I understand it, Kreuzberg currently supports .eml and .msg email file formats. Although this is great, when dealing with Microsoft Outlook data files it would be really helpful to also have support for (at least) Outlook's .pst file format.

I’ve found Kreuzberg while exploring the possibility of replacing Apache Tika in a Desktop Search tool which is being developed in Rust. Although leveraging Apache Tika in this way also works, it is really cumbersome and brittle. Having the possibility of using an all-in-one Rust only alternative for all most common desktop formats would be fantastic.

kh3rld:
## what i implemented 

- add support for extracting text and metadata from Microsoft Outlook PST (Personal Folders) archive files
- implement `PstExtractor` plugin backed by the `outlook-pst` crate with iterative folder traversal
- register `application/vnd.ms-outlook-pst` MIME type in native, WASM, C#, and PHP mime maps

## details

**core extraction** (`crates/kreuzberg/src/extraction/pst.rs`):
- Writes PST bytes to a named temp file (required by `outlook-pst::open_store`), then traverses the full IPM subtree iteratively (depth-capped at 50) to avoid stack overflow on deeply nested archives
- Extracts per-message: subject, sender, plain-text body, HTML body, date (Windows FILETIME → ISO-8601), and recipients (TO/CC/BCC) from the recipient table
- Returns a `Vec<EmailExtractionResult>`, reusing the same type as EML/MSG extraction

**extractor plugin** (`crates/kreuzberg/src/extractors/pst.rs`):
- Joins all messages with `---` separators
- Exposes `message_count` in `additional` metadata
- Uses first message for archive-level subject/date/email metadata
- Registered behind `#[cfg(feature = "email")]` alongside `EmailExtractor`

**tests**: 3 unit tests + 5 integration tests (invalid data, empty PST, message count in metadata)

## test plan

- [ ] `cargo test --features email -- pst` — all 8 tests pass
- [ ] `cargo build --features email` — clean build, no warnings
- [ ] Manual smoke test with a real PST file

closes  #502

### DIFF
--- a/crates/kreuzberg/Cargo.toml
+++ b/crates/kreuzberg/Cargo.toml
@@ -53,7 +53,7 @@ office = [
 ]
 hwp = ["dep:cfb", "dep:flate2"]
 iwork = ["dep:zip", "dep:snap"]
-email = ["dep:mail-parser", "dep:cfb", "dep:outlook-pst", "dep:tempfile"]
+email = ["dep:mail-parser", "dep:cfb", "dep:outlook-pst", "dep:tempfile", "dep:chrono"]
 html = ["dep:html-to-markdown-rs"]
 xml = ["dep:quick-xml", "dep:roxmltree"]
 archives = ["dep:zip", "dep:tar", "dep:sevenz-rust2", "dep:flate2"]
@@ -237,7 +237,7 @@ blake3 = { workspace = true }
 bytes = { workspace = true }
 calamine = { version = "0.34.0", features = ["dates"], optional = true }
 cfb = { workspace = true, optional = true }
-chrono = { workspace = true }
+chrono = { workspace = true, optional = true }
 chardetng = { version = "0.1.17", optional = true }
 dashmap = "6.1"
 dbase = { workspace = true, optional = true }
--- a/crates/kreuzberg/src/extraction/pst.rs
+++ b/crates/kreuzberg/src/extraction/pst.rs
@@ -29,7 +29,6 @@ use crate::error::{KreuzbergError, Result};
 use crate::types::{EmailExtractionResult, ProcessingWarning};
 use std::borrow::Cow;
 use std::collections::HashMap;
-use std::string::String;
 
 #[cfg(feature = "email")]
 use outlook_pst::{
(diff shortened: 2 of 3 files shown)


Goldziher:
## Code Review

The architecture is sound — reusing `EmailExtractionResult`, feature-gating behind `email`, and using iterative folder traversal with a depth cap (50) are all good choices. A few issues to address before merge:

### Critical — Build-breaking

**Missing `children` field in `ExtractionResult` struct literal** (`crates/kreuzberg/src/extractors/pst.rs`, ~line 92)

The `ExtractionResult` struct was recently extended with `children: Option<Vec<ArchiveEntry>>` for recursive archive support. The `PstExtractor::extract_sync` method constructs `ExtractionResult` manually but omits this field, which breaks compilation across all targets. Fix: add `children: None,` to the struct literal.

### Medium

1. **Binary HTML body silently returns `None`** (`crates/kreuzberg/src/extraction/pst.rs`, ~lines 183-184) — Property tag 0x1013 (`PR_HTML`) is typically stored as binary (type 0x0102), not a string. `get_str_prop` only matches `PropertyValue::String8` and `PropertyValue::Unicode`, so PST files with only HTML body content will produce empty `cleaned_text`. The EML/MSG extractor handles HTML; PST should do so consistently.

2. **Silent skip on failed messages with no warning** (`crates/kreuzberg/src/extraction/pst.rs`, ~lines 109-122) — Messages and folders that fail to open are silently skipped via `continue`. Other extractors push `ProcessingWarning` entries so callers know data was skipped. Consider doing the same here.

3. **Integration tests skip silently when fixture is missing** (`crates/kreuzberg/tests/pst_integration.rs`, lines 43, 58, 80) — Three tests do `if !pst_path.exists() { return; }`, which means they pass vacuously if the fixture is absent. If the file is deleted or a path changes, tests become silent no-ops rather than failing.

### Low

- **`outlook-pst` crate is not widely used** — since it parses untrusted binary data from email archives, confirm `cargo audit` passes before merge.
- **Manual `days_to_ymd` Gregorian algorithm** instead of existing `chrono` dep; `(invalid timestamp: N)` format is inconsistent with ISO-8601 used elsewhere — consider returning `None` and letting the caller decide.

kh3rld:
@Goldziher kindly review the new impl

PR #569, opened 2026-03-24
feat: support per-file timeouts in batch extraction

### LINKED ISSUE #546: feat: Support per-file timeouts in batch extraction
## Problem

`batch_extract_bytes` and `batch_extract_heterogeneous` accept per-file extraction configs via `FileExtractionConfig`, but timeout is applied at the batch level:

```rust
let batch_timeout_secs = prepared.iter().map(|p| p.timeout_secs).max().unwrap_or(600);

let results = timeout(
    Duration::from_secs(batch_timeout_secs),
    KreuzbergExtractor::extract_batch_heterogeneous(items, &batch_config),
).await
```

This means a batch of 9 fast text PDFs (needing 10s timeout) and 1 slow embedding job (needing 900s timeout) will block the entire batch for up to 900s. If the embedding job hangs at 800s, the 9 fast PDFs that finished in 1s are stuck waiting.

The `ExtractionConfig` struct has a single `max_concurrent_extractions` but no per-file timeout. `FileExtractionConfig` controls extraction behavior (OCR, layout, chunking) per file but has no timeout field either.

## What is needed

Per-file timeout support in batch extraction. When processing a heterogeneous batch, each file should be independently cancelable if it exceeds its own timeout, without affecting other files in the batch.

This matters for cloud workloads where a single API request can contain a mix of document types routed to the same worker batch — small text files alongside large scanned PDFs with OCR and embedding. Today the caller (kreuzberg-cloud worker) wraps the entire `batch_extract_heterogeneous` call in a single `tokio::timeout`, which is the coarsest possible granularity.

The per-file timeout could be added as a field on `FileExtractionConfig`, or as a parallel vec alongside the items, or as a new batch config option. The internal `run_timed_extraction` function already spawns each file as an independent task with a semaphore — adding a per-task timeout there would be the natural place.

Files that timeout should return an error in their position in the results vec, while other files continue to completion normally.

Goldziher:
## Summary

- Adds optional per-file timeout mechanism to batch extraction
- Batch-level default via `ExtractionConfig.extraction_timeout_secs`, per-file override via `FileExtractionConfig.timeout_secs`
- Timed-out files produce a `KreuzbergError::Timeout` error result without blocking other files
- Timeout starts after semaphore acquisition (measures extraction time, not queue wait)
- Updated all 10+ language bindings: Python, Node/TypeScript, Ruby, Go, Java, C#, PHP, Elixir, WASM, R
- Added `ExtractionTimeoutError` exception in Python (avoids shadowing builtin `TimeoutError`)
- E2e test fixtures added and tests regenerated across all languages

## Test plan

- [ ] `cargo test --workspace --all-features` passes
- [ ] Batch extraction with no timeout works as before
- [ ] Batch with `extraction_timeout_secs` set — fast files complete, slow files timeout
- [ ] Per-file `timeout_secs` overrides batch default
- [ ] `KreuzbergError::Timeout` produces correct error in each binding layer
- [ ] E2e tests pass across all languages

Closes #546

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- **Per-file batch extraction timeouts** (#546): New `extraction_timeout_secs` on `ExtractionConfig` (batch-level default) and `timeout_secs` on `FileExtractionConfig` (per-file override). Timeouts apply after semaphore acquisition. New `KreuzbergError::Timeout` variant with `elapsed_ms` and `limit_ms` fields. All binding layers updated.
+- **Page-level OCR overrides** (#432): New `force_ocr_pages` option (1-indexed) on both `ExtractionConfig` and `FileExtractionConfig`. Enables selective OCR on specific pages of mixed-quality PDFs while preserving native text on others.
+
 ### Fixed
 
 - **PDF layout engine panic on malformed input** (#544): Replaced the panicking `.expect()` inside the thread-local `LayoutEngine` initializer in `layout_runner.rs` with proper `Result`-based error propagation. A failure to initialise the layout engine now returns a descriptive error instead of crashing the host process via FFI (Python, Node, etc.).
--- a/crates/kreuzberg-py/src/config/types.rs
+++ b/crates/kreuzberg-py/src/config/types.rs
@@ -434,11 +434,13 @@ impl ExtractionConfig {
 
     fn __repr__(&self) -> String {
         format!(
-            "ExtractionConfig(use_cache={}, enable_quality_processing={}, ocr={}, force_ocr={})",
+            "ExtractionConfig(use_cache={}, enable_quality_processing={}, ocr={}, force_ocr={}, extraction_timeout_secs={:?}, force_ocr_pages={:?})",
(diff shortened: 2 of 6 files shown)


Goldziher:
## Self-Review Notes

Core implementation is solid — timeout placement after semaphore acquisition is correct, error propagation through all binding layers is thorough, and the mixed OCR assembly algorithm is well done.

### Issues to address before merge

**1. No test verifies timeout actually cancels extraction**
All timeout tests use `extraction_timeout_secs: 300` against fast documents. Need at least one test confirming `KreuzbergError::Timeout` is returned when timeout fires.

**2. `force_ocr_pages` silently ignored for non-PDF documents**
Add a `tracing::warn!` when this config is set but the document isn't a PDF.

**3. Generator bug — `true &&` in Java e2e test**
`e2e/java/.../ContractTest.java:1357`: `if (true && !Files.exists(documentPath))` — dead code from generator.

**4. Go `*[]uint64` should be `[]uint64`**
Pointer-to-slice is unconventional in Go. Nil `[]uint64` with `omitempty` is idiomatic.

### Non-blocking

- `__repr__` in Python bindings missing new fields
- Nesting >4 levels in `pdf/mod.rs:774-790`
- `extractionTimeoutSecs` is a no-op in WASM but undocumented
- Two independent features in one PR — consider splitting

PR #573, opened 2026-03-24
feat: JSONL extraction + PDF page rendering API (all languages)

tobocop2:
Two features in this PR — JSONL extraction support and PDF page rendering as a public API.

---

## 1. JSONL extraction

JSONL (JSON Lines / NDJSON) is one of the most common formats for streaming structured data — log pipelines, dataset exports, LLM training data, analytics events. It's just one JSON object per line. But right now kreuzberg doesn't recognize `.jsonl` at all:

```python
from kreuzberg import extract_file_sync

result = extract_file_sync("events.jsonl")
# ValidationError: Unknown extension: .jsonl
```

This adds `.jsonl` / `.ndjson` support through the existing `StructuredExtractor`. Each non-empty line is parsed as an independent JSON object, and the output is a pretty-printed JSON array:

```python
result = extract_file_sync("events.jsonl")
print(result.content)
# [
#   { "user": "alice", "action": "login" },
#   { "user": "bob", "action": "purchase" }
# ]
```

Blank lines are skipped. Malformed lines fail with a line-numbered error:

```
ParsingError: Failed to parse JSONL line 3: expected value at line 1 column 1
```

Implementation reuses existing `extract_from_json_value` and `JsonExtractionConfig` — no new types. UTF-8 validated once, then `str::lines()` for zero-copy line splitting. Vec pre-allocated via `memchr` newline counting.

11 unit tests + 5 integration tests.

---

## 2. PDF page rendering API

kreuzberg already has full PDF page rendering internally (for OCR fallback), but it's not exposed publicly. I ran into this building a vision-model OCR pipeline — I had to pull in `pypdfium2` separately just to rasterize pages to PNGs before sending them to a vision LLM. kreuzberg is already using PDFium under the hood, so this is just wiring it up.

New functions across all language bindings:

```python
from kreuzberg import render_pdf_pages, render_pdf_page

# All pages as PNG bytes
pages = render_pdf_pages("document.pdf", dpi=150)
for i, png_bytes in enumerate(pages):
    with open(f"page_{i}.png", "wb") as f:
        f.write(png_bytes)

# Single page
page_0 = render_pdf_page("document.pdf", 0, dpi=200)
```

Rust core adds `render_pdf_to_png_pages()`, `render_pdf_page_to_png()`, and `render_pdf_file_to_png_pages()`. These are then exposed through:

- **C FFI** — `kreuzberg_render_pdf_pages/page` with proper deallocation
- **Python** — `render_pdf_pages/page` via PyO3
- **Node.js** — sync + async via napi-rs
- **PHP** — via ext-php-rs
- **WASM** — takes bytes instead of file paths (browser-compatible)
- **Elixir** — Rustler NIFs on DirtyCpu scheduler
- **Go** — via cgo
- **Java** — via Foreign Function & Memory API
- **C#** — via P/Invoke
- **Ruby** — via native extension
- **R** — via extendr

4 core unit tests + 6 FFI tests.


tobocop2:
this was a mistake, my apologies for the noise

PR #575, opened 2026-03-24
feat: add JSONL/NDJSON extraction support

tobocop2:
JSONL (JSON Lines / NDJSON) is one of the most common formats for streaming structured data such as log pipelines, dataset exports, LLM training data, analytics events. It's just one JSON object per line. But right now kreuzberg doesn't recognize `.jsonl` at all:

```python
from kreuzberg import extract_file_sync

result = extract_file_sync("events.jsonl")
# ValidationError: Unknown extension: .jsonl
```

Since kreuzberg already has JSON support, the parsing code is already there and it is reasonably trivial to implement. 

### What this does

Adds `.jsonl` / `.ndjson` support through the existing `StructuredExtractor`. Each non-empty line is parsed as an independent JSON object, and the output is a pretty-printed JSON array of all parsed objects.

```python
from kreuzberg import extract_file_sync

# events.jsonl:
# {"user": "alice", "action": "login"}
# {"user": "bob", "action": "purchase", "item": "widget"}

result = extract_file_sync("events.jsonl")
print(result.content)
# [
#   {
#     "user": "alice",
#     "action": "login"
#   },
#   {
#     "user": "bob",
#     "action": "purchase",
#     "item": "widget"
#   }
# ]

print(result.metadata.additional["data_format"])
# "jsonl"
```

Blank lines and whitespace-only lines are skipped. Malformed lines fail with a line-numbered error so you know exactly which line is broken:

```
ParsingError: Failed to parse JSONL line 3: expected value at line 1 column 1
```

### Implementation

- `parse_jsonl()` validates UTF-8 once on the whole slice, then iterates `str::lines()` — zero-copy line splitting, no extra allocations for the input
- Pre-allocates the output vec using `memchr` newline counting
- Reuses existing `extract_from_json_value` and `JsonExtractionConfig` — no new types needed
- MIME types: `application/x-ndjson` (primary), `application/jsonl` and `application/x-jsonlines` (aliases)
- Extensions: `.jsonl`, `.ndjson`

### Tests

23 total (18 unit + 5 integration):

- **Happy path**: simple two-line, single line, multiple lines, file extraction, bytes extraction
- **Blank/whitespace handling**: blank lines skipped, only-blank-lines produces empty array, whitespace-only lines (tabs/spaces) skipped
- **Line endings**: Windows `\r\n` handled correctly
- **Error paths**: malformed JSON line includes 1-indexed line number in error, invalid UTF-8 bytes rejected upfront
- **Value types**: bare JSON scalars (numbers, strings, booleans, null), arrays as top-level values, empty objects `{}`
- **Content preservation**: Unicode (multi-byte chars, emoji), deeply nested JSON per line
- **Metadata**: text field extraction, last-writer-wins for duplicate keys across lines, format field set to "jsonl"
- **Scale**: 1000-line stress test exercising memchr pre-allocation


tobocop2:
> ## Review
> 
> Good implementation — clean integration into the existing `StructuredExtractor` pattern, solid test coverage (CRLF, unicode, stress tests), and proper use of `memchr` for pre-allocation.
> 
> Two items need fixing before merge:
> ### Blocking
> 
>     1. **Silent error swallow in `parse_jsonl`** (`crates/kreuzberg/src/extraction/structured.rs`)
>        `serde_json::to_string_pretty(...).unwrap_or_else(...)` silently returns raw bytes on serialization failure instead of propagating an error. This violates the project's error handling policy. Either use `.map_err(|e| KreuzbergError::parsing(...))?` or add a comment explaining why this is infallible (it is in practice, since `Value::Array` of parsed `Value`s can't fail serialization — but the fallback behavior of returning raw JSONL bytes is wrong regardless).
> 
>     2. **Missing rustdoc on `pub fn parse_jsonl`** (`crates/kreuzberg/src/extraction/structured.rs`)
>        All public functions need rustdoc with at least a description and example per project rules.
> 
> 
> ### Non-blocking (can be follow-up)
> 
>     * `line_count_estimate` is off-by-one for files without trailing newline — use `.count().saturating_add(1)`
> 
>     * `test_parse_jsonl_array_values_per_line` has a weak assertion (`|| result.content.contains("1")` matches almost anything)
> 
>     * Unused `mod helpers;` import in integration test

Thanks, just fixed!

PR #608, opened 2026-03-29
fix(docs): correct R metadata snippet keys

v-tan:
R metadata snippet used incorrect keys: `total_pages` → `page_count`, `author` → `created_by` to match the actual flat metadata schema.

Closes KZB-152


pratik-mahalle (approved):
Looks good!

PR #634, opened 2026-04-03
fix: Use environment variables for hf_hub::ApiBuilder construction

ishbir:
Using `ApiBuilder::new()` discards the HF_HOME and HF_ENDPOINT environment variables which are documented in various places as being supported. Switch to `ApiBuilder::from_env()` to fix this.

Without this, you would see the following warnings when running as userid 1000 on Kubernetes:
```
Model download failed: Failed to download 'rtdetr/model.onnx' from Kreuzberg/layout-models: I/O error Permission denied (os error 13)
```
And it would change to read only file system error if the root filesystem were read only.


Goldziher:
thanks for this. will be released in the next patch

PR #660, opened 2026-04-06
ci: add HOL skill-publish validate workflow

internet-dot:
I opened this as a small validate-only check for the skill metadata in the repo.

A couple of repo-specific details I checked first:
- Extensible architecture – Plugin system for custom OCR backends, validators, post-processors, document extractors, and renderers
- Flexible deployment – Use as library, CLI tool, REST API server, or MCP server
- CLI – Cross-platform binary, batch processing, MCP server mode

This adds one workflow for `.` and leaves the runtime code alone.
It only runs the schema and trust checks for the skill metadata in validate mode.

If you would rather place the workflow under a different filename or point it at a different skill directory, I can adjust the branch.

### DIFF
--- a/.github/workflows/hol-skill-validate.yml
+++ b/.github/workflows/hol-skill-validate.yml
@@ -11,7 +11,7 @@ jobs:
     steps:
       - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
       - name: Validate skill package
-        uses: hashgraph-online/skill-publish@8f3c91d7d4cde7b104549e5f0cccdbb4cd7ee58d # v1
+        uses: hashgraph-online/skill-publish@9c187f00d6706cb0c27aef4eb58f1a8869e9d6db # v1
         with:
           mode: validate
           skill-dir: .


Goldziher:
Hi there. @internet-dot -- why do we need this?

internet-dot:
Thanks for calling that out.

This branch only adds the HOL skill-publish validate workflow and leaves runtime code unchanged.

If you want it reshaped to fit the repository's existing CI layout or naming, I can update it.

Goldziher:
> Thanks for calling that out.
> 
> This branch only adds the HOL skill-publish validate workflow and leaves runtime code unchanged.
> 
> If you want it reshaped to fit the repository's existing CI layout or naming, I can update it.

I mean why do we need to publish a skill to HOL? What would that give us? 

Are you related to HOL? If so, what's the benefit of using this? Who is using this?

internet-dot:
Fair question.

This would not publish anything to HOL on its own.

It only adds a validate-only check for the existing skill metadata and leaves runtime code unchanged.

The point is optional compatibility and discoverability later on, so schema or trust issues show up in CI before merge instead of later.

Yes, this outreach is coming from hashgraph-online / HOL.

If that compatibility path is not useful here, no problem - I can close the PR instead of pushing it further.

PR #726, opened 2026-04-16
fix: prevent doubled OCR content and page mismatch in image extractor

### LINKED ISSUE #706: bug: top-level content and pages[0].content are different OCR renderings of the same page (Tesseract)
## Summary

When using the Tesseract backend, `result.content` and `result.pages[0].content` contain the same page rendered in two different ways — `content` uses HOCR-based coherent text reconstruction (789 chars) while `pages[0].content` contains a raw word-by-word dump (422 chars). They are not equal and do not concatenate to form each other.

## Steps to Reproduce

```bash
kreuzberg extract image.png --config-json '{"ocr":{"backend":"tesseract"}}' --format json \
  | python3 -c "
import sys, json
d = json.load(sys.stdin)
print('content len:    ', len(d['content']))
print('pages[0] len:   ', len(d['pages'][0]['content']))
print('equal:          ', d['content'].strip() == d['pages'][0]['content'].strip())
print()
print('content[:80]:   ', repr(d['content'][:80]))
print('pages[0][:80]:  ', repr(d['pages'][0]['content'][:80]))
"
```

Output:
```
content len:     789
pages[0] len:    422
equal:           False

content[:80]:    'Sales Report 2024\n\nThis report contains quarterly sales data for our products...'
pages[0][:80]:   'Sales\n\nReport\n\n2024\n\nThis\n\nreport\n\ncontains\n\nquarterly\n\nsales\n\ndata\n\nfor\n\nour...'
```

**Expected:** `content` == concatenation of `pages[*].content` (or at minimum, they represent the same text).

**Actual:** `content` is a coherent HOCR-rendered string. `pages[0].content` is a word-by-word dump with `\n\n` between every token — a different rendering pipeline of the same OCR pass.

## Environment

- kreuzberg version: `4.8.4` (built from source)
- OS/arch: Linux x86_64 (Fedora 41)
- Backend: `tesseract`
- Test file: any multi-word image (`test_documents/images/complex_document.png`)

## Scope / Impact

Any consumer that reads `pages[*].content` to reconstruct the document (e.g. for chunking, per-page search, or page-range extraction) receives fragmented word-by-word text instead of coherent sentences. The top-level `content` is usable but `pages` structure is broken for Tesseract.

## Suspected Area

`crates/kreuzberg/src/extractors/image.rs` — the step that assigns `doc.pages` after Tesseract runs. The page content is likely being set from a different intermediate representation (raw word list) than what the top-level `content` field uses (the HOCR-rendered output).

## Proposed Next Diagnostic

Compare what is assigned to `doc.pages[0].content` vs what is assigned to `doc.content` after a Tesseract run in `extractors/image.rs`. Verify whether the page content should be the HOCR rendering or the element-level word list.

kh3rld:
problem

  When extracting images with OCR enabled, ExtractionResult.content contained the text twice: once as a coherent HOCR-rendered string, and again as a raw word-by-word dump appended at the end. Requesting pages extraction made it worse,. pages[0].content reflected the word-dump rather than the readable text.

  Root cause: inject_ocr_elements_from_vec pushed every OcrElement into InternalDocument::elements as an ElementKind::OcrText. The rendering pipeline (render_plain) then iterated those elements and appended each raw word token back into content, on top of the already-rendered HOCR string.

  fix

  - Added InternalDocument::prebuilt_ocr_elements - a direct slot for OCR element lists that bypasses the rendering pipeline entirely.
  - Added InternalDocument::prebuilt_pages population in the image extractor,.. the coherent HOCR string is placed there directly, bypassing the word-grouped fallback in build_pages.
  - derive_extraction_result now prefers prebuilt_ocr_elements before falling back to build_ocr_elements, so ExtractionResult::ocr_elements is still correctly populated.
  - Removed the now-dead inject_ocr_elements_from_vec function.

  Applies to both the native (full TIFF multi-frame) path and the WASM fallback path in the image extractor.

  tests

  Added tests/issue_706_tesseract_page_content_mismatch.rs with two regression tests:
  - test_ocr_content_not_doubled ,. asserts content word count is ≤ 1.3× pages[0].content word count, and that the page text does not appear self-concatenated in content.
  - test_ocr_page_content_matches_top_level_content,.. asserts the same ratio holds for a second fixture image.
  
  fix #706


crates/kreuzberg/src/extractors/image.rs
```diff
@@ -73,9 +73,39 @@ impl ImageExtractor {
             // Build InternalDocument from OCR text
             let mut doc = build_image_internal_document(Some(&ocr_extraction_result.content), None);
             doc.metadata = ocr_metadata;
-            // Propagate OCR elements from the backend result into the InternalDocument
-            // so that derive_extraction_result can populate ExtractionResult::ocr_elements.
-            inject_ocr_elements_from_vec(&mut doc, ocr_elements);
+
+            // Fix #706 (part 1 of 2): store OCR elements directly on the document rather
```
Goldziher: These comments are excessive and unnecessary

crates/kreuzberg/src/extractors/image.rs
```diff
@@ -73,9 +73,39 @@ impl ImageExtractor {
...
+            // than injecting them as OcrText InternalElements. Injecting them as elements
+            // caused render_plain to append each raw word token to the top-level `content`,
+            // producing a doubled output (coherent HOCR text + word-by-word dump).
+            // prebuilt_ocr_elements bypasses the element rendering pipeline entirely while
+            // still populating ExtractionResult::ocr_elements via derive_extraction_result.
+            doc.prebuilt_ocr_elements = ocr_elements;
+
+            // Fix #706 (part 2 of 2): use the coherent HOCR-rendered string for
```
Goldziher: Same

Goldziher (approved):

PR #761, opened 2026-04-20
feat: expose extraction_method on ExtractionResult

### LINKED ISSUE #751: feat: expose OCR/native signal on ExtractionResult (the value already exists internally, just isn't surfaced)
## Summary

`ExtractionResult` tells callers *what* text came out, but not *how*. Downstream code that wants to differentiate OCR'd text from native extraction (UI badges, telemetry, per-doc quality routing) has no upstream signal and has to either run two extraction passes and compare, or heuristically inspect content.

The reason this feels frustrating is that the signal **already exists inside the extractor** — it's just not propagated out.

## The signal already exists

In [`crates/kreuzberg/src/extractors/pdf/mod.rs:553`](https://github.com/kreuzberg-dev/kreuzberg/blob/v4.8.6/crates/kreuzberg/src/extractors/pdf/mod.rs#L553) the PDF extractor computes:

```rust
let (text, used_ocr) = if config.effective_disable_ocr() {
    (native_text, false)
} else if config.force_ocr {
    // ...
    (ocr_text, true)
} else if let Some(ref ocr_pages) = config.force_ocr_pages {
    // mixed native + per-page OCR
    (mixed, true)
} else if let Some(ocr_config) = config.ocr.as_ref() {
    // per-page threshold evaluation
    // ...
}
```

`used_ocr` is already consumed internally at [line 696](https://github.com/kreuzberg-dev/kreuzberg/blob/v4.8.6/crates/kreuzberg/src/extractors/pdf/mod.rs#L696) (`use_structured_doc = !used_ocr && pre_rendered_doc.is_some()`) — it just never reaches `ExtractionResult`.

## What I observe on 4.8.6

```python
# Native (scanned PDF, no text layer extracted)
r1 = await extract_file("scan.pdf", config=ExtractionConfig(ocr=None))
# Forced OCR on the same file
r2 = await extract_file("scan.pdf", config=ExtractionConfig(
    ocr=OcrConfig(backend="tesseract", language="eng"), force_ocr=True
))

r1.metadata.keys() == r2.metadata.keys()  # True — identical
# No ocr_used, no extraction_method, no marker of any kind.
# r1.pages and r2.pages are both None for PDFs (only images get pages[] via #723).
```

## Proposed API

Three shapes, in order of preference — happy to go with whichever you prefer:

1. **`extraction_method: Option<ExtractionMethod>` enum** with variants `Native`, `Ocr`, `Mixed`. Handles the `force_ocr_pages` case cleanly (today that path returns `(mixed, true)` and collapses partial OCR into full OCR — `Mixed` would preserve that info). Extensible for future VLM / hybrid paths.
2. **`ocr_used: Option<bool>`**. Simplest, loses the `Mixed` distinction.
3. **Per-page `pages[i].extraction_method`**. Most precise, but depends on `pages[]` getting populated for PDFs first (currently `None`; [#723](https://github.com/kreuzberg-dev/kreuzberg/pull/723) only did images). Could be a follow-up after (1) or (2).

Happy to send a PR once you've picked a shape — just drop a comment.

## Environment

Observed on `kreuzberg==4.8.6`, `python:3.14.3-slim` (aarch64).

alexanderkreidich:
## Summary
- add a first-class `extraction_method` field to `ExtractionResult` with `native`, `ocr`, and `mixed` values
- propagate the existing PDF OCR/native decision into the public result while keeping the internal metadata handoff intact
- expose the new field through the Rust, Python, Node, PHP, and TypeScript result surfaces

Closes #751.

## Verification
- `cargo check -p kreuzberg --features 'pdf bundled-pdfium'`
- `cargo test -p kreuzberg --lib extraction_method --features 'pdf bundled-pdfium' -- --nocapture`

## Notes
- `cargo check -p kreuzberg-py -p kreuzberg-node` is blocked in this environment by the existing `kreuzberg-tesseract` arm64 build issue (`cpuid.h`/`simddetect.cpp` x86 SIMD detection)
- `cargo check -p kreuzberg-php` is also blocked locally because `ext-php-rs` requires a `php` executable on `PATH`

### DIFF
--- a/crates/kreuzberg/src/extraction/derive.rs
+++ b/crates/kreuzberg/src/extraction/derive.rs
@@ -16,7 +16,7 @@ use ahash::AHashMap;
 use crate::types::document_structure::{
     DocumentNode, DocumentRelationship, DocumentStructure, GridCell, NodeContent, NodeId, NodeIndex, TableGrid,
 };
-use crate::types::extraction::ExtractionResult;
+use crate::types::extraction::{ExtractionMethod, ExtractionResult};
 use crate::types::internal::{ElementKind, InternalDocument, InternalElement, RelationshipTarget};
 use crate::types::ocr_elements::{OcrConfidence, OcrElement};
 use crate::types::page::PageContent;
@@ -654,6 +654,13 @@ pub fn derive_extraction_result(
         _ => None,
     };
 
+    let extraction_method = doc
+        .metadata
+        .additional
+        .get("extraction_method")
+        .and_then(serde_json::Value::as_str)
+        .and_then(ExtractionMethod::from_metadata_value);
+
     tracing::debug!(
         content_length = content.len(),
         has_document_structure = document.is_some(),
@@ -663,6 +670,7 @@ pub fn derive_extraction_result(
         content,
         mime_type,
         metadata: doc.metadata,
+        extraction_method,
         tables: doc.tables,
         images,
         pages,
@@ -1059,5 +1067,31 @@ mod tests {
         let result = derive_extraction_result(doc, true, crate::core::config::OutputFormat::Plain);
         let ds = result.document.unwrap();
         assert_eq!(ds.source_format.as_deref(), Some("epub"));
+    }
+
+    #[test]
+    fn test_derive_extraction_result_promotes_extraction_method() {
+        let mut doc = make_doc("pdf");
+        doc.metadata.additional.insert(
+            Cow::Borrowed("extraction_method"),
+            serde_json::Value::String("mixed".to_string()),
+        );
+        doc.push_element(InternalElement::text(ElementKind::Paragraph, "Hello world.", 0));
+
+        let result = derive_extraction_result(doc, false, crate::core::config::OutputFormat::Plain);
+        assert_eq!(result.extraction_method, Some(ExtractionMethod::Mixed));
+    }
+
+    #[test]
+    fn test_derive_extraction_result_ignores_unknown_extraction_method() {
+        let mut doc = make_doc("pdf");
+        doc.metadata.additional.insert(
+            Cow::Borrowed("extraction_method"),
+            serde_json::Value::String("native_ole".to_string()),
+        );
+        doc.push_element(InternalElement::text(ElementKind::Paragraph, "Hello world.", 0));
+
+        let result = derive_extraction_result(doc, false, crate::core::config::OutputFormat::Plain);
+        assert_eq!(result.extraction_method, None);
     }
 }
--- a/e2e/typescript/tests/helpers.ts
+++ b/e2e/typescript/tests/helpers.ts
@@ -629,6 +629,17 @@ export const assertions = {
 		}
 	},
 
+	assertExtractionMethod(result: ExtractionResult, expected: string): void {
+		const metadata = isPlainRecord(result.metadata) ? (result.metadata as PlainRecord) : null;
+		const additional = metadata && isPlainRecord(metadata.additional) ? (metadata.additional as PlainRecord) : null;
+		const actual =
+			(result as unknown as PlainRecord).extractionMethod ??
+			(result as unknown as PlainRecord).extraction_method ??
+			metadata?.extraction_method ??
+			additional?.extraction_method;
+		expect(actual).toBe(expected);
+	},
+
 	assertContentNotEmpty(result: ExtractionResult): void {
 		expect(result.content).toBeDefined();
 		expect(result.content.trim().length).toBeGreaterThan(0);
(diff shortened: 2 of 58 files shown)


Goldziher:
Hi there. 

Thanks for your efforts. 

1. Use an enum. 
2. Don't modify non rust code manually.
3. Add an e2e fixture.

Goldziher (changes requested):
Please read my prev comment

e2e/typescript/tests/helpers.ts
```diff
@@ -629,6 +629,17 @@ export const assertions = {
 		}
```
kh3rld, line 629: Instead of casting to `unknown` here, we should ideally use the typed field once it's re-added to the TS interface.

crates/kreuzberg/src/extraction/derive.rs
```diff
@@ -654,6 +654,13 @@ pub fn derive_extraction_result(
         _ => None,
```
kh3rld, line 654: This promotion logic is clean. Let's make sure it's also documented in our API schema that this field is now the preferred way to access this information.

kh3rld (changes requested):
Great work on this high-impact PR! Exposing the extraction strategy is a frequent request and your implementation across the PDF pipeline is solid.

Since this changes our public API contract, we need to ensure full parity and type safety across all bindings before merging. Patch 2 appeared to revert some manual type definitions; we should restore those so that typed consumers (TS/PHP) don't lose visibility of the new field.

### Why this matters
Consistency across bindings builds user trust. If the field exists at runtime but isn't in the static types, it breaks IDE autocompletion and CI for our users.

### Requested Changes
- [ ] **TypeScript**: Re-add `extractionMethod?: "native" | "ocr" | "mixed";` to `packages/typescript/core/src/types/results.ts`.
- [ ] **PHP**: Re-add `@property-read` and constructor property for `$extractionMethod` in `packages/php/src/Types/ExtractionResult.php`.
- [ ] **Image Extractor**: Update `crates/kreuzberg/src/extractors/image.rs` to set `extraction_method = ocr`. For other formats (Office, HTML), we should decide if we default to `native` or leave as `None` (open to a follow-up issue if preferred).
- [ ] **Tests**: Update E2E assertions to prefer the first-class `extraction_method` field instead of indexing into `metadata.additional` (see inline comment).

### Verification
Since you are facing local binding constraints, please ensure the following pass in CI:
*   **Rust tests**: `cargo test -p kreuzberg --lib extractors::pdf::tests`
*   **E2E validation**: `cargo run -p e2e-generator -- --validate-fixtures`
*   **Binding CI**: Check the `bindings-python`, `bindings-node`, and `bindings-php` jobs.

Please also add a brief entry to `CHANGELOG.md` under the "Added" section. Looking forward to the next revision!

kh3rld (changes requested):
.

Goldziher (changes requested):
## Review

The feature request (#751) is valid — exposing whether OCR or native extraction was used is useful for callers. However, there are naming collisions, pipeline wiring concerns, and coverage gaps.

### Critical

**Enum name collision.** `ExtractionMethod` already exists as an internal enum in `kreuzberg-pdfium-render` (`pdf/document/page/extraction.rs`) with variants `StructureTree` and `Heuristic`. The public enum must use a distinct name — suggest `ContentExtractionMethod` or `TextExtractionSource`.

**Must flow through `InternalDocument`.** The pipeline is: extractor → `InternalDocument` → `derive_extraction_result()` → `ExtractionResult`. A field on `ExtractionResult` without a corresponding field on `InternalDocument` cannot be set by extractors. Add `pub extraction_method: Option<ExtractionMethod>` to `InternalDocument` and propagate it in `derive_extraction_result()` at `extraction/derive.rs:643`.

**Must be set in ALL extractors.** Currently the PR likely only sets it in a subset. Every `DocumentExtractor` implementation must set the field:
- PDF (pdfium path): `Native` when text layer exists, `Ocr` when OCR fallback triggered
- Image extractor: always `Ocr`
- Office formats (docx, xlsx, pptx, odt): `Native`
- HTML, markdown, text, code: `Native`
- Archive: depends on children (could be `None`)

If any extractor leaves it `None`, callers can't distinguish "native" from "not set."

### Warning

**Backward-compatible serialization.** The field must be `Option<ExtractionMethod>` with `#[serde(skip_serializing_if = "Option::is_none", default)]`. Without `default`, deserialization of any cached/persisted `ExtractionResult` from before this change will fail.

**FFI-friendliness.** Per project policy, public enums need `#[repr(C)]` or equivalent for the FFI boundary. The FFI layer needs an accessor function (e.g., `kreuzberg_extraction_result_get_method() -> i32`).

**Generated bindings.** With +1166/-160 across 55 files, changes to `packages/*/` and binding crates must come from `alef generate`, not manual edits. The freshness check must pass: `alef generate && git diff --exit-code packages/ crates/kreuzberg-node/ crates/kreuzberg-wasm/ crates/kreuzberg-ffi/`

**Remove metadata string duplication.** The current `extraction_method = "native_ole"` in `extractors/doc.rs` and `extractors/ppt.rs` (stored in `metadata.additional`) should be removed once the typed field exists — two sources of truth creates inconsistency.

### Missing

- Integration tests: PDF with text layer → `Native`, scanned PDF → `Ocr`, `.docx` → `Native`, image → `Ocr`.
- `alef.toml` update if the new enum needs to be in `[crate].sources` for binding codegen.

Goldziher:
Squash-merged into main as 4860881d2; credited in CHANGELOG [Unreleased].

Issue #759, opened 2026-04-19
feat: in-process embedding backend plugin for ChunkingConfig


tobocop2:
## Problem

`ChunkingConfig.embedding` today supports only:
- `Preset` / `Custom` — ONNX models kreuzberg downloads from HuggingFace and runs in its own ORT session
- `Llm` — liter-llm over HTTP to an OpenAI-compatible endpoint (`llamacpp/`, `ollama/`, `lmstudio/`, `vllm/`, `llamafile/`)

Callers who already have an embedder loaded **in-process** (e.g. `llama-cpp-python`, `sentence-transformers`, a tuned ONNX model) can't reuse it. They're forced to either let kreuzberg duplicate the model load or spin up a sidecar HTTP server just so kreuzberg can round-trip back into the same machine.

## Proposed solution

Add an **`EmbeddingModelType::Plugin { name }`** variant and an **`EmbeddingBackend`** plugin trait, mirroring the existing `OcrBackend` / `PostProcessor` / `Validator` pattern.

Trait contract:
- `name() -> &str`
- `dimensions() -> usize` (stable for the lifetime of the registration)
- `async fn embed(&[&str]) -> Result<Vec<Vec<f32>>>` — async to match the existing async-trait convention used by `OcrBackend::process_image` and `DocumentExtractor::extract_bytes`
- Optional `fn max_batch_size(&self) -> Option<usize>` hint so dispatcher can chunk oversized batches

Host-language bridges expose `register_embedding_backend(backend)` (PyO3 for Python, napi-rs for Node, Rustler for Elixir, extendr for R, magnus for Ruby, ext-php-rs for PHP, wasm-bindgen for WASM, C FFI for Java/Go/C#). The dispatcher in `embed_texts()` routes the `Plugin` variant to the registered backend with validation + a configurable timeout before returning.

### Dispatcher behaviour

- Validates response shape (length + per-vector dimension) before returning; shape mismatches surface as `KreuzbergError::Validation` with the specific offender named.
- Enforces an `EmbeddingConfig::max_embed_duration: Option<Duration>` timeout (default `Some(60s)` — overridable) to guard against hung host-language callbacks.
- Catches panics from the host-language call and surfaces them as `KreuzbergError::Plugin`. Trait requires `UnwindSafe` (or equivalent documented contract) so the catch is sound.
- Read-lock on the registry is released before invoking `embed()` so registration work isn't blocked by in-flight inference.

### Registration/shutdown semantics

[...]

PR #768, opened 2026-04-20
feat: in-process embedding backend plugin

tobocop2:
Refs #759.

## Why

Kreuzberg's embedding story has been "either download an ONNX model or call an HTTP provider." That breaks for a large class of callers: hosts that already have an embedder loaded and running — `llama-cpp-python`, `sentence-transformers`, a tuned ONNX session, something behind a private gRPC that isn't OpenAI-compatible. Asking them to spin up a sidecar HTTP server or to have kreuzberg download and load the same model twice (once by them, once by us) is the wrong shape.

## How

A third `EmbeddingModelType::Plugin { name }` variant, sibling to `Preset`, `Custom`, and `Llm`. Everything about the new variant goes through a plugin trait, the same pattern the codebase already uses for `OcrBackend` / `PostProcessor` / `Validator`:

1. Host registers once at startup via `kreuzberg::plugins::register_embedding_backend(Arc::new(MyEmbedder))`. The registry calls `initialize()`, then `dimensions()`, caches the reported dimension, and stores the `Arc<dyn EmbeddingBackend>`.
2. Config references the backend by name: `EmbeddingModelType::Plugin { name: "my-embedder" }`. Same serde shape everywhere — CLI, MCP, REST, env var, in-code config all round-trip it.
3. When the dispatcher sees the `Plugin` variant, it clones the `Arc` out of the registry under a short read lock, releases the lock, calls `embed(texts)` wrapped in `tokio::time::timeout` (default 60s, configurable via `max_embed_duration_secs`, `None` disables), then validates the returned shape against the cached dimension before handing back to downstream consumers.

The ONNX preset/custom path and the HTTP `Llm` path are untouched. `Plugin` is purely additive.

**Why cache `dimensions()`** — a buggy backend that drifts its reported dimension mid-session would otherwise sneak past shape validation silently. Registration captures it once; dispatch uses the captured value; a real drift is a registration-time error.

**Why the timeout** — host-language callbacks can wedge (Python GIL deadlocks, CUDA OOM retry loops, etc). Without a cap, a single bad backend can stall the whole chunking pipeline. 60s is conservative; large-batch users bump it.

## Trait shape

```rust
#[async_trait]
pub trait EmbeddingBackend: Plugin {
    fn dimensions(&self) -> usize;
    async fn embed(&self, texts: Vec<String>) -> Result<Vec<Vec<f32>>>;
}
```

`Vec<String>` rather than `&[&str]` — alef's `trait_bridge` codegen emits owned types across the FFI boundary, so the trait signature matches what the generated bridges expect. The batch is materialised once per call; the host-language hop already copies regardless.

Lifecycle is documented on the trait: `initialize()` runs before `dimensions()`, `dimensions()` is captured once at registration time, `embed()` may run concurrently with `shutdown()`, and backends must tolerate in-flight calls continuing after shutdown is requested (the `Arc` keeps the backend alive).

## Entry points

Every boundary that already understood embedding config now accepts the `Plugin` variant:

- **Rust library**: `kreuzberg::plugins::register_embedding_backend(Arc::new(backend))`
- **CLI**: `kreuzberg embed --provider plugin --plugin my-embedder`
- **MCP**: `embed_text` tool's `embedding_plugin` parameter (takes precedence over `model` / `preset` when set)
- **Env var**: `KREUZBERG_EMBEDDING_PLUGIN_NAME=my-embedder` (takes precedence over `KREUZBERG_VLM_EMBEDDING_MODEL`)
- **REST**: `EmbedRequest.config.model` with `{"type":"plugin","name":"…"}`

## Per-binding status

Regenerated with alef v0.6.1 (bumped from main's v0.6.0 — v0.6.1 carries the codegen fixes needed for plugin-pattern trait bridges across all backends).

| Binding | `registerEmbeddingBackend` | Config round-trip | Notes |
|---|---|---|---|
| Python (PyO3) | ✅ | ✅ | Full support including fork-safety guidance in docs |
| Node (napi-rs) | ✅ | ✅ | |
| PHP (ext-php-rs) | ✅ | ✅ | |
| WASM (wasm-bindgen) | ✅ | ✅ | |
| Ruby (magnus) | ✅ | ✅ | |
| Elixir (Rustler) | ✅ | ✅ | |
| R (extendr) | ✅ | ✅ | |
| Go (cgo) | ✅ | ✅ | Via C FFI vtable |
| C# (P/Invoke) | ✅ | ✅ | Via C FFI vtable |
| C FFI | ✅ | ✅ | `kreuzberg_register_embedding_backend` |
| Java (Panama) | ⏳ | ✅ | Alef's Java backend doesn't emit plugin bridges yet — matches the existing `OcrBackend` gap on Java |

All hosts can round-trip the `Plugin` config variant through JSON/serde today; host-side registration is lit up for 10 of 11 bindings.

## Docs

The advanced guide gets a new "In-Process Embedding Backends (Plugin Variant)" section covering the four entry points and a fork-safety warning for Python users running under `multiprocessing`, gunicorn prefork, or Celery prefork — native-backed embedders like `llama-cpp-python` aren't fork-safe and need re-registration in children via `os.register_at_fork`.

## Non-goals

- Pulling GGUF / llama.cpp into kreuzberg itself. The whole point is that the caller owns the model.
- Replacing the `Llm` HTTP variant. Both coexist — `Llm` for sidecar servers, `Plugin` for in-process callbacks.


tobocop2:
@Goldziher — rebased onto latest `feat/alef-migration` (874aa1fd9) and regenerated with alef v0.5.7. Single commit `b8a6e2601`.


**Two upstream alef v0.5.7 quirks I hit — both are also present on this branch's base (`feat/alef-migration`), so not regressions:**

1. **`alef verify` false-positive after `alef all --clean`.** Immediately after a clean regen, `alef verify` still reports several files as stale (`[python] crates/kreuzberg-py/src/lib.rs`, `[wasm] crates/kreuzberg-wasm/src/lib.rs`, `[stubs] packages/python/kreuzberg/_kreuzberg.pyi`, and a handful of Java/C# files). Reproduces on a clean checkout of `feat/alef-migration` alone, so it looks like a non-determinism in alef's generator vs. its stored input-hash comparison.
2. **FFI `build.rs` cbindgen parse error.** `crates/kreuzberg-ffi/src/lib.rs` around line 27910 has a docstring block that got converted to raw markdown without the `/* */` wrapping — cbindgen panics with `cannot parse string into token stream`. The broken chunk is inside `KreuzbergOcrBackendVTable::process_image_fn`'s doc (`# Arguments`, `* \`image_bytes\` - Raw image data...`). Looks like the rustdoc `///` → C doc-comment conversion strips the prefix but doesn't re-wrap.

Goldziher:
@tobocop2 please rebase. i will approve -- so you can merge when rebase is done.

Goldziher (approved):

tobocop2:
> @tobocop2 please rebase. i will approve -- so you can merge when rebase is done.

done

Goldziher (changes requested):
## Review

The feature direction is sound — in-process embedding backends follow the existing plugin system pattern (OCR, Extractor, PostProcessor, Validator registries). The PR is large (+9855/-2848, 154 files) but most should be alef-generated binding code. Key architectural concerns below.

### Critical

**Config serializability.** The new variant must be `EmbeddingModelType::Plugin { name: String }` where `name` is the registered plugin name resolved at call time from the registry — NOT an embedded `Arc<dyn EmbeddingPlugin>`. Embedding a trait object in the config enum breaks JSON serialization, config file round-trips, and binding codegen. `alef.toml` does not exclude `EmbeddingModelType` — any non-serializable variant breaks all 11 language bindings.

**Dispatch must be wired.** Both `embed_texts` (`embeddings/mod.rs:762`) and `embed_texts_async` dispatch on `EmbeddingModelType` with a `match`. The new `Plugin` variant must have explicit match arms that look up the registry and dispatch into the plugin. Without this, the feature is dead code.

**Thread safety.** The `EmbeddingPlugin` trait method must use `&self`, not `&mut self`. The registry must store `Arc<dyn EmbeddingPlugin + Send + Sync>`. If the trait's `embed()` is sync (acceptable for CPU-bound inference), callers from async contexts must use `spawn_blocking` — document this clearly.

### Warning

**Registry pattern compliance.** Must follow the established pattern:
- `static EMBEDDING_PLUGIN_REGISTRY: LazyLock<Arc<RwLock<EmbeddingPluginRegistry>>>`
- `get_embedding_plugin_registry()` accessor
- `register_embedding_plugin` / `unregister_embedding_plugin` convenience functions
- `pub use` re-exports from `plugins::mod`

**`alef.toml` updates required.** The new trait source file must appear in `[crate].sources`. `EmbeddingPluginRegistry` must be in exclude types. `register_embedding_plugin`/`unregister_embedding_plugin` must be in exclude functions (handled by trait bridge codegen, like `register_ocr_backend`).

**Async boundary.** If `EmbeddingPlugin::embed()` is async, it must use `#[async_trait]`. If sync, the dispatch arm in `embed_texts_async` must wrap it in `spawn_blocking`. Mixing sync/async without this causes deadlocks.

**Global engine cache separation.** The existing `RwLock<AHashMap<String, CachedEngine>>` in `embeddings/mod.rs:73` caches ONNX engines. The plugin registry must NOT share this cache — plugins manage their own backends.

### Missing

- Test: register a mock `EmbeddingPlugin`, create `ChunkingConfig` with `EmbeddingModelType::Plugin { name }`, assert mock is called instead of ONNX.
- Test: `EmbeddingModelType::Plugin { name: "nonexistent" }` returns `KreuzbergError`, not a panic.
- Freshness check: `alef generate && git diff --exit-code packages/ crates/kreuzberg-node/ crates/kreuzberg-wasm/ crates/kreuzberg-ffi/`

tobocop2:
Addressed @Goldziher.

**Critical**

- The Plugin variant is a plain `{ name: String }` — fully serializable, no trait object embedded.
- Both sync `embed_texts` and async `embed_texts_async` dispatch the Plugin variant, and both are now exhaustive matches over `EmbeddingModelType` so a future variant fails to compile rather than silently routing to ONNX.
- Trait method takes `&self`. Registry stores `Arc<dyn EmbeddingBackend>`. Async via `#[async_trait]`. The sync entry's runtime requirement is documented on the trait.

**Warning**

- Registry pattern matches the OCR/Validator/PostProcessor pattern: LazyLock-backed singleton, public getter, convenience register/unregister/list/clear functions re-exported from `kreuzberg::plugins`.
- `alef.toml` updates done — source include for the trait file, exclude entries for the registry type and convenience functions, and a `[[trait_bridges]]` block for `EmbeddingBackend` mirroring the `OcrBackend` entry.
- Plugin embeddings keep their own registry; the existing ONNX engine cache is untouched.

**Tests**

- Register a mock backend, dispatch via `EmbeddingModelType::Plugin`, assert the mock is called.
- Unknown plugin name returns `KreuzbergError::Plugin`, not a panic.
- Plus dim/length-mismatch validation, normalization on both sync and async paths, timeout-fires / no-timeout-completes, concurrent-registration stress, and the env-var resolution path.

tobocop2:
@Goldziher heads up — `alef e2e generate` rewrites `e2e/node/package.json` with vitest `^3.0.0`, which conflicts with main's lockfile (bumped to `^4.1.5` by dependabot); alef's e2e scaffold template appears to lag dependabot bumps.

PR #772, opened 2026-04-21
fix(pst): populate attachments and use proper MAPI EntryID hex format

### LINKED ISSUE #764: bug: Extracting data from MS Outlook PST file format excludes EntryID and attachements
Even with the latest changes to "crates/kreuzberg/src/extraction/pst.rs", aiming at EntryID retrieval, I still can't seem to be able to extract an Outlook object with the EntryID by using kreuzberg library in Rust. The returned metadata HashMap never seems to contain an EntryID for each of the contained emails. In addition, kreuzberg does not seem extract the attachements where the Outlook object also contains them.

My use case is using the kreuzberg as the data extraction library for desktop full text search. Up until now I've been using various libraries directly and I was hopping that I would be able to replace all of them with just kreuzberg. With regard to MS Outlook the aim is to also be able to open an Outlook object (e.g. email, calendar item, task item) in MS Outlook upon finding the item via an external federated index. This means using the EntryID in the following way:

"C:\Program Files\Microsoft Office\root\Office16\OUTLOOK.EXE" /select outlook:000000001B58BD4DF3D3784FAB15D7A887F6F53844012000

EntryID used in the example above is from an actual .pst example file (https://github.com/aspose-email/Aspose.Email-Python-Dotnet/blob/master/Examples/Data/Outlook.pst)

Just in case it might be useful, below is a shortened, simplified and self-contained example of how I'm currently using outlook-pst (i.e. directly, not through kreutzberg) to extract an EntryID with just a part of the relevant content (just the subject in the this case):

```Rust
use outlook_pst::{
    ltp::prop_context::PropertyValue::*,
    messaging::{folder::Folder, store::Store},
    ndb::node_id::NodeId,
    open_store,
};

fn main() {
    let store = open_store(std::env::args().nth(1).expect("No path provided")).unwrap();
    let mut record_key = [0u8; 16];

    if let Some(Binary(val)) = store.properties().get(0x0FF9) {
        record_key.iter_mut().zip(val.buffer()).for_each(|(a, b)| *a = *b);
    }

    let root_entry_id = store.properties().make_entry_id(NodeId::from(0x122)).unwrap();
    walk_folder(&*store, &*store.open_folder(&root_entry_id).unwrap(), &record_key);
}

fn walk_folder(store: &dyn Store, folder: &dyn Folder, record_key: &[u8; 16]) {
    if let Some(contents_table) = folder.contents_table() {
        for id in contents_table.rows_matrix().map(|row| u32::from(row.id())) {
            if let Ok(message) = store.properties().make_entry_id(NodeId::from(id)).and_then(|eid| store.open_message(&eid, None)) {

                let subject = match message.properties().get(0x0037) {
                    Some(String8(val)) => val.to_string(),
                    Some(Unicode(val)) => val.to_string(),
                    _ => String::new(),
                };

                let hex_id: String = record_key.iter().chain(&id.to_le_bytes()).map(|b| format!("{b:02X}")).collect();
                println!("00000000{hex_id} | {subject}");
            }
        }
    }

    if let Some(hierarchy_table) = folder.hierarchy_table() {
        for id in hierarchy_table.rows_matrix().map(|row| u32::from(row.id())) {
            if let Ok(subfolder) = store.properties().make_entry_id(NodeId::from(id)).and_then(|eid| store.open_folder(&eid)) {
                walk_folder(store, &*subfolder, record_key);
            }
        }
    }
}
```

kh3rld:
## Summary

Fixes two confirmed kreuzberg bugs reported in issue #764:

- **Attachments always empty**: `extract_message_content()` had `attachments: vec![]` hardcoded and never read from the PST message. Now reads from the attachment table using the same `TableContext` row-iteration approach used for recipients — extracting `PR_ATTACH_LONG_FILENAME` (0x3707), `PR_ATTACH_FILENAME` (0x3704), and `PR_ATTACH_DATA_BINARY` (0x3701) for each attachment.

- **EntryID wrong format**: `metadata["entry_id"]` was set to `format!("{:?}", entry_id)` which produces a Rust `Debug` string like `EntryId { record_key: ..., node_id: ... }`. Outlook expects the standard MAPI EntryID hex format: 4 zero bytes (rgbFlags) + 16-byte record_key + 4-byte node_id (little-endian) = 48 hex chars. Fixed by serialising the fields directly from `entry_id.record_key()` and `u32::from(entry_id.node_id()).to_le_bytes()`.

## Test plan

- [x] Added regression test `test_entry_id_hex_format_issue_764` that verifies the 48-char hex format, correct `00000000` prefix, and absence of Debug-style syntax
- [x] All 8 pst extraction unit tests pass (`cargo test -p kreuzberg --features email,html --lib -- extraction::pst`)
- [x] Clean compile with `--features email,html`

fixes #764

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -156,7 +156,7 @@ repos:
 
   # Shell scripts: formatting and linting
   - repo: https://github.com/scop/pre-commit-shfmt
-    rev: v3.13.1-1
+    rev: v3.14.2-1
     hooks:
       - id: shfmt
         args: ["-w", "-i", "2"]
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -74,6 +74,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - **oxlint path in node e2e lint** — `oxlint --fix typescript` changed to `oxlint --fix .` (was looking for nonexistent `typescript/` directory).
 - **Clippy warnings in benchmark-harness** — `sort_by` replaced with `sort_by_key` + `Reverse`.
 - **#766**: PDF extraction with large numbers of image fragments no longer hangs indefinitely — added `ImageExtractionConfig.max_images_per_page` (default `None`) to cap images processed per page. Batch-level `extraction_timeout_secs` now interrupts blocking pdfium threads at the next inter-page checkpoint via a `CancellationToken`, preventing the timeout from being silently bypassed.
+- **#764**: PST extractor now populates email attachments — `attachments` was hardcoded to an empty list and never read from the message; now reads attachment name, filename, MIME type, size, and binary data via the attachment table. PST entry IDs are now formatted as proper 48-char MAPI hex strings instead of Rust Debug output.
 
 ### Added
 
--- a/crates/kreuzberg/src/extraction/pst.rs
+++ b/crates/kreuzberg/src/extraction/pst.rs
@@ -26,7 +26,7 @@
 //! ```
 
 use crate::error::{KreuzbergError, Result};
-use crate::types::{EmailExtractionResult, ProcessingWarning};
+use crate::types::{EmailAttachment, EmailExtractionResult, ProcessingWarning};
 use std::borrow::Cow;
 use std::collections::HashMap;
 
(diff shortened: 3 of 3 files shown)


crates/kreuzberg/src/llm/vlm_ocr.rs
```diff
@@ -55,7 +55,14 @@ impl OcrBackend for VlmOcrBackend {
         // Detect MIME type from image bytes
         let mime = infer::get(image_bytes).map(|t| t.mime_type()).unwrap_or("image/png");
 
-        let (text, usage) = vlm_ocr(image_bytes, mime, &config.language, vlm_config).await?;
+        let (text, usage) = vlm_ocr(
+            image_bytes,
+            mime,
+            &config.language,
+            vlm_config,
+            config.vlm_prompt.as_deref(),
```
Goldziher: So, this is a separate fix. 

Should be at least mentioned in the changelog.

Goldziher (approved):
This PR does two distinct things. One is not mentioned in the title, description or changelog. 

Please address and then merge

PR #775, opened 2026-04-21
fix(pdf-structure): preserve tagged-block body text and stop mis-tagging numbered headings as lists

### LINKED ISSUE #774: bug: tagged-PDF structure tree drops body text when a block has own text and children
## Summary

On PDFs with a tagged structure tree, markdown output silently drops paragraph body text whenever a block has both its own `text` and nested children. On the same inputs, numbered section headings end up inside a `List` AST node with no `Item` children — panicking comrak's validator in debug builds and emitting malformed markdown in release:

    InvalidChildType { parent: List(...), child: Heading { level: 3 } }
      — comrak-0.52.0/src/cm.rs:29 (gated on #[cfg(debug_assertions)])

Release strips the validator and produces visible corruption. Concrete measurements on a 4-page tagged PDF (produced by Typst 0.14.2):

| output_format | content length |
|---------------|----------------|
| plain         | 9,512 chars    |
| markdown      | 4,169 chars    |

56% of the text is missing from markdown. Visible symptoms:

- Bold lead-ins (`**<title>:**`) survive but their continuation text after the colon is dropped.
- Numbered list items render as empty bullets plus a detached bold paragraph; the `— <body>` suffix is dropped or displaced to a later paragraph.

## Root cause

Two bugs that compound:

**1. `pdf/structure/adapters.rs::flatten_blocks`** — when a block has children, the walker recurses and `continue`s, skipping its own `text`. Tagged `LBody` elements in the trigger PDF look like:

    LBody   text=" — <continuation>"           children=[Paragraph]
      Paragraph  text="<bold lead-in>"

The child Paragraph is emitted; `LBody`'s " — <continuation>" is dropped.

**2. `pdf/structure/content_convert.rs::element_to_paragraph`** — `is_list_item` is set from a text-pattern heuristic (`is_list_prefix_multi_token`) even when the element is already tagged `Heading`. A heading like `"3. Conclusions"` gets both flags set; assembly wraps it in `ListStart`/`ListEnd` around a `Heading`, producing the invalid AST above.

Fixing (1) alone still trips (2); fixing (2) alone still loses text.

## Reproduction

Any tagged PDF whose structure tree has an `LBody` with both `text` and a child `P`, plus numbered section `Heading`s, reproduces both symptoms. Typst 0.14.2 emits this shape on ordinary markdown-to-PDF exports.

The PR includes two synthetic unit tests (in `adapters.rs` and `content_convert.rs`) that construct the exact block tree and content element programmatically — no binary fixture required. Both fail on `main`, pass with the fix.

## Environment

kreuzberg 4.9.1 and current `main` (verified unchanged rendering path between tags). `output_format="markdown"`, default config, no OCR.

## Fix

PR incoming.

dergachoff:
Closes #774. Two compounding bugs in the tagged-PDF structure-tree path of markdown rendering. See the issue for before/after numbers and the full symptom list.

## The fix

**`adapters.rs::flatten_blocks`** — recurse into children AND emit parent's own `text`, instead of `continue`ing past it. Children are emitted first, parent text after — this matches the reading order observed for tagged list items (`LBody` child carries the lead-in, parent text carries the continuation). Inline emphasis spans inside a parent paragraph end up emitted in reverse visual order, but all text is preserved. A bbox-aware merge that reconstructs a single paragraph with inline annotations is a possible follow-up.

**`content_convert.rs::element_to_paragraph`** — gate the text-pattern `is_list_item` detection (`is_list_prefix_multi_token`) on `heading_level.is_none()`. An element explicitly tagged `Heading` by the author beats a text-pattern heuristic.

+28 / −7 behavior change across 2 files.

## Scope

Grepped every recursive block walker in the codebase for the same "recurse on children, skip parent" shape:

- `html.rs::walk_nodes` — pushes `ListItem` text before recursing; safe.
- `docx.rs` / `pptx.rs` / `email.rs` / `pst.rs` embedded-children handling — unrelated attachment logic, not text flattening.
- `bridge.rs::filter_blocks_recursive` — preserves parent structure when filtering sidebar blocks; doesn't drop text.

Only `adapters.rs::flatten_blocks` has this bug.

## Testing

Two synthetic unit tests — no binary fixture, no external-tool dependency:

- `adapters.rs::tests::test_from_structure_tree_preserves_parent_text_with_children` — constructs `LBody(text + child P)` and asserts both texts are emitted.
- `content_convert.rs::tests::test_heading_with_numeric_prefix_not_marked_list_item` — asserts a `Heading { level: 3 }` with text `"3. Conclusions"` keeps `is_list_item=false`.

Verified fail-before (both panic / assert-fail with the production fix temporarily reverted) and pass-after.

```
cargo test -p kreuzberg --lib --features pdf
# → 1569 passed (1567 existing + 2 new), 0 failed

cargo test -p kreuzberg --features pdf \
    --test pdf_output_quality \
    --test pdf_hierarchy_quality \
    --test pdf_table_ground_truth \
    --test markdown_lint_quality \
    --test issue_359_list_whitespace_test
# → 20 passed, 1 ignored (pre-existing), 0 failed
```

Release-mode `pdf_markdown_regression` (F1-score quality gate on 23 PDFs) is also running — result will be posted as a comment once it completes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -65,6 +65,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- **#774**: Tagged-PDF structure tree dropped paragraph body text when a block had both own text and children, and wrapped numbered section headings in an invalid `List → Heading` AST (panics comrak in debug, emits malformed markdown in release). `flatten_blocks` now emits parent text alongside children; text-pattern list detection in `element_to_paragraph` is gated on `heading_level.is_none()`.
 - **Semantic chunker fallback path now respects `max_characters`** — previously the non-embedding fallback hardcoded a 4000-char ceiling and silently ignored the caller's `max_characters`. A warning is also emitted when `chunker_type='semantic'` is used without an `EmbeddingConfig` so the fallback mode is discoverable. The `ChunkerType::Semantic` docstring has been corrected to describe both paths accurately.
 - **OCR backend dispatch**: `OcrConfig(backend=...)` with a non-default backend no longer silently falls back to paddleocr when the chosen backend errors — auto-fallback is limited to the default tesseract backend; users who want multi-backend fallback configure it via `OcrConfig.pipeline` (unchanged).
 - **EasyOCR on PDFs**: `EasyOCRBackend.supports_document_processing()` returns `False` so Rust's `PdfRenderer` handles page rendering, removing the implicit `pdf2image`/`pymupdf` requirement that was never declared in the `[easyocr]` extra.
--- a/crates/kreuzberg/src/pdf/structure/adapters.rs
+++ b/crates/kreuzberg/src/pdf/structure/adapters.rs
@@ -19,11 +19,27 @@ pub(super) fn from_structure_tree(blocks: &[ExtractedBlock]) -> PageContent {
 }
 
 /// Recursively flatten `ExtractedBlock` hierarchy into `ContentElement`s.
+///
+/// A block is the author's logical unit (paragraph, list-item, table-cell).
+/// Either the block's own `text` OR its nested children can be non-empty — and
+/// in tagged PDFs both occur frequently: a `Paragraph`/`LBody` whose child span
+/// carries the bold lead-in while the continuation text lives on the parent.
+///
+/// For each block we emit:
+///   1. Its children's contents (recursive flatten), then
+///   2. Its own `text` as a separate element, if non-empty.
+///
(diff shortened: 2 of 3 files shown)


dergachoff:
This bug with disappearing and/or misplaced text blocks haunted me across many versions with many different pdfs, so I eventually settled with text output instead of markdown. Finally decided to tackle it. Hope this helps the community.

kh3rld (approved):
looks good!

Goldziher (approved):

PR #798, opened 2026-04-23
fix: image links missing from markdown when using ImageExtractionConfig from Python

### LINKED ISSUE #762: bug: Unreliable image link extraction.
I have a simple scenario:
- I create a document ( tested with google docs or LibreOffice) 
- I create some text and insert an image
- I export/download the document as pdf

Expected behaviour: 
When I extract images and content with kreuzberg from pdf (to markdown), the markdown file should contain a link to the image (E.g. ` 

Actual behaviour:
No link extracted

The scenario is easy enough to reproduce. E.g.:
```
ExtractionConfig(
  output_format="markdown",
  images=ImageExtractionConfig(
            extract_images=true
  )
)
``` 

I tried various text editors, images, text variations & kreuzberg configs, but no luck. 
So I think its a bug. ( My kreuzberg version: v4.9.2 in python)

kh3rld:
fixes #762 ,.. reported that `result.content` was missing image markdown links (e.g. ``) when using `ImageExtractionConfig(extract_images=True)` from Python,even though `result.images` correctly contained the extracted images.                   
                                  
  **Root cause**: `inject_placeholders` was typed as `bool` in `ImageExtractionConfig`.    
  Alef-generated bindings sent an explicit `false` as the default, which instructed the structure pipeline to skip injecting image link placeholders. Rust's `unwrap_or(true)`  was never reached because the field was always present.
                                                                                           
  **Fix**: Changed `inject_placeholders: bool` → `Option<bool>`. Language bindings now send  `None`/`null` (omit the field) by default, letting Rust apply `unwrap_or(true)`. Updated all 7 extractor callers from `.map()` to `.and_then()`. Regenerated all 11 language  binding packages via `alef generate`.
                                                                                           
  Also removed an unused `CharData` import in `tests/helpers/mod.rs` that was causing all integration tests to fail to compile.                                                  
                                                                                           
  ## test plan                    
                                                                                           
  - [ ] `cargo check --package kreuzberg --tests` passes with no errors
  - [ ] Existing `inject_placeholders` tests in `html.rs`, `docbook.rs`, `rst.rs`,       
  `latex/mod.rs`, `pptx.rs` continue to pass                                               
  - [ ] Python: `ExtractionConfig(output_format="markdown",                              
  images=ImageExtractionConfig(extract_images=True))` on a PDF now produces image links in 
  `result.content`                                                                       
  - [ ] Python: `ImageExtractionConfig(inject_placeholders=False)` still suppresses image  
  links (opt-out still works)


Goldziher (changes requested):
## Review

Real bug — `unwrap_or_default()` produces `false` for boolean fields that should default to `true`. However, the fix is incomplete.

### Critical

**Same bug exists in ALL language bindings, not just Python.** The identical `unwrap_or_default()` producing `false` instead of `true` for `inject_placeholders` appears in: kreuzberg-node, kreuzberg-wasm, kreuzberg-php, kreuzberg-rb (Ruby), kreuzberg-nif (Elixir), and R bindings. Since these files are alef-generated, the fix must go into the alef template/generator source and all bindings must be regenerated via `alef generate`.

**Two additional boolean fields have the same broken default.** In the same `ImageExtractionConfig` constructor, `extract_images` and `auto_adjust_dpi` also use `unwrap_or_default()`, resolving to `false` instead of `true`. The Rust core uses `#[serde(default = "default_true")]` for all three. Fix all three together.

**Numeric fields also have wrong defaults.** `target_dpi`, `min_dpi`, `max_dpi`, and `max_image_dimension` use `unwrap_or_default()` producing 0, while the Rust core specifies 300, 72, 600, and 4096 respectively. These should be `unwrap_or(300)` etc. Same class of bug, same constructor — fix together.

### Warning

**R binding has a double representation.** `packages/r/src/rust/src/lib.rs` has `inject_placeholders: bool = false` as a macro default argument, separate from the `unwrap_or_default()`. Both need fixing.

### Missing

- Regression test that constructs `ImageExtractionConfig` with no arguments and asserts `inject_placeholders=true`, `extract_images=true`, `auto_adjust_dpi=true`.
- Freshness check: `alef generate && git diff --exit-code packages/ crates/kreuzberg-node/ crates/kreuzberg-wasm/ crates/kreuzberg-ffi/`

### Correct pattern

The fix for all three boolean fields should be:
```rust
extract_images: extract_images.unwrap_or(true),
inject_placeholders: inject_placeholders.unwrap_or(true),
auto_adjust_dpi: auto_adjust_dpi.unwrap_or(true),
```
This pattern already exists in the same codebase (e.g., `use_cache: use_cache.unwrap_or(true)`).

Goldziher:
Superseded — the underlying bug (#762) is fixed on main via a different mechanism: `#[serde(default = "default_true")]` on `inject_placeholders: bool` in `ImageExtractionConfig` (crates/kreuzberg/src/core/config/extraction/types.rs:29). Regression test at line 156-162 covers the absent-field case. The pipeline reordering in #834 also contributed. Closing since the goal is achieved.

PR #799, opened 2026-04-24
fix(pdf): extract images nested inside Form XObjects

### LINKED ISSUE #788: bug: Better handling of technical drawings which contains thousands of small raster tiles

Some technical drawing of the appliance and appears to be constructed from 86 small
raster tiles; visually a single figure, structurally many embedded images. None of them are returned.

We suspect the `image_extraction.source=pdfium bitmap fallback` path
Kreuzberg logs during extraction is aggressively deduplicating or
filtering images below some size/area threshold, because the six that
do come through are the larger single-image pages. A more faithful
extraction of all `XObject` images (or at least a clear way to opt into
that) would help consumers reconstruct multi-tile figures.

There are many technical documents that have these kind of drawings

kh3rld:
Summary
                                                                                            
  - PDF image extraction now recursively descends into Subtype=Form XObjects, so images composing tiled technical drawings (and similar structures) are no longer silently dropped.                                                                                  
  - Fix a pre-existing use use ... ;; duplicate-import syntax error in the alef-generated elixir NIF binding that blocked cargo fmt from parsing the file.                          
  - Apply cargo fmt --all uniformly across the workspace and fix two doc typos (entrys → entries).                                                                                 
                                                            
  root cause (fixes #788)                                                                   
                                                            
  lopdf::get_page_images only scans the page-level Resources → XObject dictionary. PDF technical drawings commonly store their constituent raster tiles inside a Subtype=Form XObject whose own Resources dict contains the images. This one level of indirection was never traversed, so every image inside a Form XObject was invisible to kreuzberg,.. the reporter's 86 raster-tile images returned 0 extracted images while 6 larger directly-embedded images extracted fine.

  fix

  Replaced the flat get_page_images call with a new recursive collect_images_from_resources method (MAX_FORM_DEPTH = 8) that:
  - Collects Subtype=Image entries as before.                                               
  - Recurses into Subtype=Form entries by fetching the Form stream and descending into its Resources dict.                                                                         
                                                                                            
  The same recursion is applied to:                         
  - extract_images_from_page (the per-page lopdf path).                                     
  - reextract_raw_images_via_pdfium - uses pdfium's PdfPageObject::XObjectForm::iter() to walk nested objects.                                                                      
  - objects_to_page_data in bridge.rs - keeps image-position counting in sync via a new count_image_objects recursive helper.                     
                                                                                            
  test plan                                                 
                                                                                            
  - New unit regression test: builds a minimal lopdf PDF with one image nested one level inside a Form XObject and asserts extract_images_from_pdf returns exactly that image
  Test plan

  - New unit regression test: builds a minimal lopdf PDF with one image nested one level inside a Form XObject and asserts extract_images_from_pdf returns exactly that image (pdf::images::tests::test_extract_images_nested_in_form_xobject,.. passes).
  - Full pdf::images::tests suite: 14/14 pass.
  - cargo fmt --all -- --check: passes.
  - alef verify: passes.
  - prek run --all-files: passes.


Goldziher (changes requested):
## Review

Correct diagnosis — Form XObjects contain nested images that the current pipeline misses. The fix direction (recursing via pdfium's `as_x_object_form_object()`) is appropriate, but there are correctness and security gaps.

### Critical

**Index alignment invariant must be maintained.** `objects_to_page_data` (bridge.rs) and `populate_images_from_pdfium` (pipeline.rs) must walk objects in the exact same order and count. If XObject recursion is added to one but not the other, or the DFS order differs, every image after the first Form XObject is misidentified — silent data corruption, not an error. Please demonstrate both traversal sites are updated with identical traversal order.

**No depth limit on XObject recursion.** The PDF spec permits Form XObjects to reference other Form XObjects, including circular references in malformed documents. The codebase already handles this pattern — `bookmarks.rs` uses a depth counter capped at 50. Without a depth limit, a pathologically crafted PDF causes unbounded stack growth. Add a `MAX_XOBJECT_DEPTH` constant (suggest 10) with early return.

**No visited-set for cycle detection.** Even with depth limiting, a DAG with the same Form XObject referenced from multiple parents will cause the same image to be counted and extracted multiple times, corrupting the index sequence. A `HashSet<ObjectId>` per page is required. This is especially likely in the "thousands of small raster tiles" case from #788, where the same tile XObject is probably referenced repeatedly.

### Warning

**`max_images_per_page` must apply to the total including XObject-nested images.** If the cap is applied only at the top level, the per-page count that `populate_images_from_pdfium` must process is unbounded, reintroducing the hang.

**The lopdf path (`pdf/images.rs`) may also miss Form XObject images.** `document.get_page_images()` may or may not traverse XObjects internally — please verify and document.

### Missing

- Regression test constructing a minimal PDF with images inside a Form XObject (lopdf can create one inline — no large binary fixture needed).
- Test for depth > 1 nesting.
- Test verifying a Form XObject referenced from two places is not double-counted.

Goldziher:
Squash-merged into main as ce4b42507; credited in CHANGELOG [Unreleased].

PR #800, opened 2026-04-24
fix(pdf): enforce max_images_per_page cap in image extraction path (#789)

### LINKED ISSUE #766: bug: extract_bytes hangs on PDFs with many image objects per page
## Summary

`kreuzberg::extract_bytes` hangs indefinitely on PDFs where a single page
contains thousands of image objects (e.g. a detailed technical diagram stored
as 2400+ individual image fragments). The configured `extraction_timeout_secs`
does not interrupt the operation.

## Reproduction

Given a PDF with:
- 88 pages, 27 MB total
- On one page it contains a single visual diagram that the PDF producer has stored
  as **2487 separate image objects**

Calling `extract_bytes` with `extract_images: true`:

```rust
let config = ExtractionConfig {
    images: Some(ImageExtractionConfig {
        extract_images: true,
        inject_placeholders: true,
        ..Default::default()
    }),
    ..Default::default()
};

// This never returns:
let result = kreuzberg::extract_bytes(&content, "application/pdf", &config).await?;
```

The call hangs for over an hour (tested multiple times) and never returns,
despite `extraction_timeout_secs` being set to 120 seconds.

## Root cause analysis

1. **No per-page image cap**: pdfium iterates over all image objects on a page
   and extracts each one. For pages with thousands of fragments (exploded
   technical drawings, vector diagrams saved as raster tiles), this takes
   extremely long or hangs.

2. **Timeout does not work**: `extraction_timeout_secs` appears to have no
   effect when pdfium is in a long-running image extraction loop. This is
   likely because pdfium operates via native C/FFI calls that cannot be
   interrupted by a tokio timeout.

## Impact

- The calling process (in our case a document harvester) is completely blocked
  and never progresses to the remaining documents.
- There is no way to work around this via configuration — `ImageExtractionConfig`
  has no `max_images` or `max_images_per_page` field.
- Setting `extract_images: false` avoids the hang but disables image extraction
  for all documents, which is not acceptable when downstream VLM captioning
  needs the images.

### 1. Add `max_images_per_page` to `ImageExtractionConfig`

```rust
pub struct ImageExtractionConfig {
    pub extract_images: bool,
    pub max_images_per_page: Option<u32>,  // NEW
    // ... existing fields
}
```

When the number of image objects on a page exceeds this limit, kreuzberg
should either:
- **Skip individual image extraction** for that page and render the entire
  page as a single image instead (preferred — the visual content is preserved)
- Or skip image extraction for that page entirely and log a warning

### 2. Fix `extraction_timeout_secs` for pdfium operations

The timeout should actually abort long-running pdfium operations. If
pdfium's C/FFI calls cannot be interrupted, consider running the extraction
in a `spawn_blocking` task with a `tokio::time::timeout` wrapper that
drops the blocking task handle on timeout.

## Environment

- kreuzberg: 4.8.2
- kreuzberg-pdfium-render: 4.6.3
- OS: Linux (Debian Bookworm)
- Rust: 1.94
- PDF: 88-page technical installation manual, 27 MB, produced with
  complex vector diagrams stored as individual image fragments

### LINKED ISSUE #789: fix: extract_bytes hangs on PDFs with many image objects per page
### Description

# kreuzberg issue: extract_bytes hangs on PDFs with many image objects per page

## Summary

`kreuzberg::extract_bytes` hangs indefinitely on PDFs where a single page
contains thousands of image objects (e.g. a detailed technical diagram stored
as 2487 individual image fragments). The configured `extraction_timeout_secs`
does not interrupt the operation.

 Fix #766 was supposed to fix this, but didn't.

## Root cause analysis

1. **No per-page image cap**: pdfium iterates over all image objects on a page
   and extracts each one. For pages with thousands of fragments (exploded
   technical drawings, vector diagrams saved as raster tiles), this takes
   extremely long or hangs.

2. **Timeout does not work**: `extraction_timeout_secs` appears to have no
   effect when pdfium is in a long-running image extraction loop. This is
   likely because pdfium operates via native C/FFI calls that cannot be
   interrupted by a tokio timeout.

## Impact

- The calling process (in our case a document harvester) is completely blocked
  and never progresses to the remaining documents.
- There is no way to work around this via configuration — `ImageExtractionConfig`
  has no `max_images` or `max_images_per_page` field.
- Setting `extract_images: false` avoids the hang but disables image extraction
  for all documents, which is not acceptable.

### 1. Add `max_images_per_page` to `ImageExtractionConfig`

```rust
pub struct ImageExtractionConfig {
    pub extract_images: bool,
    pub max_images_per_page: Option<u32>,  // NEW
    // ... existing fields
}
```

When the number of image objects on a page exceeds this limit, kreuzberg
should either:
- **Skip individual image extraction** for that page and render the entire
  page as a single image instead (preferred — the visual content is preserved)
- Or skip image extraction for that page entirely and log a warning

### 2. Fix `extraction_timeout_secs` for pdfium operations

The timeout should actually abort long-running pdfium operations. If
pdfium's C/FFI calls cannot be interrupted, consider running the extraction
in a `spawn_blocking` task with a `tokio::time::timeout` wrapper that
drops the blocking task handle on timeout.

## Environment

- kreuzberg: 4.8.2
- kreuzberg-pdfium-render: 4.6.3
- OS: Linux (Debian Bookworm)
- Rust: 1.94
- PDF: 88-page manual, 27 MB, produced with
  complex vector diagrams stored as individual image fragments

## Reproduction

Given a PDF with:
- 88 pages, 27 MB total
- Page 77 contains a single visual diagram that the PDF producer has stored
  as **2487 separate image objects**
- Page 56 similarly contains **213 image objects**

Calling `extract_bytes` with `extract_images: true`:

```rust
let config = ExtractionConfig {
    images: Some(ImageExtractionConfig {
        extract_images: true,
        inject_placeholders: true,
        ..Default::default()
    }),
    ..Default::default()
};

// This never returns:
let result = kreuzberg::extract_bytes(&content, "application/pdf", &config).await?;
```

The call hangs for over an hour (tested multiple times) and never returns,
despite `extraction_timeout_secs` being set to 120 seconds.

```text

```

kh3rld:
## Summary

- Wire `max_images_per_page` from `ImageExtractionConfig` to the actual lopdf image decode loop — it was only applied to the structure pipeline (bridge.rs for position counting), never to the path that decodes image data.
- Move both `extract_images_from_pdf` and `reextract_raw_images_via_pdfium` off the async executor into `spawn_blocking` so `extraction_timeout_secs` can actually interrupt them.

## Root cause

Fix #766 added `max_images_per_page: Option<u32>` to `ImageExtractionConfig` and wired it to `objects_to_page_data` in bridge.rs (image-position counting for the structure pipeline). It was **never passed** to `extract_images_from_pdf` — the function that actually decodes image bytes via lopdf. Both call sites in `extractors/pdf/mod.rs` called `extract_images_from_pdf(content)` without any limit, so a page with 2487 images triggered 2487 FlateDecode decompression + pdfium fallback iterations unconditionally.

The second issue: `extract_images_from_pdf` and `reextract_raw_images_via_pdfium` ran synchronously on the async executor thread. `tokio::time::timeout` can only fire at `.await` points, so any extraction running inside a blocking FFI call was completely uninterruptible by `extraction_timeout_secs`.

## Changes

**`crates/kreuzberg/src/pdf/images.rs`**
- `extract_images(max_images_per_page: Option<u32>)`: before entering the decode loop for each page, check image count against the cap. If exceeded, emit a `WARN` log and `continue` to next page.
- `extract_images_from_pdf(…, max_images_per_page)`, `extract_images_from_pdf_with_password(…, max_images_per_page)`: propagate cap through.

**`crates/kreuzberg/src/extractors/pdf/mod.rs`** (both pdfium and oxide paths)
- Read `config.images.as_ref().and_then(|i| i.max_images_per_page)` and pass it to `extract_images_from_pdf`.
- Wrap both `extract_images_from_pdf` and `reextract_raw_images_via_pdfium` in a single `tokio::task::spawn_blocking` closure, keeping them off the async thread.

## Test plan

- [ ] New unit test `test_max_images_per_page_cap_skips_dense_page`: two-page PDF, page 1 has 3 images, page 2 has 1. With `max_images_per_page = 2`, only the page-2 image is returned. Without cap, all 4 are returned. **Passes.**
- [ ] Full `pdf::images::tests` suite: 14/14 pass.
- [ ] `alef verify`: passes.
- [ ] Confirmed this is not a user environment issue — the cap field existed in config but was never read by the extraction path.

Closes #789


Goldziher (changes requested):
## Review

The core fix in `bridge.rs` is structurally sound — capping image count in the pdfium structure pipeline addresses the hang from #789.

### Critical

**The lopdf image extraction path has no cap.** `PdfImageExtractor::extract_images` in `pdf/images.rs` iterates all pages and all images per page with no bound. If a PDF with 1,924 inline images per page is processed through the lopdf path (triggered by `config.images.extract_images == true`), it hangs just as badly. `PdfImageExtractor::extract_images` must accept and honor `max_images_per_page`, breaking out of the per-page loop once the cap is hit, with a `tracing::warn!`.

### Warning

**No regression test for the cap logic.** There is no test that exercises `objects_to_page_data` with a non-zero cap and verifies both the returned `ImagePosition` count and the final `image_offset` value. Edge cases needed: cap=0, cap=exactly-equal-to-image-count, cap-exceeded.

**`unwrap_or(0)` when `.is_some()` is guaranteed.** At `bridge.rs:559-566`, the `if capped` block is only entered when `max_images_per_page.is_some_and(...)` returned true. Using `unwrap_or(0)` on the next line is needlessly defensive — restructure to pass the cap value directly.

**INVARIANT comment needs update.** The `debug_assert!` at `pipeline.rs:2746-2755` documents contiguous image indices but does not mention that the cap preserves contiguity because uncapped images form a dense prefix. Update the comment to reflect this.

### Notes

The +1524/-1282 diff is proportionate given that most is alef-generated binding code. Please confirm the freshness check passes: `alef generate && git diff --exit-code packages/ crates/kreuzberg-node/ crates/kreuzberg-wasm/ crates/kreuzberg-ffi/`

Goldziher (approved):
All four review points addressed:

- ✅ **lopdf path cap**: `PdfImageExtractor::extract_images` now takes `max_images_per_page: Option<u32>` and skips pages exceeding the cap with a `tracing::warn!`.
- ✅ **Edge case tests**: cap=0 and cap=exactly-page-image-count tests added.
- ✅ **`unwrap_or(0)` cleanup**: Restructured to `if let Some(cap)` pattern.
- ✅ **INVARIANT comment**: Updated to explain how skipped images still advance the offset, keeping the dense-prefix invariant.

LGTM.

Issue #352, opened 2026-02-03
Bug: `@kreuzberg/node` in `azure function` fails to load native bindings


paciops:
Hello,

I have this error when I try to use `@kreuzberg/node` inside an `azure function`.
The error is
```
Failed to load Kreuzberg native bindings. 
Report this error and attach the logs/stack trace for investigation. 
Underlying error: 
 Cannot find native binding. npm has a bug related to optional dependencies (https://github.com/npm/cli/issues/4828).
 Please try `npm i` again after removing both package-lock.json and node_modules directory.: 
 Cannot find native binding. npm has a bug related to optional dependencies (https://github.com/npm/cli/issues/4828). 
 Please try `npm i` again after removing both package-lock.json and node_modules directory.: /lib/x86_64-linux-gnu/libstdc++.so.6: version `GLIBCXX_3.4.31' not found (required by /home/site/wwwroot/node_modules/@kreuzberg/node-linux-x64-gnu/kreuzberg-node.linux-x64-gnu.node): 
 Cannot find module './kreuzberg-node.linux-x64-gnu.node' Require stack: - /home/site/wwwroot/node_modules/@kreuzberg/node/index.js - /home/site/wwwroot/node_modules/@kreuzberg/node/dist/index.mjs
```
Any idea on how to solve this? Thanks

Goldziher:
Thanks for reporting .

Can you can give me some context and maybe code to show me how you use the Azure function?

Also, do you know what is the arch / platform you are targeting? I.e. is this running on Linux? What arch?

paciops:
I am using the image [mcr.microsoft.com/azure-functions/node:4-node22](https://hub.docker.com/r/microsoft/azure-functions-node).
I am trying to see what I can share since the code is not open source, I'll try to send minimal reproduction to debug this error

Goldziher:
We investigated this issue and identified the root cause.

## Root Cause

The `@kreuzberg/node-linux-x64-gnu` binaries require `GLIBCXX_3.4.31`, but the `mcr.microsoft.com/azure-functions/node:4-node22` image only provides up to `GLIBCXX_3.4.30`.

| Image | Base OS | glibc | Max GLIBCXX |
|-------|---------|-------|-------------|
| `4-node22` | Debian 12 | 2.36 | 3.4.30 |
| `4-node24` | Ubuntu 24.04 | 2.39 | 3.4.33 |

## Solution

Upgrade to the `4-node24` Azure Functions image:

[...]

PR #803, opened 2026-04-24
fix(node): lower glibc floor of @kreuzberg/node prebuild to RHEL 8

hellno:
**Problem**

`@kreuzberg/node-linux-x64-gnu@4.8.5` requires `GLIBC_2.38` / `GLIBCXX_3.4.31`, which fails on RHEL 8 / AlmaLinux 8 / Rocky 8 / RHEL 9 (glibc 2.28 / 2.34). The Node prebuild runs on `ubuntu-latest` without an older sysroot, so it inherits the runner's glibc. Same class of bug as the Python side fixed in #596; reported for the Node binding in #352.

**Solution**

Switch the Linux gnu Node prebuilds to `cargo-zigbuild` via napi-rs's `--cross-compile` flag. Zig 0.14's linker caps the glibc floor at link time (targeting 2.28) and statically bundles libstdc++, so no `GLIBCXX_*` symbols leak into the `.node`.

`@napi-rs/cross-toolchain` was considered first but rejected: its GCC 4.8.5 predates `<stdatomic.h>` and can't compile `aws-lc-sys` (pulled transitively via rustls/reqwest). See [napi-rs/cross-toolchain#4](https://github.com/napi-rs/napi-rs/issues/cross-toolchain#4).

Two tesseract build-script tweaks are needed under zigbuild and gated on detecting zigcc via env (non-zigbuild targets — musl, macOS, Windows, native dev — unaffected):

- `HAVE_AVX512F=OFF`: zig/clang splits `avx512f` from `evex512`, breaking tesseract's AVX512 intrinsics.
- Skip linking `stdc++fs`: zig's bundled libstdc++ has `std::filesystem` inline; no separate `libstdc++fs.so` exists.

**Verification**

Built locally in a clean Ubuntu 24.04 container against the full feature set (running `pnpm exec napi build --cross-compile --target x86_64-unknown-linux-gnu`). Resulting artifact:

```
$ objdump -T kreuzberg.linux-x64-gnu.node | grep -oP 'GLIBC_[0-9.]+' | sort -uV | tail -1
GLIBC_2.28
$ objdump -T kreuzberg.linux-x64-gnu.node | grep GLIBCXX_ || echo "no GLIBCXX symbols"
no GLIBCXX symbols
```

Floor is `2.28`, no `GLIBCXX_*` dependency. The full reproduction is short — happy to share on request.

Refs #352, #588, #596.


Goldziher (changes requested):
## Review

The direction is correct — napi-cross for a glibc 2.28 floor is the right approach. However, the Python Linux build already hit this exact problem and the Node build needs equivalent guards.

### Critical

**No C23 glibc symbol verification.** The Python build sets `CFLAGS="-std=gnu11"` and `CXXFLAGS="-std=gnu++17"` inside the build container to prevent GCC 14+ from emitting C23-versioned glibc symbols (`__isoc23_strtoll` etc.) that require glibc >= 2.38 — defeating the purpose of the 2.28 container. See `publish.yaml` lines 614-628. Without equivalent CFLAGS guards or evidence that the napi-cross container pins an older GCC, the resulting `.node` binary may still require glibc 2.38+.

**No post-build symbol verification.** The Python pipeline has a verification step (publish.yaml lines 697-735) that scans for C23 symbols and fails if found. The Node build needs an equivalent `objdump -T` or `nm` check on the `.node` file, failing if any `GLIBC_2.3[2-9]+` symbol is present.

### Warning

**ARM path unclear.** `aarch64-unknown-linux-gnu` currently builds natively on `ubuntu-24.04-arm`. If `--use-napi-cross` applies to ARM as well, it switches to cross-compilation inside a container — this changes the build environment substantially and may break if build scripts (ORT download, pdfium staging) assume native ARM. Please clarify whether napi-cross is applied to both arch targets or only x86_64.

**Generated files.** Files in `packages/node/` and `crates/kreuzberg-node/package.json` are generated by Alef — manual edits will be overwritten on the next `alef generate` run. If changes to those files are needed, they should go through the alef workflow.

### Missing

- A CI job that installs the built `.node` binary inside a `ubi8` or equivalent glibc-2.28 container and runs the smoke test suite.

hellno:
**1. C23 symbol verification.** Doesn't apply under zigbuild: zig invokes its bundled clang, which inherits clang's default of `-std=gnu17` (the GCC 14 → 15 default switch to `gnu23` doesn't reach this toolchain). Local audit on the built `.node` confirms zero `__isoc23_*` symbols. To make that explicit and catch any future toolchain change, the new audit script (point 2) also hard-rejects any `__isoc23_*` reference, mirroring the Python pipeline's check at `publish.yaml` 697-735.

**2. Post-build symbol verification.** Added `scripts/publish/node/verify-glibc-floor.sh`, run as a new step after `Build native module` in the linux-gnu matrix runs. Uses `objdump -T` and rejects any `GLIBC_*` > 2.28, any `GLIBCXX_*`, or any `__isoc23_*`. Locally exercised on three artifacts (real `.node` → pass, fake .so referencing `GLIBC_2.32` → fail, glibc-free .so → fail).

why this audit isn't redundant with zigbuild? → `cargo-zigbuild` 0.22.2 doesn't pin a glibc minor unless the Rust target carries a `.<major>.<minor>` suffix, and that suffix flows into napi-rs's `platformArchABI` and breaks loader resolution. So the floor today is effectively "highest glibc symbol any dep references". fine for right, but a future dep that calls a newer libc symbol could silently raise the effective floor. the script is the only mechanism that catches that.

**3. ARM path.** No napi-cross is involved on either arch. The matrix at `publish.yaml` 825-836 builds `aarch64-unknown-linux-gnu` natively on `ubuntu-24.04-arm` with `use_zigbuild: true` — same toolchain, same env, no container cross-compile. ORT and pdfium staging paths are unchanged.

**4. Generated files.** This PR only touches `publish.yaml`, `CHANGELOG.md`, `crates/kreuzberg-tesseract/build.rs`, and `scripts/publish/node/build-native-module.sh`. no alef files collision.

**5. ubi8 smoke job.** Added `node-smoke-glibc-2-28` job (matrix x64 + arm64 via QEMU). Downloads each `node-bindings-*` artifact, mounts the `.node` into `registry.access.redhat.com/ubi8/nodejs-22:latest`, and runs `scripts/publish/node/smoke-glibc-2-28.mjs` to dlopen and exercise the napi surface. `publish-node` now `needs:` it and gates publish on success. Locally re-ran the smoke against the built `.node` inside actual ubi8: 15/15 OK.

CHANGELOG entry updated to mention both guards.

PR #822, opened 2026-04-25
fix(helm): resolve init container root conflict and KREUZBERG_PORT collision

### LINKED ISSUE #794: fix: init container root conflict and KREUZBERG_PORT collision
### Description

# Helm chart issues: init container root conflict and KREUZBERG_PORT collision

## Summary

Two issues prevent the kreuzberg Helm chart from working out of the box in a standard Kubernetes environment when `cache.enabled: true`:

1. The cache init container runs as root while the pod security context enforces `runAsNonRoot: true`
2. Kubernetes service discovery injects `KREUZBERG_PORT` into the pod, which conflicts with the binary's own config variable of the same name

---

### What happens

When `cache.enabled: true`, the chart creates an init container to chown the cache directory:

```yaml
initContainers:
  - name: init-cache
    securityContext:
      runAsUser: 0   # root
```

At the same time, the chart's default `podSecurityContext` sets:

```yaml
podSecurityContext:
  runAsNonRoot: true
```

The kubelet enforces this combination as a contradiction and refuses to start the pod:

```
Error: container's runAsUser breaks non-root policy
       (container: init-cache)
```

### Why it's a problem

- `runAsNonRoot: true` is the chart's own default — users shouldn't have to fight it to use the chart's own cache feature
- The workaround is to explicitly set `podSecurityContext.runAsNonRoot: false`, which disables the non-root guard for the entire pod including the main container

### Why the init container may not be needed

The pod security context already sets `fsGroup: 1000`. For block-backed volumes (e.g. EBS/gp2), Kubernetes automatically chowns the volume root to the `fsGroup` on mount, making the `chown -R 1000:1000` in the init container redundant.

### Suggested fixes

**Option A** — Remove `runAsNonRoot: true` from the default `podSecurityContext`, since `runAsUser: 1000` on the main container already enforces non-root execution in practice.

**Option B** — Add a chart value to disable the init container for environments where `fsGroup` handles ownership (most cloud block storage):

```yaml
cache:
  enabled: true
  initChown: true   # set to false to skip the init container
```

**Option C** — Make the init container run as non-root and rely on `fsGroup` alone for directory ownership.

---

### What happens

Kubernetes injects environment variables for every Service in the same namespace using the pattern `{SVCNAME}_PORT=tcp://<clusterIP>:<port>`. When the Helm release is named `kreuzberg`, the resulting Service is also named `kreuzberg`, and Kubernetes injects:

```
KREUZBERG_PORT=tcp://172.20.x.x:8000
```

The kreuzberg binary reads `KREUZBERG_PORT` as a config key expecting a `u16` port number. It finds the Kubernetes-injected value and panics at startup:

```
Validation error: KREUZBERG_PORT must be a valid u16 number,
got 'tcp://172.20.x.x:8000': invalid digit found in string
```

### Why it's a problem

- The default release name recommended in the docs (`kreuzberg`) triggers this every time
- The binary is unusable without a workaround
- The error message doesn't hint at Kubernetes service discovery as the cause, making it hard to diagnose

### Suggested fix

Set `enableServiceLinks: false` in the pod spec to prevent Kubernetes from injecting service discovery environment variables:

```yaml
# In deployment.yaml
spec:
  template:
    spec:
      enableServiceLinks: false
```

Expose it as a chart value with `true` as the default for backwards compatibility, or just default it to `false` since CoreDNS makes service env vars unnecessary in modern clusters:

```yaml
# values.yaml
enableServiceLinks: false
```

---

## Environment

- Kubernetes 1.29+ (standard PSA enforcement)
- Chart version: 4.8.4
- `cache.enabled: true`
- Release name: `kreuzberg` (default)

### Steps to reproduce

Deploy the chart with `cache.enabled: true`

```text

```

kh3rld:
Fixes #794

## What broke and why

Two chart bugs that reproduce with default values:

**1. Init container / `runAsNonRoot` conflict**

With `cache.enabled: true` (the default), the chart creates an init container with `runAsUser: 0`, while `podSecurityContext.runAsNonRoot: true` is also the default. Kubernetes enforces this at the container level and refuses to start the pod:

```
Error: container's runAsUser breaks non-root policy (container: init-cache)
```

**2. `KREUZBERG_PORT` collides with Kubernetes service discovery**

When the release is named `kreuzberg`, Kubernetes injects `KREUZBERG_PORT=tcp://<clusterIP>:8000` via service discovery (`enableServiceLinks` defaults to `true`). The binary parses `KREUZBERG_PORT` as a `u16` and panics immediately with:

```
Validation error: KREUZBERG_PORT must be a valid u16 number,
got 'tcp://172.20.x.x:8000': invalid digit found in string
```

### `deployment.yaml`

- Added `enableServiceLinks: {{ .Values.enableServiceLinks }}` to the pod spec.
- Added `runAsNonRoot: false` to the init container's `securityContext` so it can override the pod-level restriction for its own process.
- Gated the init container behind `cache.initChown` (and `cache.enabled`) so users on fsGroup-aware storage can opt out.

### `values.yaml`

- `enableServiceLinks: false` — prevents Kubernetes from injecting service discovery env vars. CoreDNS makes these unnecessary in modern clusters; users who rely on them can set this to `true`.
- `cache.initChown: true` — controls whether the init container is created. Defaults to `true` for backwards compatibility and for storage classes (NFS, some distributed stores) that don't honour `fsGroup`. Users on EBS, GKE PD, or Azure Disk can set this to `false`.

## Test plan

- [ ] `helm lint charts/kreuzberg/` passes (verified locally)
- [ ] `helm template kreuzberg charts/kreuzberg/` renders `enableServiceLinks: false` and `runAsNonRoot: false` on the init container
- [ ] `helm template ... --set cache.initChown=false` produces no `initContainers` block
- [ ] `helm template ... --set enableServiceLinks=true` renders `enableServiceLinks: true`
- [ ] Deploy with default values + `cache.enabled: true` + release name `kreuzberg` on a 1.29+ cluster — pod starts cleanly

### DIFF
--- a/.github/workflows/ci-lint.yaml
+++ b/.github/workflows/ci-lint.yaml
@@ -11,6 +11,7 @@ on:
       - "fixtures/**"
       - "scripts/**"
       - "tools/**"
+      - "charts/**"
       - "Cargo.toml"
       - "Cargo.lock"
       - "alef.toml"
@@ -27,6 +28,7 @@ on:
       - "fixtures/**"
       - "scripts/**"
       - "tools/**"
+      - "charts/**"
       - "Cargo.toml"
       - "Cargo.lock"
       - "alef.toml"
@@ -52,7 +54,7 @@ jobs:
       CARGO_TERM_COLOR: always
       FORCE_JAVASCRIPT_ACTIONS_TO_NODE24: "true"
     steps:
-      - uses: actions/checkout@v6
+      - uses: actions/checkout@v4
 
       - name: Setup Rust
         uses: kreuzberg-dev/actions/setup-rust@v1
--- a/.github/workflows/publish-helm.yaml
+++ b/.github/workflows/publish-helm.yaml
@@ -34,7 +34,7 @@ jobs:
       packages: write
     steps:
       - name: Checkout
-        uses: actions/checkout@v6
+        uses: actions/checkout@v4
 
       - name: Resolve version
         id: meta
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,9 +9,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-### Added
+### Fixed
 
-- **#768**: In-process embedding backend plugin. Callers that already own an embedder (sentence-transformers, llama-cpp-python, a tuned ONNX session, etc.) can register it once via `kreuzberg::plugins::register_embedding_backend(Arc::new(MyEmbedder))` and route kreuzberg's chunking and standalone embed paths into it through the new `EmbeddingModelType::Plugin { name }` config variant. Reachable from REST, MCP (`embedding_plugin`), CLI (`--provider plugin --plugin NAME`), and the `KREUZBERG_EMBEDDING_PLUGIN_NAME` environment variable; registry pre-flight returns the available backends list when the name isn't found. A new `EmbeddingConfig.max_embed_duration_secs` field bounds the wait on a hung backend (default 60s; `None` disables, `Some(0)` is treated as disabled). Host-side `registerEmbeddingBackend` is wired through Python (PyO3), Node (NAPI-RS), PHP (ext-php-rs), WASM (wasm-bindgen), Ruby (Magnus), Elixir (Rustler), R (extendr), Go (cgo), C# (P/Invoke), and the C FFI (`kreuzberg_register_embedding_backend`); Java's Panama backend can round-trip the `Plugin` config but does not yet emit plugin bridges. Adds fork-safety guidance for Python users running native-backed embedders under prefork servers (`os.register_at_fork`).
+- **#794**: Fix Helm chart default install broken by two conflicts: (1) the cache init container ran as `root` while `podSecurityContext.runAsNonRoot: true` is the default, causing kubelet to reject the pod; (2) Kubernetes service discovery injects `KREUZBERG_PORT=tcp://...` when the release is named `kreuzberg`, which the binary parses as a `u16` and panics. Fixed by adding `runAsNonRoot: false` to the init container's `securityContext`, a new `cache.initChown` toggle (default `true`, set to `false` on fsGroup-aware storage to skip the init container entirely), and defaulting `enableServiceLinks: false` in the pod spec.
 
 ---
 
--- a/charts/kreuzberg/templates/deployment.yaml
+++ b/charts/kreuzberg/templates/deployment.yaml
@@ -25,15 +25,17 @@ spec:
         {{- toYaml . | nindent 8 }}
       {{- end }}
       serviceAccountName: {{ include "kreuzberg.serviceAccountName" . }}
+      enableServiceLinks: {{ .Values.enableServiceLinks }}
       securityContext:
         {{- toYaml .Values.podSecurityContext | nindent 8 }}
-      {{- if .Values.cache.enabled }}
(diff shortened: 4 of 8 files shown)


Goldziher (changes requested):
## Review

The init container `runAsUser: 0` fix and the port collision are real fixes, but the chart's defaults remain broken in ways that will silently fail in real clusters.

### Critical

**1. RWO PVC + `replicaCount: 2` defaults break multi-node deployments.** `values.yaml` ships:
- `replicaCount: 2`
- `cache.enabled: true`
- `cache.accessModes: [ReadWriteOnce]`
- `strategy` unset → defaults to `RollingUpdate`

On any cluster where the 2 replicas land on different nodes, the second pod fails with `Multi-Attach error`. Even on a single node, `RollingUpdate` + `RWO` means the new pod stays `Pending` because the old pod still holds the volume. This is the most common failure mode for new chart users and there's no warning in `values.yaml` or the docs.

Pick one: default `cache.enabled: false`, default `replicaCount: 1` when cache is enabled, default `strategy.type: Recreate`, or document `ReadWriteMany` as a hard requirement. The current docs at `docs/guides/kubernetes.md:341` even show `replicas: 3` + `RWO` together, which is doubly wrong.

**2. Init container missing capability drop.** `templates/deployment.yaml:35-38` sets `runAsUser: 0` but no `capabilities`. The container runs as root with the default capability set, failing `restricted` PSA admission. Minimum fix:
```yaml
capabilities:
  add: ["CHOWN"]
  drop: ["ALL"]
allowPrivilegeEscalation: false
```
Better: with `fsGroup: 1000` already set on the pod, most CSI drivers chown the volume automatically — the init container is redundant. The docs at `docs/guides/kubernetes.md:244` already describe `fsGroup` as the alternative. Drop the init container in favor of `fsGroup` for the common case.

**3. Test pod has no security context at all.** `templates/tests/test-connection.yaml` runs `busybox` as root with no SC. `helm test` will fail in restricted PSA namespaces. Add pod-level `runAsNonRoot: true`, `runAsUser: 65534`, and container-level `allowPrivilegeEscalation: false`, `readOnlyRootFilesystem: true`, `capabilities.drop: ["ALL"]`.

### Warning

**4. `tag: "latest"` + `IfNotPresent` defaults** — non-reproducible, and `IfNotPresent` won't pull updates. Drop the default tag and let `kreuzberg.image` fall back to `Chart.AppVersion`.

**5. HPA + PDB conflict in defaults.** `autoscaling.minReplicas: 2` + `podDisruptionBudget.minAvailable: 1` means scale-down to 1 is blocked by the PDB. Configuration trap if both are enabled together.

**6. `actions/checkout@v6` doesn't exist.** `ci-lint.yaml:55` and `publish-helm.yaml:37`. Latest stable is `v4`. Will fail or resolve unexpectedly.

**7. `charts/**` not in CI lint paths filter.** `.github/workflows/ci-lint.yaml` push/pull_request `paths` doesn't include `charts/**`, so chart-only PRs skip helm-lint and kubeconform entirely.

**8. PVC has no retention annotation.** `helm uninstall` deletes the PVC and all cached models. Add `helm.sh/resource-policy: keep` or document the behavior.

### Note

**9. No `topologySpreadConstraints` or anti-affinity defaults.** With `replicaCount: 2`, both pods may land on the same node — no actual redundancy.

**10. `Chart.yaml` version drift.** Committed version is `0.1.0` while `appVersion: 4.8.x`. The publish workflow `sed`-replaces it but `0.1.0` shows up in `helm lint` output and any local install. Sync these explicitly or use `appVersion` directly.

PR #824, opened 2026-04-25
fix(pdf): robust image extraction across xobject references

### LINKED ISSUE #802: fix:  images are unreliably extracted from pdf pages
### Description

Hey there 👋 

I have tested kreuzberg with image extraction on a pdf.
My problem is, that images from page 14 and other pages are combined into a single image and than reported from kreuzberg as being a single image from page 14.
This causes issues for my downstream pipeline.

### Steps to reproduce

Use a pdf with images.
Expected behaviour: 
Images belong to the page they were found
Actual behaviour:
Images are mixed with images from other pages and reported as a single image

If you need an example pdf, I used the one available from:
https://www.figma.com/reports/design-systems-power-the-pace/

The pdf is rather large, but let me know if I should provide it somewhere/somehow ( e.g google drive) 

```text

```

kh3rld:
Fixes #802. Resolves PDF Image Extraction failures caused by inherited `Resources` in the page tree and images nested within `Form` XObjects.

### Changes
- Updated `PdfImageExtractor` to correctly retrieve inherited page resources via `get_page_resources()` instead of `get_page_images()`.
- Added recursive descent into `Form` XObjects to locate embedded images properly.
- Removed redundant `#[cfg(feature = "pdf")]` feature flags scattered inside the `images.rs` module, since the module itself is entirely guarded by the same feature flag.
- Ensured robustness against cyclic `Form` XObject references using a `seen` HashSet.
- Added regression tests for both reported bugs using `nougat_024.pdf` and `nougat_040.pdf`.


Goldziher (changes requested):
## Review

The `class` → `class_name` rename and dead validator removal are reasonable, but the PR ships broken bindings and an unflagged JSON API break.

### Critical

**1. Broken Java bindings — runtime `NoSuchFieldError`.** `packages/java/.../DepthValidator.java` and `StringGrowthValidator.java` reference `NativeLib.KREUZBERG_DEPTH_VALIDATOR_FREE` / `KREUZBERG_STRING_GROWTH_VALIDATOR_FREE` constants that no longer exist (the Rust types are gone but the binding files weren't deleted). Class init will fail. Either delete these files or add the FFI free functions back.

**2. C# memory leaks — `Dispose()` is a no-op stub.** `packages/csharp/.../{Depth,StringGrowth,ZipBomb}Validator.cs` `Dispose()` bodies contain only `// Native free will be called by the runtime`. There's no `DllImport` for `kreuzberg_*_validator_free` in `NativeMethods.cs`. Every validator instance leaks the native `Box<T>`. For `ZipBombValidator` (used in archive extraction) this is a real production leak.

**3. Freshness check gap.** `alef generate && git diff --exit-code packages/` won't catch this — `alef generate` doesn't delete orphaned files. The removed types should be added to `alef.toml`'s `[exclude]` list so future regenerations don't recreate stubs, and the orphan files need to be deleted manually.

### Warning

**4. Breaking JSON API change.** `LayoutRegion.class_name` serializes as `"class_name"` instead of `"class"`. No `#[serde(alias = "class")]` for backward compat. Cargo.toml is `4.9.5` but `#[deprecated(since = "4.10.0", ...)]` on the shim — self-referential and confusing. Bump version, document in CHANGELOG, add the serde alias for a transition period.

**5. Vacuous security tests.** `crates/kreuzberg/src/extractors/security_tests.rs` is full of tautologies: `assert!(true)` (7×), `assert!(text.len() >= 0)` (always true for usize), `assert!(text.contains("inline") || true)`. `test_epub_entity_expansion_protection` line ~109 asserts `malicious.len() > 100` where `malicious = "&;"` (length 2 — will fail). Either delete or rewrite as actual behavioral tests.

**6. Stale AI rules doc.** `.ai-rulez/domains/document-extraction/rules/extraction-safety.md` line 5 still says "Always use SecurityLimits validators: ZipBombValidator, DepthValidator, StringGrowthValidator". Future code following this rule won't compile.

### Note

**7. `LayoutRegion.bounding_box` typed as `String` in Java binding** (`LayoutRegion.java` line 23) instead of the structured `BBox` used by `LayoutDetection.java`. Likely an alef codegen issue but worth flagging.

**8. Inconsistent deprecated shims.** `LayoutDetection::class()` returns `LayoutClass` (by copy), `LayoutRegion::class()` returns `&str`. Different return shapes for a "compat" wrapper.

Goldziher:
Squash-merged into main as 88b3a33ec; credited in CHANGELOG [Unreleased].

PR #836, opened 2026-04-28
fix(pdf): prevent image data leak when extraction is disabled in structured output

### LINKED ISSUE #796: fix: pdf extraction without images still includes extracted images when output_format is markdown
### Description

Extraction result includes extracted images (base64) from a PDF even though the configuration explicitly sets
```
"pdf_options": {
  "extract_images": false,
}
```

This seems to only occur, when the `output_format` is set to either `markdown` or `djot`. Setting `plain` or `structured` works as expected: no image is extracted (no base64 excluded).

Expectation: Independent of the `output_format` setting, the result should not include extracted images when explicitly disabled. Especially as the `output_format` doesn't define the result format but just what's included in the content per the [docs](https://docs.kreuzberg.dev/reference/configuration/?h=output#result-format-vs-output-format):

> output_format - Controls the **text format** within the content:

This finally leads to longer response times but more importantly larger responses. A productive sample was a 1.6MB large PDF with 3 scanned pages included. The result was over 100MB. Whereas the plain text extraction (so without the `images`) was just 10KB. 

### Steps to reproduce

1. Use this baseline configuration: 
```
{
  "include_document_structure": true,
  "pdf_options": {
    "extract_images": false,
    "extract_metadata": true,
    "hierarchy": {
      "enable": true,
      "k_clusters": 6,
      "include_bbox": true,
      "ocr_coverage_threshold": 0.2
    }
  },
  "postprocessor": {
    "enabled_processors": [
      "whitespace_normalization",
      "mojibake_fix",
      "quality_scoring"
    ]
  },
  "ocr": {
    "backend": "tesseract",
    "tesseract_config": {
      "psm": 11,
      "oem": 1,
      "min_confidence": 0.8,
      "enable_table_detection": true,
      "language": "deu",
      "output_format": "markdown"
    }
  },
  "images": {
    "extract_images": false,
    "target_dpi": 200,
    "max_image_dimension": 20000
  },
  "language_detection": {
    "enabled": true,
    "min_confidence": 0.8,
    "detect_multiple": true
  },
  "token_reduction": {
    "mode": "off"
  },
  "enable_quality_processing": true,
  "force_ocr": false,
  "result_format": "unified"
}
```

2. Extract from the following test file (see attached **TestOCR.pdf**) - it uses `"output_format": "plain"` as that is the default
3. The output is as expected and doesn't include any `images` (see **response_plain.json**)
4. Explicitly add `"output_format": "markdown"` as root property.
5. Extract the same test file again
6. The output contains `"images": [{ "data": "..."}]` --> BUG (see **response_markdown.json**)

* verified on Kreuzberg v[v4.8.5](https://github.com/kreuzberg-dev/kreuzberg/releases/tag/v4.8.5) only so far 
  * scanned for fixed defects in that area but couldn't find a related one

[TestOCR.pdf](https://github.com/user-attachments/files/27015459/TestOCR.pdf)
[response_plain.json](https://github.com/user-attachments/files/27015605/response_plain.json)
[response_markdown.json](https://github.com/user-attachments/files/27015610/response_markdown.json)

kh3rld:
fixes #796 a critical bug where base64 image data was included in the `ExtractionResult` despite `extract_images=false` being set in the configuration. 

The issue occurred only when `output_format` was set to `Markdown`, `Djot`, or `Html`. In these cases, the PDF extractor activates the "structured document" pipeline, which was computing the `inject_placeholders` flag based solely on the user's placeholder preference, ignoring whether image extraction was actually enabled. This caused the pipeline to render and embed full image data even when the caller explicitly disabled it.

### changes
- Updated `extraction.rs` (both `pdfium` and `oxide` backends) to gate `inject_placeholders` on the `extract_images` config flags.
- Ensured that if image extraction is disabled, the structure pipeline skips `populate_images_from_pdfium` entirely, improving performance for non-image extractions.
- Added regression tests in `pdf_image_extraction_tests.rs` covering both `Markdown` and `Djot` formats.

#### automated tests
- Added 5 new regression tests in `crates/kreuzberg/tests/pdf_image_extraction_tests.rs` which verify:
    - `images.extract_images = false` suppresses image data in Markdown.
    - `pdf_options.extract_images = false` suppresses image data in Markdown.
    - Parity for Djot output format.
    - Sanity check that images *are* still present when explicitly enabled.

#### quality checks
- Ran `cargo fmt` and `cargo clippy`.
- Ran `prek run --all-files` to ensure project-wide consistency.

### DIFF
--- a/crates/kreuzberg-ffi/README.md
+++ b/crates/kreuzberg-ffi/README.md
@@ -73,6 +73,10 @@ Extract text, tables, images, and metadata from 91+ file formats and 248 program
 ### Package Installation
 
 
+
+
+
+
 ### System Requirements
 
 - See [Installation Guide](https://kreuzberg.dev/getting-started/installation/) for requirements
@@ -93,13 +97,20 @@ Extract text, metadata, and structure from any supported document format:
 Most use cases benefit from configuration to control extraction behavior:
 
 
+
 #### Table Extraction
 
 
 See [Table Extraction Guide](https://kreuzberg.dev/features/table-extraction/) for detailed examples.
 
 
+
 #### Processing Multiple Files
+
+
+
+
+
 
 
 ### Next Steps
@@ -108,6 +119,7 @@ See [Table Extraction Guide](https://kreuzberg.dev/features/table-extraction/) f
 - **[API Documentation](https://kreuzberg.dev/api/)** - Complete API reference
 - **[Examples & Guides](https://kreuzberg.dev/guides/)** - Full code examples and usage guides
 - **[Configuration Guide](https://kreuzberg.dev/guides/configuration/)** - Advanced configuration options
+
 
 
 ## Features
@@ -209,14 +221,18 @@ Powered by [tree-sitter-language-pack](https://github.com/kreuzberg-dev/tree-sit
 | **Web formats** | 50-200 MB/s | Streaming | HTML, XML, JSON |
 
 
+
 ## OCR Support
 
 Kreuzberg supports multiple OCR backends for extracting text from scanned documents and images:
 
 
+
 ### OCR Configuration Example
 
 <!-- snippet not found:  -->
+
+
 
 
 ## Async Support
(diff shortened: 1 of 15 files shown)


Goldziher:
Squash-merged into main as 2a3de2a75; credited in CHANGELOG [Unreleased].

PR #843, opened 2026-04-29
fix(python): unblock ExtractionConfig + add pdf_page_count and embedding registry helpers

tobocop2:
## Problem

Every public `extract_*` call on the Python wrapper crashes:

```python
from kreuzberg import ExtractionConfig, extract_bytes_sync
ec = ExtractionConfig(disable_ocr=True)
extract_bytes_sync(b"hi", "text/plain", ec)
# AttributeError: 'dict' object has no attribute 'use_cache'
```

`ExtractionConfig` was generated as a `TypedDict`, so calling it returns a plain `dict`, but the converters in `api.py` access fields with dot syntax.

A few public helpers were also missing: there was no `pdf_page_count`, and the embedding-backend registry only exposed `register_*` — no way to list, unregister, or clear without reaching into private modules.

## Solution

Flip the codegen DTO style for Python back to dataclasses so `ExtractionConfig` instances have real attributes again:

```python
ec = kreuzberg.ExtractionConfig(disable_ocr=True)
ec.disable_ocr  # True
```

Add `pdf_page_count(pdf_bytes, password=None) -> int` as a public Rust function and surface it across every binding:

```python
from kreuzberg import pdf_page_count
n = pdf_page_count(pdf_bytes)
```

Wire `unregister_embedding_backend`, `list_embedding_backends`, and `clear_embedding_backends` through the bindings:

```python
from kreuzberg import list_embedding_backends, unregister_embedding_backend
list_embedding_backends()
unregister_embedding_backend('my-backend')  # idempotent for unknown names
```

Tests cover all three across every binding via a shared fixture, plus Python regression tests for the original crash and the registry round-trip.

## Scope

This removes the first crash but doesn't yet make `extract_*` usable end to end. Once the `AttributeError` is gone, a chain of older codegen bugs surfaces and keeps that path broken (the next call shape raises a different error). Those live in the codegen tool rather than this repo and are tracked together at https://github.com/kreuzberg-dev/alef/issues/44.

Follow-up tracked here: #844.

The PR stays as a draft until those fixes land. The new helpers and registry surface stand on their own, so it can also be merged early if that's preferred.


Goldziher:
Closing — superseded by [`853f0732a`](https://github.com/kreuzberg-dev/kreuzberg/commit/853f0732a) which picks up the alef 0.11.24+ codegen fixes (kreuzberg-dev/alef#44) via regen. The hand-fixes to `packages/python/kreuzberg/api.py` in this PR were correct workarounds, but they would conflict with the regenerated wrappers; the codegen fix-set addresses all three failure modes (TypedDict dot-access, `#[serde(skip)]` propagation, kw-arg forwarding) at the source.

Thanks for digging into the root causes — your write-up in #844 made it possible to file a focused alef issue with reproducers.

PR #846, opened 2026-04-30
 feat: polling async API - POST /extract-async + GET /jobs/{job_id}

### LINKED ISSUE #619: feat: Streaming Responses Support in Docker API
### Problem

The current Docker API follows a synchronous request-response model, where the client must wait until the entire processing pipeline is completed before receiving any result.

For document processing workloads such as OCR, parsing, and chunking, this introduces several limitations:

- High latency for large files
- No visibility into processing progress
- No access to partial or intermediate results
- Difficult integration into interactive or long-running workflows

This makes it harder to build responsive systems on top of the API.

---

### Proposed Solution

Add support for streamed responses in the Docker API, allowing the server to return incremental results as they become available.

This could be implemented using:

- Server-Sent Events (SSE), recommended for simplicity
- HTTP chunked transfer encoding
- Optionally WebSockets, though not necessary for an initial version

Streaming could be enabled either through a dedicated endpoint or through content negotiation.

---

### Content Negotiation

One possible implementation would be to enable streaming via standard HTTP content negotiation instead of introducing a separate endpoint.

Example:

- `Accept: application/json` -> standard JSON response
- `Accept: text/event-stream` -> streamed response using SSE

This approach:

- avoids duplicating endpoints
- keeps the API surface smaller
- aligns with standard HTTP semantics

---

### Example API Design

Endpoint:

`POST /process`

Example request header:

`Accept: text/event-stream`

Example SSE response:

```text
event: started
data: { "jobId": "123" }

event: progress
data: { "stage": "ocr", "progress": 0.4 }

event: chunk_generated
data: { "chunkId": "c1", "type": "heading" }

event: chunk_generated
data: { "chunkId": "c2", "type": "table_like" }

event: partial_result
data: { "type": "entity", "value": "Example AG" }

event: completed
data: { "status": "done" }
```

---

### Benefits

1. Improved user experience  
   Clients can receive immediate feedback and display progress instead of waiting for the full response.

2. Access to partial results  
   Enables early validation and incremental processing on the client side.

3. Better support for large documents  
   Reduces perceived latency and allows handling long-running tasks more efficiently.

4. Easier integration with async systems  
   Streaming fits naturally with queue-based or distributed processing models.

---

### Trade-offs

- Increased implementation complexity, especially around connection handling and streaming lifecycle
- Requires a stable event schema and versioning strategy
- More complex client-side handling
- Harder to test compared to synchronous APIs

---

### Suggested Scope for Initial Version

- Support SSE only
- Define a minimal set of events:
  - `started`
  - `progress`
  - `partial_result`
  - `completed`
  - `error`
- Keep the response format simple and extensible

---

### Why This Matters

The current API behaves like a batch processing interface. Adding streaming capabilities would enable more responsive and interactive use cases, especially for large or complex documents.

This would significantly improve developer experience and flexibility when integrating the Docker API into real-world systems.

---

### Optional Future Extensions

- Ability to resume streams via job ID
- Integration with async job endpoints
- Event filtering, for example subscribing only to selected stages
- Backpressure handling

kh3rld:
Adds a non-blocking extraction path. `POST /extract-async` accepts the same multipart form as `POST /extract` and returns HTTP 202 with a job_id immediately; a background `tokio::spawn` task runs the extraction pipeline (bounded by `extraction_timeout_secs`, defaulting to 300 s) and writes the result into an in-memory `JobStore. GET /jobs/{job_id}`returns `JobStatus` - `state` (`Pending` → `Running` → `Completed` / `Failed`), ISO 8601 timestamps, `result`, and error. Jobs are evicted after 5 minutes and cleared on restart. Both endpoints are gated behind `--features api` and have no effect on the sync path.

  SSE event types (`PipelineEvent`, `StartedEvent`, etc.) are deliberately excluded - they'll land when the SSE endpoint is implemented, not before.
                                                                                           
  Tests: `JobStore` unit tests (state transitions, eviction); integration tests for the 202 response, 404 on unknown job, poll-to-`Completed`, and poll-to-`Failed` (unsupported MIME type).                                                                                   
                                                                      
  closes #619


crates/kreuzberg/src/api/error.rs
```diff
@@ -151,9 +151,23 @@ impl ApiError {
     /// Create a bad gateway error (502).
     ///
     /// Use when an upstream service (e.g., model download from HuggingFace) fails.
+    #[cfg(any(feature = "paddle-ocr", feature = "layout-detection", feature = "embeddings"))]
     pub(crate) fn bad_gateway(error: KreuzbergError) -> Self {
         Self::new(StatusCode::BAD_GATEWAY, error)
     }
+
+    /// Create a not found error (404).
+    pub(crate) fn not_found(message: impl Into<String>) -> Self {
```
Goldziher: why do we need this? 

Axum has `fallback` -- we dont need to bake our own.

crates/kreuzberg/src/api/handlers.rs
```diff
@@ -1139,6 +1146,165 @@ fn resolve_cache_base() -> std::path::PathBuf {
...
+#[cfg(feature = "api")]
+#[utoipa::path(
+    post,
+    path = "/extract-async",
+    tag = "extraction",
+    request_body(content_type = "multipart/form-data"),
+    responses(
+        (status = 202, description = "Job accepted", body = AsyncJobResponse),
+        (status = 400, description = "Bad request", body = crate::api::types::ErrorResponse),
+        (status = 413, description = "Payload too large", body = crate::api::types::ErrorResponse),
```
Goldziher, line 1183: what is the payload limit?

crates/kreuzberg/src/api/handlers.rs
```diff
@@ -1139,6 +1146,165 @@ fn resolve_cache_base() -> std::path::PathBuf {
...
+        let field_name = field.name().unwrap_or("").to_string();
+        match field_name.as_str() {
+            "files" => {
+                let file_name = field.file_name().map(|s| s.to_string());
+                let content_type = field.content_type().map(|s| s.to_string());
+                let data = field
+                    .bytes()
+                    .await
+                    .map_err(|e| ApiError::validation(crate::error::KreuzbergError::validation(e.to_string())))?;
+                let mut mime_type = content_type.unwrap_or_else(|| "application/octet-stream".to_string());
```
Goldziher: lets use constants for this stuff -- "application/octet-stream" and similar strings. DRY this out.

crates/kreuzberg/src/api/handlers.rs
```diff
@@ -1139,6 +1146,165 @@ fn resolve_cache_base() -> std::path::PathBuf {
...
+        }
+    }
+
+    if files.is_empty() {
+        return Err(ApiError::validation(crate::error::KreuzbergError::validation(
+            "No files provided",
+        )));
+    }
+
+    let job_id = uuid::Uuid::new_v4().to_string();
```
Goldziher: lets move this section to its own helper function - to create a job id. test it separately

crates/kreuzberg/src/api/jobs.rs
```diff
@@ -0,0 +1,182 @@
+//! In-memory job store for async extraction polling.
```
Goldziher, line 1: This implementation feels brittle and too weak. Why not use a proper library for this? What are our options? Please look around in the axum/tower ecosystem.

crates/kreuzberg/src/api/router.rs
```diff
@@ -130,8 +130,23 @@ pub(crate) fn create_router_with_limits_and_server_config(
     let state = ApiState {
         default_config: Arc::new(config),
         extraction_service: Arc::new(std::sync::Mutex::new(extraction_service)),
+        #[cfg(feature = "api")]
+        job_store: Arc::new(super::jobs::JobStore::new()),
     };
 
+    // Spawn periodic job eviction (every 60 seconds)
```
Goldziher: yes, all of these mechanisms. This is too bug prone I think.

Goldziher (changes requested):

crates/kreuzberg/src/api/handlers.rs
kh3rld: It uses the exact same limits as `POST /extract` - enforced at the router layer via `DefaultBodyLimit` + `RequestBodyLimitLayer` (100 MB default, configurable via `ApiSizeLimits`),.. I've added the missing docstring here to make that explicitly clear!

crates/kreuzberg/src/api/jobs.rs
kh3rld: I looked into the ecosystem options. Full task queues like `apalis` or `rusty-celery` are massive overkill since they require external databases (Redis/Postgres) and we just need an ephemeral in-memory cache for polling.

Instead, I swapped `DashMap` for the `moka` crate (`moka::sync::Cache`). It's heavily used in the ecosystem (DataFusion, Polars) and has built-in TTL expiration. This allowed us to drop both the manual lock retention logic and the brittle eviction loop entirely!

crates/kreuzberg/src/api/error.rs
```diff
@@ -151,6 +151,7 @@ impl ApiError {
     /// Create a bad gateway error (502).
     ///
     /// Use when an upstream service (e.g., model download from HuggingFace) fails.
+    #[cfg(any(feature = "paddle-ocr", feature = "layout-detection", feature = "embeddings"))]
     pub(crate) fn bad_gateway(error: KreuzbergError) -> Self {
```
Goldziher, line 154: why is this feature gated?

crates/kreuzberg/src/api/handlers.rs
```diff
@@ -7,15 +7,20 @@ use tower::Service;
...
-        EmbedResponse, ExtractResponse, HealthResponse, InfoResponse, ManifestEntryResponse, ManifestResponse,
-        VersionResponse, WarmRequest, WarmResponse,
+        ApiState, AsyncJobResponse, CacheClearResponse, CacheStatsResponse, ChunkRequest, ChunkResponse,
+        DetectResponse, EmbedRequest, EmbedResponse, ExtractResponse, HealthResponse, InfoResponse, JobStatusResponse,
+        ManifestEntryResponse, ManifestResponse, VersionResponse, WarmRequest, WarmResponse,
     },
 };
 
+/// Fallback MIME type used when the client does not provide a Content-Type.
+pub(crate) const MIME_OCTET_STREAM: &str = "application/octet-stream";
```
Goldziher: we already have constants for this, for mimetypes. reuse.

crates/kreuzberg/src/api/jobs.rs
Goldziher: ok

crates/kreuzberg/src/api/openweb.rs
```diff
@@ -65,7 +65,7 @@ pub(crate) async fn openweb_external_handler(
         .unwrap_or_else(|| "unknown".to_string());
 
     // Detect MIME type from filename if the header was generic
-    let mime_type = if mime_type == "application/octet-stream" {
+    let mime_type = if mime_type == super::handlers::MIME_OCTET_STREAM {
```
Goldziher: same point regarding shared constants

Goldziher (changes requested):

PR #847, opened 2026-04-30
fix(demo): prevent browser hang and OCR timeout on image uploads

### LINKED ISSUE #845: fix: prevent demo from hanging on small image uploads
# Demo hangs and drives CPU high when extracting image files under 1MB

## Summary

The live demo at <https://docs.kreuzberg.dev/demo.html> becomes unresponsive when a small image file is uploaded. I can reproduce this with a 725-byte PNG, which is far below the page's 1MB sandbox limit. Text files upload and extract normally in the same browser session.

## Environment

- Date tested: 2026-04-30
- URL: <https://docs.kreuzberg.dev/demo.html>
- Browser: Google Chrome 144.0.7559.109 on Ubuntu 24.04 / Linux 6.17.0
- Automation: agent-browser 0.26.0, headless Chrome
- Demo dependency observed in page source:
  - `@kreuzberg/wasm@latest/dist/index.js`
  - `@kreuzberg/wasm@latest/dist/pkg/kreuzberg_wasm_bg.wasm`
- jsDelivr response/version observed:
  - `dist/index.js`: `x-jsd-version: 4.8.6`
  - `dist/pkg/kreuzberg_wasm_bg.wasm`: `x-jsd-version: 4.9.4`
  - npm `@kreuzberg/wasm` latest: `4.9.4`

## Reproduction Steps

1. Open <https://docs.kreuzberg.dev/demo.html>.
2. Wait for the page to load.
3. Upload a small text file such as:

   ```txt
   hello from text file
   ```

4. Observe that the demo returns output normally.
5. Reload the page.
6. Upload a small PNG image under 1MB. My repro file was a generated `320x180` PNG, 725 bytes.

   Example file generation:

   ```bash
   mkdir -p /tmp/kg-demo-repro
   cd /tmp/kg-demo-repro
   node -e "const fs=require('fs'); const w=320,h=180; const header=Buffer.from('P6\\n'+w+' '+h+'\\n255\\n'); const data=Buffer.alloc(w*h*3); for(let y=0;y<h;y++){for(let x=0;x<w;x++){const i=(y*w+x)*3; data[i]=Math.round(255*x/(w-1)); data[i+1]=Math.round(255*y/(h-1)); data[i+2]=80;}} fs.writeFileSync('test.ppm', Buffer.concat([header,data]));"
   convert test.ppm test.png
   wc -c test.png
   ```

## Actual Behavior

- The page stops responding after the PNG upload.
- Browser automation calls that read page text time out with:

  ```txt
  CDP command timed out: Runtime.evaluate
  ```

- CPU stays high after the upload. In my run, the Chrome renderer and GPU processes together were close to one full CPU core:

  ```txt
  13:43:35 renderer 47.0% / gpu-process 44.4%
  13:43:36 renderer 47.6% / gpu-process 44.4%
  13:43:37 renderer 48.1% / gpu-process 44.4%
  13:43:39 renderer 48.7% / gpu-process 44.4%
  13:43:40 renderer 49.2% / gpu-process 44.5%
  13:43:41 renderer 49.8% / gpu-process 44.5%
  13:43:42 renderer 50.3% / gpu-process 44.5%
  13:43:43 renderer 50.9% / gpu-process 44.5%
  13:43:44 renderer 51.4% / gpu-process 44.5%
  13:43:45 renderer 51.9% / gpu-process 44.5%
  ```

The only way to recover the session was closing the browser.

## Expected Behavior

The demo should either:

- extract text/OCR output from supported image files,
- return a clear error if image OCR cannot run in the browser, or
- time out gracefully without locking the page.

It should not spin indefinitely or make the browser unresponsive for a tiny image well below the documented sandbox limit.

## Additional Observations

The page initializes OCR on load:

```js
await initWasm('https://cdn.jsdelivr.net/npm/@kreuzberg/wasm@latest/dist/pkg/kreuzberg_wasm_bg.wasm');
await enableOcr();
```

Uploads are processed with:

```js
const result = await extractFromFile(file, file.type || null);
```

Given that non-image files work and the hang happens on small PNG input, this looks likely related to the browser/WASM OCR image path rather than the 1MB sandbox limit.

### LINKED ISSUE #854: bug: @kreuzberg/wasm browser OCR pipeline silently degrades to synchronous fallback when loaded cross-origin (related #845)
## Summary

  When `@kreuzberg/wasm` is loaded from a different origin than the consuming page — including the live
  demo at `docs.kreuzberg.dev/demo.html` loading via `cdn.jsdelivr.net` — the inner OCR Web Worker that
  `worker-bridge.ts` is supposed to create is rejected by the browser's same-origin policy. The failure
  is caught silently by `createOcrWorker`'s outer `try/catch`, which sets `useFallback = true`. From
  that point forward all OCR calls take the synchronous `wasm.ocrRecognize(...)` fallback path on
  whatever thread invoked them, with no per-request timeout protection (the 20s `setTimeout` in
  `runOcrInWorker` cannot fire while a synchronous wasm call blocks the event loop).

  A second, latent issue compounds the first: the kreuzberg-tesseract WASM build with default Tesseract
  page-segmentation mode (PSM AUTO) appears to hang or run for tens of seconds on small images (60 KB
  screenshots, 725-byte PNGs). With the synchronous fallback active, only an external (main-thread)
  timeout can recover.

  ## Relationship to #845

  This is the underlying cause of #845's "demo hangs on small image uploads." PR #847's commit 127c62bfa
   ("fix: move WASM extraction to Web Worker to prevent browser freeze") successfully eliminates the
  *browser-hang* symptom — the demo page stays responsive — but extraction itself does not complete.
  With #847's 90s outer timeout, the demo now fails predictably with `Extraction timed out after 90s` on
   any image, including ones that contain text and would normally OCR.

  In other words: #845's symptom (frozen tab) is fixed by #847; #845's underlying functional bug
  (browser OCR doesn't work) is documented here.

  ## Evidence

  ### 1. Cross-origin Worker construction is rejected by the browser

  Pasted in the page's console:

  ```js
  try {
    const w = new Worker(
      'https://cdn.jsdelivr.net/npm/@kreuzberg/wasm@latest/dist/ocr-worker.js',
      { type: 'module' }
    );
    console.log('created:', w);
  } catch (e) {
    console.log('threw:', e.name, e.message);
  }
  ```

  Output:

  ```
  threw: SecurityError Failed to construct 'Worker': Script at
  'https://cdn.jsdelivr.net/npm/@kreuzberg/wasm@latest/dist/ocr-worker.js' cannot be accessed from
  origin 'http://127.0.0.1:8000'.
  ```

  This is the same `new Worker(url, { type: 'module' })` call performed by `createBrowserWorker` in
  `crates/kreuzberg-wasm/typescript/ocr/worker-bridge.ts`. Inside the bundled `dist/index.js`,
  `import.meta.url` resolves to the CDN URL of the bundle, so `new URL('./ocr-worker.js',
  import.meta.url)` is cross-origin to the page. Browsers reject cross-origin Worker URLs
  unconditionally — module workers only relax the same-origin policy for *imports inside* the worker
  (via CORS), not for the worker script itself.

Image

  ### 2. Inner OCR worker is never created — `ocr-worker.js` is never fetched

  After WASM cache warm-up, a single extraction triggers exactly seven worker-initiated fetches (Network
   tab, gear icon = service/web worker initiator):

  | Resource | Origin |
  |---|---|
  | `blob:http://127.0.0.1:8000/...` (outer extraction worker) | local |
  | `index.js` | cdn.jsdelivr.net |
  | `kreuzberg_wasm.js` | cdn.jsdelivr.net |
  | `kreuzberg_wasm_bg.wasm` | cdn.jsdelivr.net |
  | `pdfium.js` | cdn.jsdelivr.net |
  | `pdfium.esm.wasm` | cdn.jsdelivr.net |
  | `eng.traineddata` | raw.githubusercontent.com |

  `ocr-worker.js` is not in the list, even though it is published at `dist/ocr-worker.js` and serves a
  200. The only way the worker URL can remain unrequested is if the `new Worker(...)` constructor threw
  before the browser could initiate the fetch. This is consistent with the SecurityError above.

Image

  ### 3. Tessdata fetches quickly; 86+ seconds of pure CPU follow

  `eng.traineddata` Timing on a warm-cache run:

  - Queued at 3.02 s
  - Started at 3.02 s
  - Waiting for server response: 233.61 ms
  - Content download: 58.65 ms (4 MB payload)
  - Total: 369.83 ms

  The fetch completes well under 4 seconds from extraction start. The network tab then records zero
  further activity for ~86 seconds, until the outer 90 s timer fires. The worker is in a pure CPU state
  during this window — `wasm.ocrRecognize(...)` is the only thing it can be doing. A 60 KB image taking
  86+ seconds to OCR is well outside Tesseract's normal range and matches the documented PSM AUTO
  degenerate-input behavior.

Image

  ### 4. The inner 20 s OCR timeout never fires

  `runOcrInWorker` schedules a `setTimeout(..., 20_000)` that terminates the inner OCR worker and
  rejects with `OCR timed out after 20s — Tesseract hung on this image. Try a different image or a
  text-based file format.` This message is never observed. Only the main-thread message `Extraction
  timed out after 90s ...` appears. Two consistent explanations:

  - The fallback path is active, so `runOcrInWorker` returns via `Promise.resolve(fallbackFn(...))` and
  never sets the timer.
  - Even if the timer were set, a synchronous wasm call blocks the worker's event loop and prevents
  `setTimeout` callbacks from firing.

  Both follow from `useFallback = true` being set after the SecurityError above.

  ## Affected code paths

  - `crates/kreuzberg-wasm/typescript/ocr/worker-bridge.ts` — `createBrowserWorker` (`new Worker(new
  URL('./ocr-worker.js', import.meta.url), { type: 'module' })`)
  - `crates/kreuzberg-wasm/typescript/ocr/worker-bridge.ts` — `createOcrWorker` outer `try/catch`
  swallows the SecurityError and sets `useFallback = true`; no signal is propagated to callers
  - `crates/kreuzberg-wasm/typescript/ocr/worker-bridge.ts` — `runOcrInWorker` fallback branch invokes
  `fallbackFn(...)` synchronously
  - `kreuzberg-tesseract` Rust crate — default Tesseract page-segmentation mode (PSM AUTO) on
  small/sparse images

  ## Scope

  Affects browser consumers of `@kreuzberg/wasm` loading the package from any cross-origin source (CDN,
  separate static host, etc.), including the published `cdn.jsdelivr.net` distribution. Does not affect:

  - Node.js consumers (`createNodeWorker` uses a filesystem path)
  - Browser consumers who bundle `@kreuzberg/wasm` into their own application served same-origin

  ## Environment

  - Browser: Chrome 144.0.x on Linux (also reported on the same Chrome version in #845)
  - `@kreuzberg/wasm` version: 4.9.4 (latest published from `main`)
  - Demo: `docs/demo.html` on branch `fix/demo-hang-version-mismatch`, commit 127c62bfa
  - Test images: 60 KB and 240 KB screenshots containing text; 725-byte gradient PNG from #845 

### Steps to reproduce

 1. Check out `fix/demo-hang-version-mismatch` at commit 127c62bfa (or later). Serve `docs/demo.html`
  from any host that is *not* `cdn.jsdelivr.net` (e.g. `python -m http.server 8000`).
  2. Open the page in Chrome on your OS. Open DevTools → Network and Console.
  3. Upload any image — text-bearing or blank, of any size under 1 MB.
  4. Observe: after ~90 seconds the demo displays `Extraction timed out after 90s. The file may be too
  complex for in-browser processing. Try a simpler file or use the API.`

Image

```text

```

kh3rld:
fixes #845. fixes #854.                                                                  
  
  root cause chain:                                                                        
                                                                      
  1. @kreuzberg/wasm loaded from a CDN constructs the inner OCR worker with new Worker(cdnUrl). Browsers reject cross-origin worker URLs with a SecurityError, which createOcrWorker silently swallows, setting useFallback = true. OCR then runs synchronously on the worker thread.                                 
  2. Tesseract PSM_AUTO on small or near-uniform images enters a degenerate loop. With the sync fallback active, the worker's event loop is blocked,... the inner 20s setTimeout can never fire.
  3. The demo previously called the extraction pipeline directly on the main thread, so a hung Tesseract froze the tab entirely.                                                   
  
  fixes:                                                                                   
                                                                      
  - worker-bridge.ts - createBrowserWorker now constructs a one-line Blob (import "<cdn-url>";) and passes the Blob URL to new Worker(). The browser accepts the same-origin Blob; the module import inside fetches the real CDN script via CORS. The inner OCR worker is now actually created and its 20s timeout is armed.
  - demo.html - extraction runs in a dedicated Web Worker (keeps the UI thread free). The 30s watchdog starts only after the worker signals WASM is ready, so cold CDN loads (15–25s) no longer consume the extraction budget.
                                                                                           
_kindly note: the PSM_AUTO degenerate-input behavior in kreuzberg-tesseract is a separate issue and should be tracked independently_


v-tan:
Tested the issue - the demo still hangs for small image uploads. @kh3rld 

possible reason for the hang:
```
It's not version drift, not classify(), not Lanczos3. The enableOcr() in the
  published @kreuzberg/wasm@4.9.4 registers NativeWasmOcrBackend (kreuzberg's own tesseract-in-wasm) in
  browsers, which runs in a Web Worker via runOcrInWorker(). That function (line 1494):

  return new Promise((resolve, reject) => {
    pendingRequests.set(id, { resolve, reject });
    workerHandle?.postMessage({ type: "ocr", id, imageData, tessdata, language });
  });

  posts to the worker and waits indefinitely. No timeout. If the worker hangs, the page hangs.

  The worker hangs because Tesseract's page-segmentation analyzes the 725-byte 320×180 gradient PNG,
  which has no text and is a well-known Tesseract degenerate input. With the default PSM (AUTO),
  Tesseract's layout analysis loops on small near-uniform images. That matches the observed symptoms
  perfectly:

  - Renderer ~50% CPU + GPU process ~50% CPU: the OCR Web Worker is computing (Tesseract loop), main
  thread is parked waiting on the postMessage response.
  - Text files work: text path never invokes the OCR worker.
  - Tiny PNG specifically triggers it: Tesseract's PSM AUTO is known to spin on small gradient/blank
  images.
  - PR #847 doesn't help: pinning JS↔WASM versions is orthogonal to a Tesseract input-degeneracy hang
  inside the worker.

  The fix needs to be one or both of these (in crates/kreuzberg-wasm/typescript/ocr/, which lives in a
  separate repo / alef-generated, not in this repo):

  1. Timeout in runOcrInWorker — race the worker promise against setTimeout(reject, 30_000). Even if
  Tesseract spins, the user gets a clean error in 30s instead of a frozen page. This is the must-have.
  2. Set Tesseract PSM to 6 (uniform block of text) by default in the wasm wrapper, or add a pre-OCR
  check that rejects images below some min size / with insufficient variance. PSM AUTO is dangerous on
  arbitrary user input.
```

kh3rld:
@v-tan kindly check this updates here!

Issue #830, opened 2026-04-27
bug: `extraction_timeout_secs` silently ignored in single-file extraction paths


kh3rld:
### Description

ExtractionConfig.extraction_timeout_secs` is only enforced in the batch extraction path (`batch_extract_file` / `batch_extract_bytes`). Single-file extraction via `extract_file` / `extract_bytes` (used by Python bindings, Node bindings, and the MCP `extract_file` / `extract_bytes` tools) silently ignores the timeout. Extraction can hang indefinitely.

### Steps to reproduce

```python
import asyncio, kreuzberg, time

async def main():
    cfg = kreuzberg.ExtractionConfig(extraction_timeout_secs=2)
    start = time.time()
    # This should raise after 2 seconds; in practice it never does
    result = await kreuzberg.extract_bytes(large_slow_pdf_bytes, "application/pdf", cfg)
    print(f"Completed in {time.time()-start:.1f}s — no timeout raised")

asyncio.run(main())
```

**Expected:** `kreuzberg.ExtractionTimeoutError` after ~2 seconds  
**Actual:** Extraction runs to completion regardless of `extraction_timeout_secs`

**Fix direction:** Apply `tokio::time::timeout` (using `config.extraction_timeout_secs`) inside `extract_bytes` and `extract_file`, or route all single-file calls through `run_timed_extraction`.

**Impact:** Users who set `extraction_timeout_secs` to protect against slow/hostile documents will find the timeout non-functional on the most common single-file API paths. PDF hang (#789) recurs for MCP callers.

**Regression?** Unknown,..  timeout may never have worked for single-file paths.

### Relevant files and configuration

```text
- `crates/kreuzberg/src/core/extractor/batch.rs` 
- `crates/kreuzberg/src/core/extractor/bytes.rs`
- `crates/kreuzberg/src/core/extractor/file.rs` 
- `crates/kreuzberg/src/service/extraction.rs`
```

Goldziher:
Fixed by #839 (merged to main, commit 8d8ca6d3a). Timeout now applies to single-file `extract_file`/`extract_bytes` paths under the `tokio-runtime` feature. Will ship in v4.10.0; currently rolling out via v4.10.0-rc.* prereleases.

PR #929, opened 2026-05-09
fix(core): return Validation error for extraction timeout without tokio-runtime

### LINKED ISSUE #911: bug: extraction_timeout_secs silently discarded on WASM and non-tokio builds
### Description

On WASM and non-tokio builds, `extraction_timeout_secs` is silently discarded with `let _ = config.extraction_timeout_secs`. No warning is emitted. Callers who configure a timeout believe they are protected against runaway extractions,.. they are not. A malicious document can hang indefinitely.

### Steps to reproduce

```rust
// WASM build with timeout configured
let config = ExtractionConfig {
    extraction_timeout_secs: Some(5),
    ..Default::default()
};
// Extract a document designed to hang
// → no timeout fires, no log message, hangs indefinitely
```

- Fix: emit `tracing::warn!` when `extraction_timeout_secs.is_some()` on non-tokio paths.
  Document in WASM README and `ExtractionConfig` rustdoc.

```text
crates/kreuzberg/src/extractor/bytes.rs
```

kh3rld:
Resolves #911.

Silently dropping the timeout configuration in WASM or non-tokio builds led to hanging threads and a false sense of security. Now, passing a timeout in these environments explicitly errors with `KreuzbergError::Validation`.

- Added test for non-tokio timeout rejection.
- Ensured existing timeout tests are gated behind the `tokio-runtime` feature.


Goldziher (changes requested):
All items below are required before merge.

### 1. `otel` telemetry bypass — `bytes.rs:130-136` / `file.rs:144-150`

The early `return Err(KreuzbergError::Validation { ... })` short-circuits the `#[cfg(feature = "otel")] if let Err(ref e) = result { ... }` block below. `otel` doesn't depend on `tokio-runtime` (per `Cargo.toml`), so this combo is reachable in the wild — Validation errors will silently miss telemetry. Replace with:

```rust
#[cfg(not(feature = "tokio-runtime"))]
let result = if config.extraction_timeout_secs.is_some() {
    Err(crate::KreuzbergError::Validation {
        message: "extraction_timeout_secs requires the 'tokio-runtime' feature; rebuild with --features tokio-runtime or remove the timeout".to_string(),
        source: None,
    })
} else {
    extraction_future.await
};
```

Also addresses the error-message-actionability nit (mentions the cargo flag).

### 2. Wrong issue number in module doc — `tests/extraction_timeout_tests.rs:1`

Says `//! Regression tests for #830` but PR closes **#911**. Reference #911 (or both, if #911 is a WASM follow-up to the same #830 root cause).

### 3. Non-deterministic 0-timeout tests

`test_extract_bytes_zero_timeout_returns_timeout_error` accepts `Ok(_)` "if synchronous text extraction may beat a 0s timeout". The test passes whether the timeout fires or not — it isn't testing what its name claims. Either tighten with a slow-extraction path or rename to reflect what's actually asserted (e.g. `test_extract_bytes_zero_timeout_does_not_panic`).

### 4. Rebase required

Branch is 104 commits behind `main`. The 5 red CI checks are unrelated to your changes — they're pre-existing Dart codegen-staleness errors fixed weeks ago on `main` (`OcrConfig`, `ProcessingStage`, `OcrBackendType` undefined, `invariant_booleans` removed lint). Rebase onto `main` and force-push to get a fresh CI signal on your actual changes.

PR #930, opened 2026-05-10
ci: add Dart native library build for macOS x86_64 (macos-15-intel)

arrrrny:
## Summary

Adds a dedicated CI workflow (`build-dart-native.yml`) to build the Dart native library (`libkreuzberg_dart`) on **macOS x86_64** (`macos-15-intel`) in addition to the existing macOS ARM64 and Linux targets.

Also includes a minor fix to restore the `"full"` feature flag on the kreuzberg dependency in `packages/dart/rust/Cargo.toml`.

## Why

The upstream CI currently only builds the Dart native library on `macos-latest` (ARM64). macOS Intel (x86_64) users cannot use prebuilt binaries and must compile from source, which fails with ONNX Runtime dependency issues on that platform.

## Changes

- **New**: `.github/workflows/build-dart-native.yml` — builds the Dart `kreuzberg_dart` native library on:
  - `macos-15-intel` (x86_64-apple-darwin)
  - `macos-latest` (aarch64-apple-darwin) 
  - `ubuntu-latest` (x86_64-unknown-linux-gnu)
- Runs Dart analysis and tests on the built library
- Uploads the compiled `.dylib`/\.so` as build artifacts


arrrrny:
Replacing with proper feature branch

PR #933, opened 2026-05-11
feat: implement heuristic PDF table extraction gated by config flag

### LINKED ISSUE #897: feat:  Lightweight PDF table extraction without ONNX layout detection models

Kreuzberg's default PDF text extraction serializes table content as inline text, losing row/column structure. The `result.tables` field is always empty unless `LayoutDetectionConfig` is enabled — but enabling it loads the RT-DETR v2 and SLANeXT ONNX models, which add **~12.4 GB RSS**.

For applications that run alongside other memory-intensive processes (e.g. headless browsers, web crawlers), this memory footprint is not viable. Our production environment runs in containers with 4–8 GB RAM budgets, and even on local dev machines the ONNX models cause significant memory pressure.

The result is that there's **no middle ground** between "no table extraction at all" and "load 12+ GB of ONNX models." Many use cases — particularly LLM-based extraction pipelines — just need the tabular data rendered with its structure intact, not pixel-perfect bounding boxes.

### Our workaround

We added a supplemental pdfplumber pass that runs after Kreuzberg's extraction and appends any detected tables as markdown to the output. This works, but it means we're running two separate PDF libraries on every PDF and managing the table rendering ourselves:

```python
import pdfplumber
from kreuzberg import ExtractionConfig, extract_file_sync

EXTRACTION_CONFIG = ExtractionConfig(output_format='markdown')


def _table_to_markdown(table: list[list[str | None]]) -> str | None:
    """Convert a pdfplumber table to a markdown table string."""
    if not table or not table[0]:
        return None

    col_count = len(table[0])
    # Skip columns that are entirely None (merged-cell artefact)
    non_empty_cols = [
        c for c in range(col_count)
        if any(row[c] for row in table if len(row) > c)
    ]
    if not non_empty_cols:
        return None

    rows: list[str] = []
    for row in table:
        cells = [
            (row[c] or '').replace('\n', ' ').strip()
            for c in non_empty_cols if c < len(row)
        ]
        if any(cells):
            rows.append('| ' + ' | '.join(cells) + ' |')

    if len(rows) < 2:
        return None

    header_sep = '| ' + ' | '.join('---' for _ in non_empty_cols) + ' |'
    rows.insert(1, header_sep)
    return '\n'.join(rows)


def _extract_pdf_tables(file_path: str) -> list[tuple[int, str]]:
    """Extract tables from a PDF via pdfplumber."""
    tables = []
    with pdfplumber.open(file_path) as pdf:
        for page_num, page in enumerate(pdf.pages, start=1):
            for table in page.extract_tables():
                md = _table_to_markdown(table)
                if md:
                    tables.append((page_num, md))
    return tables


def extract_with_tables(file_path: str) -> str:
    """Kreuzberg extraction + pdfplumber table supplement."""
    result = extract_file_sync(file_path, config=EXTRACTION_CONFIG)
    content = result.content

    tables = _extract_pdf_tables(file_path)
    if tables:
        parts = [f'<!-- Table from page {n} -->\n{md}' for n, md in tables]
        content += '\n\n## Tables\n\n' + '\n\n'.join(parts)

    return content
```

This adds ~91 MB RSS (pdfplumber) vs ~12.4 GB (ONNX layout detection) — a 136× difference.

### Why LayoutDetectionConfig doesn't work for us

We tested `LayoutDetectionConfig` and it produces correct results. The problem is purely resource consumption:

| Approach | RSS overhead | Table quality |
|---|---|---|
| Default (no layout) | 0 | No tables (`result.tables` always empty) |
| `LayoutDetectionConfig(preset="fast")` | ~12.4 GB | Excellent |
| pdfplumber (our workaround) | ~91 MB | Good (text-layer PDFs) |

The ONNX models are designed for visual layout detection on scanned/image-heavy PDFs, which is powerful but overkill when the PDF has a proper text layer with embedded table structures — which is the common case for government documents, grant applications, regulatory filings, etc.

### Proposed solution

Add a lightweight, non-ONNX table extraction mode that populates `result.tables` for text-layer PDFs without loading the layout detection models. This could be:

1. **A `PdfConfig` option** like `extract_tables=True` (or `table_extraction="text_layer"`) that uses pdfplumber, tabula, or an equivalent text-layer table parser under the hood
2. **A new `LayoutDetectionConfig` preset** like `preset="lightweight"` or `preset="text_only"` that skips the ONNX models entirely and uses heuristic/text-based table detection
3. **Populate `result.tables` by default** for PDFs with a text layer, using the existing PDF text extraction pipeline to detect tabular structures without ONNX

Any of these would let users get structured table data from `result.tables` without the 12 GB memory cost. The key requirement is that **table markdown appears in `result.content`** (when `output_format='markdown'`) so downstream consumers don't need to manually stitch `content` and `tables` together.

### Environment

- Kreuzberg version: 4.6.3
- Python: 3.12
- OS: Linux (x86_64), also deployed on AWS ECS (4–8 GB containers)

---

Allows for a pragmatic middle ground for table generation.

kh3rld:
## Description
This PR implements a lightweight, heuristic-based table detection system for text-layer PDFs, resolving Issue #897. 
By leveraging native spatial detection in `pdf_oxide` and gating it behind a configuration flag, we avoid the overhead of heavy ML models for standard text-layer documents.

### Key Changes
- Added `extract_tables: bool` to `PdfConfig` (defaults to `false`).
- Added `--pdf-extract-tables` flag to CLI overrides.
- Conditioned `extract_tables_native` execution on the new config flag in the PDF extraction pipeline.
- Updated `CHANGELOG.md`.

## Verification
- Configuration integrity verified via `cargo check` and targeted unit tests.
- Polyglot propagation verified via `alef generate`.
- Note: This PR strictly excludes generated binding files as per repository policy.

Fixes #897


kh3rld:
Superseded by the heuristic table extraction implementation already landed in main (60c7193). Issue #897 is resolved.

PR #941, opened 2026-05-12
fix(php): make wrapper installable via plain Composer

### LINKED ISSUE #940: bug: composer require kreuzberg/kreuzberg fails because type: php-ext routes through PIE
```
$ composer require kreuzberg/kreuzberg
PHP extension packages cannot be installed with Composer; please use PIE instead

$ pie install kreuzberg/kreuzberg
# fails at phpize — no config.m4
```

The package on Packagist is `"type": "php-ext"`, which on Composer 2.7+ forces install via PIE. PIE then runs phpize and looks for `config.m4`, but the extension is Rust (`crates/kreuzberg-php`, cargo build) so there's no `config.m4`. Dead end either way.

The Packagist dist also only ships `packages/php/src/` — the wrapper stubs. No Rust source, no `.so`. Nothing PIE could build even if it got past phpize.

To get this working in our stack we ended up building the `.so` ourselves in our Docker image (`cargo build -p kreuzberg-php` from the monorepo) and then bypassing Composer's `php-ext` gate by declaring kreuzberg as a `"type": "package"` repo in our `composer.json`, pinning the git tag, and overriding the autoload to `packages/php/src/`. Works, but the custom repo pulls the full ~200MB monorepo zipball, we have to keep the wrapper version and the extension version in lockstep manually, and security advisory tooling doesn't see the package.

The underlying mismatch is that the wrapper and the extension are two different things sharing one package name. The wrapper is pure PHP and wants to install like any library; the extension is native and wants whatever delivery channel fits (PIE, prebuilt binary, base image). If `packages/php/composer.json` flipped to `"type": "library"` with `ext-kreuzberg` as a suggest, `composer require kreuzberg/kreuzberg` would just work. The extension stays a separate concern.

Happy to send a one-line PR if you want, but figured I'd check first — if you'd rather fix this by getting the PIE path working (config.m4 shim around cargo, prebuilt release binaries) I won't get in the way.

NiekNijland:
`type: php-ext` on Packagist forces Composer to hand off to PIE, which can't build this package — no `config.m4`, the extension is cargo-built. So `composer require kreuzberg-dev/kreuzberg` fails outright on Composer 2.7+.

Flipping the type to `library` lets the wrapper install like any other Composer package. The extension still has to be present at runtime, but that's already true today and `suggest` makes it explicit. The `php-ext.*` block is dropped because it has no effect on `type: library`.

Verified locally: `composer require kreuzberg-dev/kreuzberg` now resolves, autoload picks up `packages/php/src/` correctly, the wrapper calls into the native functions when the extension is loaded and surfaces the existing `KreuzbergException` when it isn't. No behavior change.

Out of scope: making the extension installable via PIE (would need a `config.m4` shim that invokes cargo, or PIE's newer build-cwd hooks), and publishing prebuilt `.so` artifacts per platform. Both fit better as a separate `kreuzberg-dev/kreuzberg-ext` package — happy to do that as a follow-up if you want.

Fixes #940.

### DIFF
--- a/packages/php/composer.json
+++ b/packages/php/composer.json
@@ -2,9 +2,12 @@
   "name": "kreuzberg-dev/kreuzberg",
   "description": "High-performance document intelligence library",
   "license": "Elastic-2.0",
-  "type": "php-ext",
+  "type": "library",
   "require": {
     "php": ">=8.2"
+  },
+  "suggest": {
+    "ext-kreuzberg": "Required at runtime. Build from crates/kreuzberg-php (cargo) or install a prebuilt .so — see https://github.com/kreuzberg-dev/kreuzberg/releases"
   },
   "require-dev": {
     "phpstan/phpstan": "^2.1",
@@ -23,11 +26,6 @@
     "test": "php vendor/bin/phpunit",
     "lint": "@phpstan",
     "lint:fix": "PHP_CS_FIXER_IGNORE_ENV=1 php vendor/bin/php-cs-fixer fix && php -d detect_unicode=0 vendor/bin/phpstan --configuration=phpstan.neon --memory-limit=512M"
-  },
-  "php-ext": {
-    "extension-name": "kreuzberg",
-    "support-zts": true,
-    "support-nts": true
   },
   "keywords": ["document", "extraction", "pdf", "ocr", "text"]
 }


kh3rld (approved):
looks good!

Goldziher:
Thanks for digging into this. Closing — flipping the type to `library` short-circuits the intended packaging story: this package *does* build native binary assets (`crates/kreuzberg-php` via `ext-php-rs`) and PIE is exactly the install path we want. If PIE 2.7+ is rejecting the package because there's no `config.m4`, that's the packaging bug to track and fix — either a `config.m4` shim that invokes cargo, or PIE's prebuilt-binary mode pointed at our release artifacts. Happy to take a follow-up PR in that direction.

PR #953, opened 2026-05-13
fix(pdf): MCID-tagged content dropped in markdown/html output

### LINKED ISSUE #917: bug: text-layer content inside Marked-Content (MCID) blocks is dropped from PDF extraction in output_format="markdown"/"html" (plain output captures it correctly)
### Description

PDF text-layer extraction loses content rendered inside Marked-Content sequences (BDC/EMC operators with MCID dictionaries) when `output_format` is `"markdown"` or `"html"`. The same characters reach `result.content` correctly when `output_format="plain"` is used. Additionally, `"plain"` triggers OCR on scanned pages of the same PDF while `"markdown"`/`"html"` do not — suggesting these formats route through a separate, more limited extraction pipeline.

Concretely on a 20-page Russian-language PDF (mix of text-layer pages and scanned pages), comparison on the same `OcrConfig(backend="tesseract", language="eng+rus")`:

| `output_format` | `result.content` length | Page-1 title block extracted | Scanned-page text extracted |
| --- | ---: | --- | --- |
| `"markdown"` | 3 777 chars | ❌ — output starts mid-paragraph, full top-of-page block missing | ❌ |
| `"html"` | 4 443 chars | ❌ — same omission, same start point | ❌ |
| `"plain"` | 40 286 chars | ✅ | ✅ — OCR fired on scanned pages without `force_ocr` |

Diagnostics on the affected page (page 1):

- `pypdf.PdfReader.pages[0].extract_text()` returns 1 854 characters including the title block; this is the "ground-truth" content that's actually in the PDF.
- `pypdf.PdfReader.pages[0].images == []`; page `Resources` has no `/XObject`; the title block is not a bitmap. It's text rendered with Type0/TrueType fonts inside `/P <</MCID 0 >>BDC ... EMC` content blocks.
- 8 fonts on the page, 22 annotations.
- Kreuzberg's markdown pipeline returns 1 376 characters for that page — consistent regardless of whether `force_ocr` is set or not, indicating the loss is in the text-layer-to-markdown conversion, not in OCR.

Forcing OCR via `force_ocr_pages=[1]` causes Kreuzberg to render and OCR the page; the title block then appears (with normal Tesseract artifacts on uppercase Russian letters). So the rasterised representation is fine, the original text is in the PDF, and `pdfium` has it — but the markdown/html extractor specifically skips the marked-content blocks.

This appears related-in-spirit to closed issue #431 (markdown garbles positioned text), but this case is plain prose rendered via MCID-tagged blocks rather than positioned/tabular text, and the markdown/html extractors completely drop it instead of garbling.

### Steps to reproduce

Take any PDF whose body text is wrapped in Marked-Content sequences (typical for tagged PDFs targeting accessibility / PDF-UA). Tested on a real-world Russian academic publication; happy to provide the file privately if useful.

```python
import asyncio
import pypdf
from kreuzberg import (
    ExtractionConfig, OcrConfig, PageConfig, extract_file,
)

PDF = "fixture.pdf"
print("pypdf chars (page 1):",
      len((pypdf.PdfReader(PDF).pages[0].extract_text() or "")))

async def run():
    for fmt in ("markdown", "html", "plain"):
        cfg = ExtractionConfig(
            pages=PageConfig(extract_pages=True, insert_page_markers=True),
            output_format=fmt,
            ocr=OcrConfig(backend="tesseract", language="eng+rus"),
        )
        r = await extract_file(PDF, mime_type="application/pdf", config=cfg)
        print(f"output_format={fmt!r}: result.content chars =",
              len(r.content or ""))

asyncio.run(run())
```

Expected: all three `output_format` values surface the same character coverage (modulo formatting).

Actual: `markdown`/`html` return roughly 4 KB without the marked-content block and without OCR; `plain` returns ~40 KB with the marked-content block and with OCR fired on scanned pages.

- Kreuzberg 4.9.4 from PyPI (latest available there as of 2026-05-08; GitHub releases have 4.9.5 / 4.9.6 but no Python wheels yet)
- Python 3.13.9 on Windows 11; reproduces identically on Python 3.12 in \`python:3.12-slim\` Docker
- Repro PDF available on request — it's an OCR-relevant 982 KB Russian textbook scan, public-domain in our jurisdiction; can share via gist or email
- I can also try to construct a synthetic minimum-repro PDF with hand-rolled MCID blocks if that helps narrow it down — happy to follow up

kh3rld:
Two independent failure chains both originate from the needs_structured flag that activates the pre-rendered document path for markdown/djot/html output.

  **root cause a: furniture over-marking (`classify.rs`)**
  mark_cross_page_repeating_text and mark_cross_page_repeating_short_text had no first-occurrence exemption. A title block whose text also appears as a running header on subsequent pages was furniture-marked on all pages, including the first, silently dropping it from the output. pdf_oxide's own `mark_running_artifact_spans` already has this exemption,.. this change aligns kreuzberg's classifier with it. Three existing test assertions were encoding the buggy behaviour and have been corrected.

  **root cause b: OCR suppressed on mixed documents (`mod.rs, ocr.rs`)**
  `has_substantive_doc = true` (enough aggregate native text) was sufficient to skip OCR entirely, even when `decision.fallback = true` (a per-page quality check had flagged a scanned page). Scanned pages in otherwise text-layer documents never received OCR. Fix: gate becomes `has_substantive_doc && !decision.fallback`. The decision logic is extracted into` ocr::evaluate_ocr_skip_gate` so it can be tested without the async pipeline; two tests cover both the bug and its non-regression.

  Why both fixes ship together:  Fix #917 by including both components needed to reproduce the end-to-end behavior: a document with a title-as-header and a scanned page.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,20 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- **MCID-tagged PDF content dropped in markdown/html output**: two
+  independent failure chains in the structured-PDF path caused content loss.
+  (1) `mark_cross_page_repeating_text` and `mark_cross_page_repeating_short_text`
+  had no first-occurrence exemption — a title block whose text also appeared as
+  a running header on later pages was furniture-marked on every page including
+  the first, silently dropping it from output. Fixed by preserving the first
+  occurrence of any repeating span, matching the exemption already present in
+  pdf_oxide's own `mark_running_artifact_spans`. (2) The OCR skip gate
+  incorrectly suppressed OCR when `fallback=true` (a per-page quality check
+  detected a scanned page): the gate now runs OCR whenever `fallback=true`,
+  regardless of aggregate character count. OCR is still skipped when the
+  pre-rendered doc is substantive and no per-page fallback is needed, or when
+  the content is non-textual.
+
 - **Email encoding data corruption (#910)**: replaced brittle 4-byte heuristic
   for UTF-16 detection in `EmailExtractor` with a robust statistical approach
   using `chardetng`. Tiny 4-byte ASCII files (e.g., binary sequences with
--- a/crates/kreuzberg/src/extractors/pdf/mod.rs
+++ b/crates/kreuzberg/src/extractors/pdf/mod.rs
@@ -289,32 +289,32 @@ impl PdfExtractor {
                 1.0
             };
 
-            let has_substantive_doc = pre_rendered_doc.is_some()
-                && total_chars >= thresholds.substantive_min_chars
-                && alnum_ws_ratio >= thresholds.alnum_ws_ratio_threshold;
-
-            let skip_ocr_for_non_text = pre_rendered_doc.is_some()
-                && total_chars >= thresholds.non_text_min_chars
-                && alnum_ws_ratio < thresholds.alnum_ws_ratio_threshold;
-
-            if skip_ocr_for_non_text {
-                tracing::debug!(
-                    alnum_ws_ratio,
-                    total_chars,
-                    alnum_ws_chars,
-                    "Skipping OCR: content is non-textual and pre-rendered structured doc available"
-                );
-                (native_text, ExtractionMethod::Native)
-            } else if has_substantive_doc {
-                tracing::debug!(
-                    total_chars,
-                    alnum_ws_ratio,
-                    ocr_fallback = decision.fallback,
-                    "Skipping OCR: pre-rendered structured doc available with substantive native text"
-                );
-                (native_text, ExtractionMethod::Native)
-            } else if decision.fallback {
-                match run_ocr_with_layout(content, config, path).await {
+            match ocr::evaluate_ocr_skip_gate(
+                pre_rendered_doc.is_some(),
+                total_chars,
+                alnum_ws_ratio,
+                decision.fallback,
+                &thresholds,
+            ) {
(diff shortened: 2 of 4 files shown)


Goldziher (changes requested):
## Request changes — PR will not compile

`crates/kreuzberg/src/extractors/pdf/mod.rs` calls `ocr::evaluate_ocr_skip_gate` and `ocr::OcrGateOutcome`, but neither exists in `crates/kreuzberg/src/extractors/pdf/ocr.rs` (verified via grep on both this branch and current `main`). The implementation must have been left out of the commit. This is the root cause of the 10 failing platform builds — not stale-base drift.

### Furniture-marking fix (classify.rs)
Conceptually sound. The `first_seen_page` tracking + first-occurrence exemption mirrors the established pattern in pdf_oxide's `mark_running_artifact_spans`. The three existing assertions you flip from `true` → `false` are correctly identified as encoding the bug, and the new regression tests assert exact behavior across every page. Good.

### Required before merge
1. **Include the `ocr::evaluate_ocr_skip_gate` + `OcrGateOutcome` implementation** — currently the production code references symbols that don't exist anywhere. Add them to `crates/kreuzberg/src/extractors/pdf/ocr.rs`.
2. **Rebase onto current `main`** — this branch is 85+ commits behind (alef bindings, layout-types/auto-rotate-types/paddle-ocr-types feature groups, android-target, test_documents submodule, etc.). CI signal won't be trustworthy without it.

Once the missing implementation is in and the branch is rebased, the furniture-marking half is good to go and the OCR-gate refactor (matched against the new tests) should land cleanly.

PR #954, opened 2026-05-13
fix(djot): skip orphaned Image elements when extract_images=false

### LINKED ISSUE #914: chore: no regression test for image data leak guard when extract_images=fals
### Description

The guard at `extractors/pdf/extraction.rs` correctly prevents image data from appearing in `ExtractionResult` when `extract_images=false`. However there is no test asserting this behavior,... a future refactor could silently regress it and leak image data into structured output with nothing catching it.

### Steps to reproduce

Remove the guard at `extractors/pdf/extraction.rs` and run the test suite,... no test fails, demonstrating the gap.

- Fix: add a test asserting `result.images.is_empty()` when `extract_images=false`

```text
crates/kreuzberg/src/extractors/pdf/extraction.rs
```

kh3rld:
The Djot renderer unconditionally emitted `` for `ElementKind::Image` elements with no matching entry in `doc.images`. This is the state produced when `extract_images=false`, `inject_placeholders` is false so no images are populated, but if the extraction-level guard were ever loosened, orphaned elements would leak markup into Djot content.

  All other renderers (comrak_bridge, html_styled, plain) already had a `doc.images.get()` None guard. djot.rs was the only gap.

  ## changes

  **`djot.rs`** - add `if image.is_none() { continue; }` before any allocation in the `Image` match arm, matching the pattern in the other renderers.

  **`pdf_image_extraction_tests.rs`** - add:
  - A renderer-level unit test that constructs `InternalDocument` directly with an orphaned `Image` element (empty `doc.images`), bypassing the extraction gate entirely, the only test that proves the renderer fix independently.
  - Three end-to-end contract tests asserting `result.content` contains no image markup when `extract_images=false`, covering Djot/Markdown output and both config paths (`ImageExtractionConfig` and `PdfConfig`).

  fixes #914

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   pre-rendered doc is substantive and no per-page fallback is needed, or when
   the content is non-textual.
 
+- **Djot renderer emitted dangling `![]()` for orphaned image elements (#914)**:
+  when an `ElementKind::Image` referenced an index outside `doc.images`, the
+  Djot renderer fell through to its normal output path and produced empty
+  image markup. Aligned with the existing `comrak_bridge` behavior: skip the
+  element only when both the image lookup is `None` and the description text
+  is empty. When alt text is present, it is still emitted so user-visible
+  content is preserved.
+
 - **Email encoding data corruption (#910)**: replaced brittle 4-byte heuristic
   for UTF-16 detection in `EmailExtractor` with a robust statistical approach
   using `chardetng`. Tiny 4-byte ASCII files (e.g., binary sequences with
--- a/crates/kreuzberg/src/rendering/djot.rs
+++ b/crates/kreuzberg/src/rendering/djot.rs
@@ -95,11 +95,15 @@ pub(crate) fn render_djot(doc: &InternalDocument) -> String {
             }
             ElementKind::Image { image_index } => {
                 let image = doc.images.get(image_index as usize);
-                // Skip orphaned elements — doc.images has no entry when extract_images=false.
-                if image.is_none() {
+                let desc = image
+                    .and_then(|img| img.description.as_deref())
+                    .unwrap_or(elem.text.as_str());
+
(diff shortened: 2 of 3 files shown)


Goldziher (changes requested):
## Request changes — premise is incorrect, claimed tests not in tree

### The framing in the PR body is wrong
- "djot.rs was the only renderer missing a `doc.images.get()` None guard" — current `djot.rs` already calls `doc.images.get(image_index as usize)` at lines 97–107 and chains `.and_then()`. No `None` panic exists; on `None` it emits ``. The actual change you're making is adding `continue` on `None`, not adding a guard.
- "skip orphaned Image elements when `extract_images=false`" — renderers never see `ExtractionConfig`. `extract_images` is consumed upstream in the extractor layer. The renderer-level fix has no relationship to that flag.

The fix is still defensible (drop dangling `` from djot output) but please correct the PR body so the rationale matches the code.

### Behavioral inconsistency with sibling renderers
`comrak_bridge.rs` (~lines 679–685) skips only when `image == None && desc.is_empty()` — when `desc` is non-empty it still emits the description with an empty URL. This PR's `if image.is_none() { continue; }` is more aggressive: it drops the element even when `desc` carries text, silently losing user-visible content.

Match `comrak_bridge` semantics:
```rust
if image.is_none() && desc.is_empty() { continue; }
```

### Claimed tests are not in the branch
The PR body lists "1 unit test at the renderer level + 3 end-to-end contract tests" — none of them exist in `djot.rs`'s test module (lines 285–937) on this branch. Looks like they were described in the PR body but not committed. Either push them, or remove the claim and write the renderer-level test now — mirror the pattern in `comrak_bridge.rs:1318` (`test_image_out_of_bounds_index_dropped`).

### Required before merge
1. Match `comrak_bridge` semantics on `None`.
2. Add the renderer-level out-of-bounds test that's claimed in the body.
3. Correct the PR body so the title and root-cause description reflect what the code actually does.
4. Rebase onto current `main` (85+ commits behind — most of the 16 failing checks should clear once rebased).

PR #986, opened 2026-05-15
fix(pdf): tighten glyph-fragmentation heuristic; prevent false positives

### LINKED ISSUE #962: bug: PDF text extracted one character per line on glyph-spaced text (Word "image not found" placeholder)
### Description

# Bug report: PDF text extracted one character per line on glyph-spaced text (Word "image not found" placeholder)

## Summary

When a PDF contains text where individual glyphs are positioned with their
own bounding boxes (instead of as a single text run), kreuzberg's PDF text
extractor emits each glyph as its own line of output. The result is content
of the form:

```
H
e
t

a
f
b
e
e
l
d
i
n
g
o
n
d
e
r
d
e
e
l
m
e
t
r
e
l
a
t
i
e
-
i
d
r
I
d
9
6
i
s
n
i
e
t
a
a
n
```

…which is the Dutch string `Het afbeeldingonderdeel met relatie-id rId96 is niet aan…`
(`The image part with relationship id rId96 is not found …`) fragmented one
character per line. This is the placeholder Microsoft Word leaves in a PDF
export when an embedded image's relationship is broken.

Downstream consequences for our application:

- The fragmented text pollutes search/retrieval — no semantic value, but it
  occupies a chunk and contributes to the embedding.
- The PDF chunker sometimes splits the artifact across two chunks, so
  reattribution to the correct section header is degraded.
- A chunk whose entire body is this artifact still consumes an embedding
  call and a row in the vector store.

Affected: `kreuzberg = 4.9.6` (rust crate), feature `pdf`, default
extraction (any `ResultFormat`).

## Environment

- kreuzberg crate `4.9.6`
- features: `pdf`, `office`, `excel`, `email`, `chunking`, `html`, `xml`, `archives`
- PDF: Microsoft Word export with a broken image relation (relationship id
  not found in the document parts). The placeholder text "Het afbeelding…
  is niet aangetroffen in het bestand" is rendered glyph-by-glyph, each
  glyph in its own positional command, often near (but not exactly on) the
  same baseline.
- Host: Linux x86_64 via `ghcr.io/kreuzberg-dev/kreuzberg:latest`. Same
  behaviour observed on macOS arm64 with the native crate.

## Reproduction

Sample PDF available on request.
We can also share a minimal Word document configured to export with a
broken image relation that reproduces the same fragmentation.

Minimal Rust reproducer:

```rust
use kreuzberg::{
    extract_bytes, ChunkerType, ChunkingConfig, ExtractionConfig, OutputFormat,
    PageConfig, PdfConfig,
    types::OutputFormat as ResultFormat,
};

#[tokio::main]
async fn main() -> anyhow::Result<()> {
    let bytes = std::fs::read("doc_with_broken_image.pdf")?;
    let cfg = ExtractionConfig {
        chunking: Some(ChunkingConfig {
            chunker_type: ChunkerType::Markdown,
            max_characters: 1200,
            overlap: 150,
            ..Default::default()
        }),
        pages: Some(PageConfig { extract_pages: true, ..Default::default() }),
        pdf_options: Some(PdfConfig { extract_annotations: true, ..Default::default() }),
        output_format: OutputFormat::Markdown,
        result_format: ResultFormat::ElementBased,
        ..Default::default()
    };
    let res = extract_bytes(&bytes, "application/pdf", &cfg).await?;
    // Look for any element whose text starts with a long single-char-per-line run.
    for el in res.elements.unwrap_or_default() {
        let single_char_lines = el
            .text
            .lines()
            .filter(|l| l.trim().chars().count() == 1)
            .count();
        if single_char_lines >= 5 {
            println!("Fragmented element on page {:?}: {} single-char lines",
                el.metadata.page_number, single_char_lines);
            break;
        }
    }
    Ok(())
}
```

Equivalent CLI invocation (against the public Docker image):

```bash
docker run --rm -v "$PWD/doc_with_broken_image.pdf:/work/in.pdf:ro" -w /work \
    ghcr.io/kreuzberg-dev/kreuzberg:latest \
    extract -f json --disable-ocr true --no-cache true in.pdf
```

The fragmented placeholder appears in `content` as a multi-line block where
each line is a single character.

## Expected behaviour

Glyphs positioned at the same (or near-same) baseline should be grouped
into a single logical line of text, regardless of whether the PDF source
renders them via one text-show operator or many. Other PDF text extractors
(pdftotext from poppler, for example) coalesce same-line glyphs based on
y-coordinate proximity, producing the intended string
`Het afbeelding onderdeel met relatie-id rId96 is niet aangetroffen in het bestand`
(or, for the broken-image case, the same as a single noisy line that
downstream filters can recognise and drop).

A reasonable threshold: when two consecutive glyphs share a baseline
within, say, 0.5× the line-height of the document's median font, they
should be the same line. This is the same heuristic that already prevents
fragmentation of normal body text — it just isn't being applied to the
glyph-spaced placeholder case.

## Actual behaviour

Each glyph is emitted as its own line in the element's `text` field and in
the page-content output. With chunking enabled, the artifact also crosses
chunk boundaries because `max_characters` is reached mid-fragmentation.

## Hypothesis

The PDF text-extraction pass likely uses the source PDF's text-show
operators as the unit of grouping rather than re-grouping by baseline
geometry. When a PDF places each glyph via a separate operator (which
some Word exports do for embedded text-like placeholders, and which
several other PDF producers do for graphical text), the per-operator
grouping leaves each glyph on its own line.

## Impact

The placeholder text appears whenever a Word-exported PDF references an image whose relationship is
broken — likely common across corporate / governmental documents that
went through Word export with a missing or renamed image asset.

### Steps to reproduce

-

```text

```

kh3rld:
Fixes #962. Supersedes the pdfium-path implementation from the first push; implementation now lives in the pdf_oxide backend as the reviewer suggested.

**Root cause:** Word-exported PDFs position each glyph via its own `BT…ET` block with a sinusoidal y-jitter (~3–5 pt). pdf_oxide's ColumnAware reading order groups spans by y-level rather than reading order, producing single-character spans out of sequence — one character per line.

**Detection** (`is_fragmented_span_list`): ≥ 3 same-line x-disorder events among short (≤ 3 char) spans. "Same-line" uses a 5 pt absolute ceiling (`MAX_GLYPH_JITTER_PT`) instead of a font-size fraction. The old `font_size * 0.25` fallback was at the jitter amplitude boundary, triggering false positives on height-zero span lists from normal documents.

**Reconstruction** (`rebuild_text_from_fragmented_spans`): sort by y-descending, group by 5 pt y-proximity, re-sort each group by x-ascending, insert spaces where x-gap > `font_size * 0.5`.

**Upstream TODO** added referencing pdf_oxide issue #518 — if pdf_oxide's `Tm` continuation check is fixed to tolerate the full jitter range, this heuristic can be removed.

**Tests:** 11 integration tests (programmatic PDF fixtures, no binary files) + 6 unit tests covering boundary conditions at exactly `MIN_DISORDER_COUNT - 1` and `MIN_DISORDER_COUNT` events. Includes negative regression: genuine single-char-per-line content (20 pt y-spacing) is not collapsed.

**Reviewer items addressed:**
1. False-positive guard: `MAX_GLYPH_JITTER_PT = 5.0` absolute ceiling replaces the `font_size * 0.25` fraction that overlapped with normal leading.
2. Negative regression test: `test_genuine_single_char_lines_not_collapsed` — poetry/CJK/code columns with 20 pt y-gaps round-trip unchanged.
3. Positive regression test: `test_coalesced_content_is_coherent` — programmatic Word-jitter PDF fixture, no binary required.
4. Unrelated file changes removed (pool.rs, kreuzberg-wasm) — clean 4-file diff.
5. `.typos.toml` entry explains the Dutch `relatie` comes from issue #962 test comments.


Goldziher (changes requested):
The detection heuristic (`is_fragmented_span_list`) and reconstruction (`rebuild_text_from_fragmented_spans`) in `crates/kreuzberg/src/pdf/oxide/text.rs` are reasonable for the Word-export jitter pattern but the 13 failing CI checks (Lint, Rust macos/arm, iOS cargo check, 8 of 13 E2E suites) indicate the heuristic mis-fires on real test corpus.

Concrete asks before this can merge:
1. False-positive guard: `MIN_DISORDER_COUNT = 2` with the trigger `cur.bbox.x < prev.bbox.x - prev.font_size` will fire on any page with a column break, a list item that wraps, or a centred title followed by left-aligned body. Tighten the same-line test (`y_gap < eff_height * 0.5`) — currently `eff_height = font_size * 0.5` fallback when heights are zero, which makes the threshold `font_size * 0.25` and treats almost any pair as "same line". A height-zero span list with normal y-deltas (e.g. 14 pt line height) should *not* be classified as same-line.
2. Add a negative regression test: a PDF with genuine single-character lines (formula, poetry, code listing, vertical CJK) must round-trip unchanged through the reconstruction path. Failures in `E2E (java/kotlin_android/dart/swift/wasm/ruby/r/elixir)` and `Rust (macos/arm)` likely indicate exactly this class of false positive on existing fixtures.
3. Add a positive regression test: include a fixture PDF reproducing #962's per-glyph BT/ET fragmentation and assert the reconstructed text matches the expected paragraphs. Currently there is no `test_documents/pdfs/issue_962_*.pdf` and no new test referenced.
4. `crates/kreuzberg/src/utils/pool.rs` and `crates/kreuzberg-wasm/src/lib.rs` show in the diff but the change is unrelated to PDF glyph reconstruction — if these were edited by hand, restore them (alef-generated); if intentional, split into a separate PR.
5. CHANGELOG mention `relatie` in `.typos.toml`: ok but the entry should explain *which* fixture introduces the Dutch word, not just that issue #962 has it.

Goldziher:
Isn't this a bug in PDF oxide? We should evaluate. If any part of it is a bug there, we should open a GitHub issue on their repo and document this with to-do comments in our code. @kh3rld - please do so

Goldziher:
Re-reviewed and confirming @Goldziher's `CHANGES_REQUESTED` is still blocking. The 13 failing CI checks are split: the broad Rust-macos / Android / iOS failures are due to an out-of-date base (the `frb_generated.rs` `Option<DocxAppProperties>` / `CoreProperties` mismatch was fixed on `main` after this PR branch was cut — a rebase will clear those). The remaining concerns are still real:

1. **False-positive guard on `is_fragmented_span_list` is too loose.**
   - `MIN_DISORDER_COUNT = 2` with the trigger `cur.bbox.x < prev.bbox.x - prev.font_size` fires on column breaks, wrapping list items, and centred-title-then-left-body layouts.
   - The same-line test uses `y_gap < eff_height * 0.5` where `eff_height = font_size * 0.5` is the height-zero fallback. That collapses the threshold to `font_size * 0.25`, treating almost any close-y-pair as same-line. A height-zero span list with normal 14pt line-height deltas should *not* be classified as same-line.
   - Add the threshold/floor as named constants in `pdf/structure/constants.rs` with measured justification, not magic literals at the call site.

2. **Negative regression test missing.** A PDF with genuine single-character lines (formula, poetry, code listing, vertical CJK) must round-trip unchanged through the reconstruction path. Add such a fixture + test before flipping the heuristic on.

3. **Positive regression test missing.** Add `test_documents/pdfs/issue_962_*.pdf` reproducing the per-glyph BT/ET fragmentation and assert the reconstructed text matches the expected paragraphs. Currently there is no fixture for #962 and no test references one.

4. **Scope leak.** `crates/kreuzberg/src/utils/pool.rs` and `crates/kreuzberg-wasm/src/lib.rs` are touched in this diff but unrelated to PDF glyph reconstruction. If hand-edited, restore (the wasm file is alef-generated). If intentional, split into a separate PR.

5. **CHANGELOG / `.typos.toml`.** The `relatie` typo allow-list entry should name which fixture introduces the Dutch token, not just point at #962.

Diff is +676 lines for what should be a heuristic-tightening fix — please justify the scope in the description (new module vs. fixture additions vs. parallel heuristic path), since the surface area of the change determines reviewer load and false-positive risk.

Once 1–5 are addressed and the PR is rebased on current `main`, CI failures should collapse to the existing arm flake (`monitoring::tests::test_snapshot_collection`, preexisting on main).

kh3rld:
**Changes in this update:**

1. **Constants moved to `pdf/structure/constants.rs`** (`pub(crate)`) - `MAX_GLYPH_JITTER_PT`, `MIN_DISORDER_COUNT`, `COALESCE_THRESHOLD` now live alongside other PDF spatial analysis thresholds, each with measurement justification. `oxide/text.rs` imports them via `use crate::pdf::structure::constants::*`.

2. **`.typos.toml` entry updated** - now names the specific file (`crates/kreuzberg/tests/pdf_glyph_spacing_issue_962.rs`) and explains the word's origin ("Dutch for relation-id, from Microsoft Word broken-image placeholder").

3. **Test docstring corrected** - module doc in `pdf_glyph_spacing_issue_962.rs` said "≥ 5 same-line x-disorder events" but `MIN_DISORDER_COUNT = 3`; fixed to match the actual value and links to `pdf::structure::constants`.

4. **CHANGELOG fixed** -`#962` entry moved from the `[5.0.0-rc.1]` release section to `[Unreleased] → Fixed` where it belongs; description expanded to reference the test file path, constants module, and upstream pdf_oxide issue #518.

5. **Rebased on current `main`** - clears the `frb_generated.rs` `DocxAppProperties`/`CoreProperties` mismatch that was failing the macOS/Android/iOS CI jobs.

The heuristic values (`MIN_DISORDER_COUNT = 3`, `MAX_GLYPH_JITTER_PT = 5.0`) are unchanged - the reviewer's first-review values (MIN=2, fraction fallback) were already addressed in the prior push; this update only resolves the structural feedback (constants file location, CHANGELOG, typos entry).

PR #988, opened 2026-05-16
fix(pdf): eliminate unconditional image decompression pass

### LINKED ISSUE #985: bug: populate_images_from_pdfium` ignores `max_images_per_page` cap (regression of #789)
### Description

# Bug: `populate_images_from_pdfium` ignores `max_images_per_page` cap (regression of #789)

**Kreuzberg version:** 4.9.6 (still reproducible — also seen in 4.9.3 and 4.9.4)
**Related closed issues:**
- [#766](https://github.com/kreuzberg-dev/kreuzberg/issues/766) — first attempt at fixing extract_bytes hangs
- [#789](https://github.com/kreuzberg-dev/kreuzberg/issues/789) — second attempt, **closed completed 2026-04-26**, added the `max_images_per_page` field to `ImageExtractionConfig`

**Severity:** High — `extract_bytes` still hangs (and is killed only by an external wall-clock timeout) on PDFs with many image objects per page, despite the cap field added by #789.

## Summary

The `max_images_per_page` cap introduced by #789 limits which images are
included in chunk data — that part works. But the decoder loop in
`populate_images_from_pdfium()` (pipeline.rs) still iterates **every** image
object on each page and calls `get_processed_image()` on it, regardless of the
cap. On PDFs where a single visual diagram is stored as hundreds or thousands
of tiny image fragments (Adobe InDesign / Acrobat exports, custom-font CMaps),
this turns each page into a multi-minute hot loop.

The cap therefore acts as a memory-saver (chunks stay small) but not as a
CPU-saver — the expensive part of image extraction still runs. The
`extraction_timeout_secs` cancellation token is only checked between pages,
not between images, so it can't interrupt the hot loop either.

## Reproduction

common pattern: dense pages of small vector diagrams stored
as individual image objects. 

Sample public PDFs that exhibit the same pattern (publisher-hosted, freely
downloadable, same family of installation manuals):
- `https://www.intergas-verwarming.nl/app/uploads/2018/01/Installatiehandleiding-Kombi-kompakt-HR-88147804.pdf` (15 MB)

### Stage 1: cap is applied to chunk data, but positions are still collected

In `bridge.rs`, `objects_to_page_data()` (around line 543):

```rust
if max_images_per_page.is_some_and(|cap| page_image_count >= cap) {
    capped = true;
    *image_offset += 1;      // offset still incremented
    page_image_count += 1;
    continue;                 // skips adding to segments, but...
}
```

The `ImagePosition` entries are built from the `image_offset` counter. Even
though the segment data is skipped, the offset advances. Later,
`extract_heuristic_segments()` returns `all_image_positions` containing
positions for ALL images, including those beyond the cap.

### Stage 4b: all positions are decoded

In `pipeline.rs`, `populate_images_from_pdfium()` (around line 2688) receives
the full `all_image_positions` vector and iterates every page object calling
`get_processed_image()`:

```rust
for obj in page.objects().iter() {
    if let Some(image_obj) = obj.as_image_object() {
        let global_idx = first_idx_on_page + current_image;
        if indices_set.contains(&global_idx)
            && let Ok(dynamic_image) = image_obj.get_processed_image(document)
        {
            // decode → rgba → PNG encode for EVERY matched position
        }
        current_image += 1;
    }
}
```

## Suggested fix

Two independent issues; either alone is a meaningful improvement, together
they fully resolve the hang. Both live in the existing PDF extraction
pipeline — no new public API, no new dependencies.

### Fix 1 — Apply `max_images_per_page` to the decoder loop too

**Where (source of truth):** `crates/kreuzberg/src/pdf/extraction/bridge.rs`,
function `objects_to_page_data()`, around the cap-check branch (~line 543).
This is where the cap is already consulted for chunk-data; the same branch
needs to skip the `ImagePosition` push.

Today:

```rust
if max_images_per_page.is_some_and(|cap| page_image_count >= cap) {
    capped = true;
    *image_offset += 1;      // offset still increments → position is later emitted
    page_image_count += 1;
    continue;                 // skips chunk segment data only
}
```

Proposed:

```rust
if max_images_per_page.is_some_and(|cap| page_image_count >= cap) {
    capped = true;
    page_image_count += 1;
    // Do NOT increment image_offset and do NOT push an ImagePosition;
    // populate_images_from_pdfium then has no work to do for this image.
    continue;
}
```

**Where (alternative point of enforcement):**
`crates/kreuzberg/src/pdf/extraction/pipeline.rs`,
`populate_images_from_pdfium()` (~line 2688). Filter
`all_image_positions` upfront to drop indices beyond the per-page cap before
the per-object loop runs. This is downstream of the source of truth but
cheap if the bridge-side change is non-trivial:

```rust
let positions: Vec<&ImagePosition> = all_image_positions
    .iter()
    .scan((0usize, 0usize), |(prev_page, count), pos| {
        if pos.page_number != *prev_page {
            *prev_page = pos.page_number;
            *count = 0;
        }
        *count += 1;
        Some((*count, pos))
    })
    .filter(|(count_on_page, _)| {
        max_images_per_page.map_or(true, |cap| *count_on_page as u32 <= cap)
    })
    .map(|(_, pos)| pos)
    .collect();
```

This single change alone reduces CPU time on the test files from minutes to
seconds and makes the new field added by #789 effective for the actual hang
it was meant to fix.

### Fix 2 — Check cancellation between individual images, not just between pages

**Where:** same file, `crates/kreuzberg/src/pdf/extraction/pipeline.rs`,
`populate_images_from_pdfium()`. The current cancel check is at the page
boundary (~line 2715):

```rust
for page in pages {
    if cancel_token.is_some_and(|t| t.is_cancelled()) {
        // skip remaining pages — but not the rest of THIS page
    }
    for obj in page.objects().iter() {
        if let Some(image_obj) = obj.as_image_object() {
            let _ = image_obj.get_processed_image(document); // 1–2s per call
        }
    }
}
```

Proposed — add the same check inside the per-object loop:

```rust
for page in pages {
    if cancel_token.is_some_and(|t| t.is_cancelled()) {
        break;
    }
    for obj in page.objects().iter() {
        if cancel_token.is_some_and(|t| t.is_cancelled()) {
            break;
        }
        if let Some(image_obj) = obj.as_image_object() {
            let _ = image_obj.get_processed_image(document);
        }
    }
}
```

`CancellationToken::is_cancelled()` is a cheap atomic load, so per-image
polling is essentially free relative to the 1–2 s `get_processed_image()`
call it gates. This makes `extraction_timeout_secs` actually enforce its
contract on pathological PDFs and is independent of the cap fix above.

### Optional follow-up — make `inject_placeholders` independent of the decoder hot loop

A separate but related ergonomic gap: today the only way to safely opt
out of the decoder loop is to set **both** `images: None` (or
`extract_images: false`) **and** `inject_placeholders: false`. If a caller
asks for placeholders but disables image extraction, the placeholder path
still triggers the same hot loop — that's surprising and worth either
documenting or short-circuiting in
`populate_images_from_pdfium()`: when `extract_images` is false, do not
decode; emit position-only placeholders directly. Out of scope for this
issue but worth flagging while the file is open.

### Steps to reproduce

-

```text

```

kh3rld:
Fixes #985. All review items from both review rounds are now addressed.

**Root cause:** `extract_image_positions` called `doc.doc.extract_images(page_idx)` unconditionally on every page — even when `extract_images=false` — triggering full image decompression before returning. PDFs with many embedded images caused multi-minute hangs for text-only callers.

**Fix summary:**
- Remove `extract_image_positions` entirely. Positions derived from `(img.page_number, img.image_index)` off what `extract_images_with_data` already returns — no second decompression pass.
- Gate `extract_images_with_data` behind `images_extraction_enabled || ocr_inline_images` in `extraction.rs` — text-only extraction never enters the decompression path.
- Add `max_images_per_page == Some(0)` short-circuit at the top of `extract_images_with_data` — zero-cap returns immediately without calling `extract_images()`.
- For positive caps, enumerate the page XObject resource dictionary and stop decompressing after `limit` images (`extract_n_images_from_xobject_resources`). Inline images (`BI`/`EI` operators) fall back to the eager path with `.take(limit)` as a safety guard; full inline-image support tracked in #989.
- Add `cancel_token: Option<&CancellationToken>` checked at the top of the page loop so `extraction_timeout_secs` can interrupt multi-page extraction between pages.
- Remove dead `PoolError::LockPoisoned` — `parking_lot::Mutex` never poisons; `acquire()` and `clear()` are now infallible.

**alef.toml changes (prerequisite, not new features):** The `alef_version` bump to 0.16.71 and the type-exclusion additions (`HwpxExtractor`, `CacheStats`, `OcrBackend`/`register_ocr_backend`/`unregister_ocr_backend`, `process_images_with_ocr`) are required to make `alef generate` complete cleanly. These types caused type-collision and unsupported-type build errors in the FFI and WASM codegen targets. Without excluding them, `alef generate` would fail, making it impossible to commit a fresh alef regen. They are not independent feature additions; they are the prerequisite for the freshness check to pass.

**Review items addressed:**
1. `extract_image_positions` removed; `extract_images_with_data` gated in `extraction.rs:105` — confirmed fixed.
2. Positions derived from the already-extracted `images` vec — no second decompression pass.
3. `max_images_per_page == Some(0)` short-circuits before `extract_images()`. For positive caps, XObject decompression is now bounded at `limit` via resource-dict enumeration; inline images use `.take(limit)` fallback (#989).
4. `PoolError::LockPoisoned` removed; `acquire()` returns infallible `PoolGuard<T>`.
5. `alef_version` corrected to 0.16.71 (was accidentally set to 0.16.32); `alef generate` run — confirmed `lib.rs`/`index.d.ts` match alef output; 26 stale generated files pruned.
6. Added `test_no_decompression_trace_when_images_disabled`: installs a `tracing_subscriber` layer filtered to `kreuzberg::pdf::oxide::images`, runs text-only extraction, and asserts the `decompression_started` event is never emitted — direct structural proof the function was not called.

**Tests:**
- `test_cancellation_stops_extraction_early` — pre-cancelled token returns empty vec immediately
- `test_max_images_per_page_cap_respected_in_output` — cap respected in returned images
- `test_no_images_returned_when_extraction_disabled_on_dense_pdf` — no decompression when disabled
- `test_image_positions_consistent_with_image_data` — image_index uniqueness + orphaned-placeholder detection
- `test_no_decompression_when_images_disabled` — output-based: text-only returns `images: None`
- `test_no_decompression_trace_when_images_disabled` — trace-span: `decompression_started` event absent during text-only extraction

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,6 +46,32 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   Only legitimate WASM/Rust platform-imposed skips remain (7× WASM-async,
   1× Rust-missing-file-path).
 
+### Fixed (#985 pdf image extraction hang)
+
+- **PDF image extraction hang on dense pages (#985)**: `extract_image_positions`
+  previously ran a full decompression pass over every page unconditionally — even
+  when `extract_images=false` — causing multi-minute hangs on PDFs with many image
+  objects per page (e.g. InDesign/Acrobat exports). The pre-pass is eliminated;
+  image positions are now derived from the capped extraction result, so the
+  decompression path is skipped entirely when images are not requested. Also adds
+  cancellation-token support inside `extract_images_with_data` so
+  `extraction_timeout_secs` can interrupt multi-page extraction between pages, and
+  inside the inline-OCR image loop so a timeout can interrupt per-image OCR.
+
+  `max_images_per_page` now bounds decompression for XObject images: the pdf_oxide
+  backend enumerates the page XObject resource dictionary and stops decompressing
+  after `limit` images, avoiding the eager cost of `extract_images()` for images
+  beyond the cap. Inline images (`BI`/`EI` content-stream operators) fall back to
+  the eager path with a `.take(limit)` guard; full inline-image support is tracked
+  in #989.
+
+- **`kreuzberg::utils::pool` API simplified**: `Pool::acquire()` is now infallible
+  — it returns `PoolGuard<T>` directly instead of `Result<PoolGuard<T>, PoolError>`.
+  `PoolError` is removed. `parking_lot::Mutex` cannot poison, so the `Result`
+  wrapper was dead code. `Pool`, `PoolGuard`, `Recyclable`, `StringBufferPool`, and
+  `ByteBufferPool` are marked `#[doc(hidden)]` — they were technically reachable via
+  `kreuzberg::utils::pool` but were never part of the stable public API.
+
 ### Added
 
 - **`kreuzberg` crate root**: plugin_api surface (`list_*`, `clear_*`, `register_*`,
--- a/crates/kreuzberg/src/core/batch_optimizations.rs
+++ b/crates/kreuzberg/src/core/batch_optimizations.rs
@@ -18,7 +18,6 @@
 
 use crate::utils::pool::{ByteBufferPool, StringBufferPool, create_byte_buffer_pool, create_string_buffer_pool};
 use crate::utils::pool_sizing::PoolSizeHint;
-use crate::{KreuzbergError, Result};
 use parking_lot::Mutex;
 use std::sync::Arc;
 use std::sync::atomic::{AtomicBool, Ordering};
@@ -221,20 +220,13 @@ impl BatchProcessor {
     ///
     /// Useful for memory-constrained environments or to reclaim memory
     /// after processing large batches.
-    pub fn clear_pools(&self) -> Result<()> {
-        let pool_opt = self.string_pool.lock();
-        if let Some(pool) = pool_opt.as_ref() {
-            pool.clear()
-                .map_err(|e| KreuzbergError::Other(format!("string pool error: {}", e)))?;
+    pub fn clear_pools(&self) {
+        if let Some(pool) = self.string_pool.lock().as_ref() {
+            pool.clear();
         }
-
-        let pool_opt = self.byte_pool.lock();
-        if let Some(pool) = pool_opt.as_ref() {
-            pool.clear()
-                .map_err(|e| KreuzbergError::Other(format!("byte pool error: {}", e)))?;
+        if let Some(pool) = self.byte_pool.lock().as_ref() {
+            pool.clear();
         }
-
-        Ok(())
     }
 }
 
@@ -276,12 +268,12 @@ mod tests {
         let pool = processor.string_pool();
 
         {
-            let mut s = pool.acquire().unwrap();
+            let mut s = pool.acquire();
             s.push_str("test");
         }
 
         {
-            let s = pool.acquire().unwrap();
+            let s = pool.acquire();
             assert_eq!(s.len(), 0);
         }
     }
@@ -292,12 +284,12 @@ mod tests {
         let pool = processor.byte_pool();
 
         {
-            let mut buf = pool.acquire().unwrap();
+            let mut buf = pool.acquire();
             buf.extend_from_slice(b"test");
         }
 
         {
-            let buf = pool.acquire().unwrap();
+            let buf = pool.acquire();
             assert_eq!(buf.len(), 0);
         }
     }
@@ -306,8 +298,8 @@ mod tests {
     fn test_batch_processor_clear_pools() {
         let processor = BatchProcessor::new();
 
-        let s1 = processor.string_pool().acquire().unwrap();
-        let s2 = processor.byte_pool().acquire().unwrap();
+        let s1 = processor.string_pool().acquire();
+        let s2 = processor.byte_pool().acquire();
 
         drop(s1);
         drop(s2);
@@ -315,7 +307,7 @@ mod tests {
         assert!(processor.string_pool_size() > 0);
         assert!(processor.byte_pool_size() > 0);
 
-        processor.clear_pools().unwrap();
+        processor.clear_pools();
 
         assert_eq!(processor.string_pool_size(), 0);
         assert_eq!(processor.byte_pool_size(), 0);
--- a/crates/kreuzberg/src/extractors/pdf/extraction.rs
+++ b/crates/kreuzberg/src/extractors/pdf/extraction.rs
@@ -92,15 +92,8 @@ pub(crate) fn extract_all_from_oxide_document(
         None
     };
 
-    // --- Image positions for assembly pipeline ---
-    let image_positions = crate::pdf::oxide::images::extract_image_positions(&mut doc).map_err(|e| {
-        crate::error::KreuzbergError::Parsing {
-            message: format!("pdf_oxide image position extraction failed: {e}"),
-            source: None,
-        }
-    })?;
-
     // --- Extracted images (data + OCR) ---
+    // Positions are derived from extracted data — no separate decompression pass needed.
     let images_extraction_enabled = config.images.as_ref().map(|c| c.extract_images).unwrap_or(false)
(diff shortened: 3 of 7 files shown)


Goldziher (changes requested):
The title says "eliminate unconditional image decompression" but `extract_image_positions` in `crates/kreuzberg/src/pdf/oxide/images.rs` still calls `doc.doc.extract_images(page_idx)` unconditionally for every page (see `extraction.rs` around line 95). For the common text-only extraction path this decompresses every embedded image for no reason — i.e. the bug from #985 is not fixed.

Concrete asks before this can merge:
1. When `images_extraction_enabled || ocr_inline_images`, derive `image_positions` from the already-extracted `images` vec instead of calling `extract_image_positions` separately — otherwise `extract_images` runs twice per page.
2. When images are *not* needed, skip `extract_image_positions` entirely (gate behind `needs_structured && inject_placeholders`).
3. The `max_images_per_page` cap is applied via `.take(count)` *after* `extract_images` has already decompressed the full set; either request a `count_images(page_idx)` from `pdf_oxide` or stop reading the iterator early before decompression.
4. `PoolError::LockPoisoned` in `crates/kreuzberg/src/utils/pool.rs` is unreachable with `parking_lot::Mutex` — drop the `Result` wrapper from `acquire()`/`clear()` so callers stop writing dead error-handling arms.
5. CI: 13 failing checks (Lint, Rust macos/arm, iOS cargo check, 8 E2E). The alef-generated `crates/kreuzberg-wasm/src/lib.rs` and `crates/kreuzberg-node/index.d.ts` look hand-touched in this PR — if so, restore them and re-run `alef generate`.
6. Regression test missing: add a fixture asserting that text-only extraction of a PDF with embedded images does not trigger image decompression (measure via a counter or trace span).

Goldziher:
Re-reviewed and confirming @Goldziher's `CHANGES_REQUESTED` is still blocking. Specifically:

1. **The bug from #985 is not fixed.** `extract_image_positions` in `crates/kreuzberg/src/pdf/oxide/images.rs` still calls `doc.doc.extract_images(page_idx)` unconditionally for every page (called from `extraction.rs` ~L95). Text-only extraction still decompresses every embedded image. Gate this call behind `images_extraction_enabled || (needs_structured && inject_placeholders)`.

2. **Duplicate decompression.** When images *are* needed, derive `image_positions` from the already-extracted `images` vec instead of calling `extract_image_positions` separately. Currently `extract_images` runs twice per page.

3. **`max_images_per_page` is applied too late.** `.take(count)` runs *after* the iterator has already decompressed the full set. Either expose `count_images(page_idx)` from `pdf_oxide` or stop reading the iterator before decompression.

4. **`PoolError::LockPoisoned` is unreachable** with `parking_lot::Mutex` — drop the `Result` wrapper from `acquire()` / `clear()` in `crates/kreuzberg/src/utils/pool.rs` so callers stop carrying dead error-handling arms.

5. **alef-generated artifacts in the diff.** `crates/kreuzberg-wasm/src/lib.rs` and `crates/kreuzberg-node/index.d.ts` appear hand-touched. If intentional, that means the next `alef generate` will overwrite them; restore + run `alef generate` instead.

6. **No regression test for the no-images-needed path.** Add a fixture / counter / trace-span assertion that confirms text-only extraction of a PDF with embedded images does *not* trigger image decompression — that's the property #985 wants protected against future regressions.

Also: CI shows the `monitoring::tests::test_snapshot_collection` Rust-arm flake (preexisting on main) — that one can be ignored. The rest of the failing E2Es match the current main baseline.

Once 1–6 are addressed and CI is re-run, this can land.

Goldziher (approved):

Goldziher:
Finish up and merge

PR #994, opened 2026-05-17
fix(pdf): tighten glyph-fragmentation heuristic to prevent false positives

### LINKED ISSUE #962: bug: PDF text extracted one character per line on glyph-spaced text (Word "image not found" placeholder)

kh3rld:
## Summary

Fixes glyph-fragmented PDF text (issue #962) where Word-exported PDFs place each character in its own `BT…ET` block with sinusoidal y-jitter. pdf_oxide's ColumnAware reading order groups spans by y-level rather than reading order, corrupting the output.

**Detection** (`is_fragmented_span_list`): ≥ 3 same-line x-disorder events among spans with ≤ 3 chars, where "same-line" uses an absolute 5 pt ceiling instead of a `font_size` fraction. The 5 pt ceiling matches the known Word jitter amplitude; the old `font_size * 0.25` fallback (≈ 3 pt for a 12 pt font) was at the amplitude boundary, causing false positives.

**Why MIN_DISORDER_COUNT = 3**: pdf_oxide groups consecutive same-y chars into multi-char spans. A 32-char word with period-6 jitter (3 distinct y-levels) produces exactly 4 disorder events — 2 per XY-Cut column at each y-level transition. MIN_DISORDER_COUNT=5 was unreachable; detection never fired. MIN_DISORDER_COUNT=3 is sufficient and, combined with the short-span guard and 5 pt ceiling, is essentially impossible to trigger on normal multi-column text.

**Reconstruction** (`rebuild_text_from_fragmented_spans`): sort by y-descending, group by 5 pt y-proximity, re-sort each group by x, insert spaces where x-gap > `font_size * 0.5`. With actual Helvetica metrics all consecutive span gaps are < 6 pt, so the word concatenates cleanly.

## Addresses reviewer feedback from #986

- **Code review (4303579790)**: tightened heuristic — `MAX_GLYPH_JITTER_PT=5.0` absolute ceiling for zero-height spans replaces the `font_size * 0.5 / 2` fraction that was triggering on normal PDFs; short-span guard filters word-level spans entirely.
- **Thread comment (4469214766)**: added upstream TODO comment for evaluating whether pdf_oxide's ColumnAware mode should absorb per-glyph y-jitter internally.

## Test results

- 11/11 integration tests pass (`tests/pdf_glyph_spacing_issue_962.rs`), including:
  - `test_coalesced_content_is_coherent` — previously failing, now passes
  - `test_genuine_single_char_lines_not_collapsed` — genuine single-char-per-line content is not corrupted
- 6/6 unit tests pass (`pdf::oxide::text::tests`) — boundary tests at exactly N-1 and N disorder events

Fixes #962


kh3rld:
Implementation folded into #986 (rebased, squashed to one commit on current main). Closing this PR.

PR #1017, opened 2026-05-20
feat: mixed-mode OCR extraction (run_ocr_on_images)

sbfkcel:
## Summary

Adds three new fields to \`ImageExtractionConfig\` for controlling per-image OCR rendering, routes image OCR through the plugin backend registry, and fixes \`JsOcrBackendBridge\` so custom Node.js OCR backends work correctly from async tokio threads.

## New fields on \`ImageExtractionConfig\`

| Field | Default | Behaviour |
|---|---|---|
| \`run_ocr_on_images\` | \`true\` | When \`false\`, per-image OCR is suppressed even if an OCR backend is configured |
| \`ocr_text_only\` | \`false\` | Replace \`alt\` image placeholders with OCR text |
| \`append_ocr_text\` | \`false\` | Append an OCR text paragraph after the image placeholder |

All three fields have \`#[serde(default)]\` so existing config files and JSON payloads that omit them continue to deserialize correctly.

## Breaking change — \`ImageExtractionConfig\` struct literal construction

\`ImageExtractionConfig\` is a public struct without \`#[non_exhaustive]\`. Adding three fields is a **compile-time breaking change** for any downstream code that constructs it using an exhaustive struct literal (i.e. without \`..Default::default()\`):

```rust
// breaks after this PR — missing run_ocr_on_images, ocr_text_only, append_ocr_text
let cfg = ImageExtractionConfig {
    extract_images: true,
    inject_placeholders: true,
    ..  // <-- required now
};
```

**Fix:** add \`..Default::default()\` to all \`ImageExtractionConfig\` struct literals. All callsites inside this repo already do this.

## PDF \`RunFallback\` behaviour (opt-in, not a regression)

When a caller explicitly sets \`images.run_ocr_on_images = true\`, the PDF document-level OCR fallback (\`RunFallback\`) is skipped in favour of per-image OCR — native text is preserved and only embedded images are OCR'd.

Callers who do **not** set an \`images\` config are unaffected: the gate defaults to \`false\` when \`config.images\` is \`None\`, so \`RunFallback\` still runs as before.

## Core changes

- **\`extractors/pdf/mod.rs\`**: \`RunFallback\` gate now requires \`config.images.is_some() && run_ocr_on_images = true\` — callers without an \`images\` config see no behaviour change.
- **\`extraction/image_ocr.rs\`**: dispatches through \`get_ocr_backend_registry()\` instead of hard-coding \`OcrProcessor\`; semaphore acquire error is a proper \`Result\` instead of \`.expect()\`.
- **\`core/pipeline/mod.rs\`**: respects the \`run_ocr_on_images\` gate; propagates \`ocr_text_only\` and \`append_ocr_text\` into the internal document before rendering.
- **\`rendering/html_styled.rs\`**: \`render_image\` now takes \`ocr_text_only\`/\`append_ocr_text\` flags — figcaption is only emitted when \`append_ocr_text = true\`; \`ocr_text_only\` emits a plain \`<p>\` instead of \`<figure>\`.

## Node.js \`JsOcrBackendBridge\` fixes

- Replaced \`unsafe { std::mem::transmute(js_obj) }\` with a \`napi_ref\` that keeps the JS object alive after \`registerOcrBackend\` returns.
- \`process_image\` uses \`ThreadsafeFunction\` — safely callable from tokio worker threads with real binary image bytes.
- Annotated with \`#[cfg_attr(alef, alef(skip))]\` so alef regeneration does not overwrite the hand-written bridge.

## Tests added

- \`run_ocr_on_images = false\` — placeholder not replaced when \`ocr_result\` is \`None\`
- \`run_ocr_on_images = true\` with native text — native text survives \`run_pipeline\`
- \`ocr_text_only = true\` on table-cell path — placeholder in a table cell replaced by OCR text
- Custom backend dispatch — mock backend in \`get_ocr_backend_registry()\`, verified used by \`process_images_with_ocr\`
- \`test_pdf_run_fallback_not_suppressed_without_images_config\` — gate defaults to \`false\` when \`config.images\` is \`None\`
- HTML rendering: \`image_no_flags\`, \`image_append_ocr_text\`, \`image_ocr_text_only\` (require \`--features html\`)


Goldziher (changes requested):
Codex agent review: requesting changes.

This is the right problem area, but the current patch has correctness regressions and no coverage for the new behavior.

1. The Node bridge expects `process_image` to return `Promise<String>` and then parses the string as JSON, but the README and natural JS API document `processImage` as returning an object. A user following the new docs will return `{ content, mime_type, ... }`, not a JSON string, so the bridge will fail or coerce to `[object Object]`. The binding should accept the documented object shape or the docs/types must match the real ABI.

2. `render_image` emits an extra quote in the figcaption: `<figcaption ...>{content}"</figcaption>`. That will render a stray quote after every OCR caption in styled HTML output.

3. The PDF fallback gate changes behavior whenever `images` config is present and `run_ocr_on_images` defaults to true. That means a caller who configures images but relies on sparse-page document-level OCR fallback now silently gets native text only. The new mixed-mode path needs to be opt-in or otherwise preserve the existing fallback semantics.

4. Please add tests covering the Node custom backend return shape, `runOcrOnImages` config mapping, HTML figcaption rendering, and the PDF fallback decision when image extraction config is present.

Goldziher:
Thanks for the iteration since the previous review. The direction is right — adding `run_ocr_on_images`, `ocr_text_only`, and `append_ocr_text`, and routing through the OCR backend registry are all useful fixes. The PDF guard (skip document-level fallback when image OCR is enabled) addresses the original critique about native text being replaced.

A few blockers before this can land. We'd like to merge — please push fixes for these and we'll re-review.

**1. Test coverage is too thin for the surface area.**

The Rust diff changes pipeline gating, PDF extractor behavior, image OCR backend resolution, and rendering inline-image substitution across multiple formats. Only one new test (`test_append_ocr_text_for_pptx_images`) was added. Please add coverage for:

- `run_ocr_on_images = false` — assert no OCR is performed on extracted images even when `config.ocr.is_some()`.
- `run_ocr_on_images = true` with PDF + native text present — assert native text is preserved and not replaced by document-level OCR.
- `ocr_text_only = true` — assert the `` placeholder is replaced by the OCR text in both paragraph and table-cell paths (you already cover paragraph; please add a table-cell test).
- Custom OCR backend lookup via `get_ocr_backend_registry` — assert that a registered custom backend is used by `process_images_with_ocr` (a Rust-side mock backend is fine).
- A regression test for the property name mapping (`processImage` → `process_image`, etc.) — even an integration test that calls the JS bridge with a JS object having `processImage` and verifies it routes correctly.

**2. The Node bindings (`crates/kreuzberg-node/index.d.ts`, `src/lib.rs`, `README.md`) are alef-generated.** Hand-editing them will be overwritten the next time `task alef:generate` runs (it already happened in this branch — `c9212e7ba3` on `chore/merge-head` regenerated index.d.ts against alef v0.19.21).

For the napi `ThreadsafeFunction` custom backend bridge to stick:
- Either land the change upstream in alef (kreuzberg-dev/alef) so the codegen produces this shape, or
- Mark the JsOcrBackendBridge module as a hand-customisation that alef skips (the `#[cfg_attr(alef, alef(skip))]` pattern, used elsewhere — search for it for examples).

As written, your changes to `index.d.ts`, `src/lib.rs`, and `README.md` will be lost on the next regen.

**3. The PR branch is your fork's `main`.** This means every rebase pulls in merge commits from upstream main into your branch, and the diff includes all that merge churn. Please rebase onto a fresh feature branch (e.g. `feat/run-ocr-on-images`) off the current `kreuzberg-dev/kreuzberg` main, so the diff is just your commits.

**4. Cargo.lock has a transitive shuffle.** Worth pinning down whether `tokio` and `base64` being added to `kreuzberg-node/Cargo.toml` is intentional or could be avoided by using the existing workspace deps.

If you'd prefer, we can split this into two PRs: (a) `run_ocr_on_images` + OCR backend registry routing (Rust-only, smaller scope), and (b) the Node ThreadsafeFunction bridge as an alef codegen change. Either way works.

Goldziher:
@kh3rld - please rebase this PR, and take it the full way to being merge ready

Goldziher (changes requested):
Code-quality review:

- Replacing `std::mem::transmute(js_obj)` with `napi_ref` + `ThreadsafeFunction` is the right call — that path was unsound for any tokio worker thread.
- `#[serde(default)]` on the three new fields keeps existing JSON/TOML payloads deserializing.
- `#[cfg_attr(alef, alef(skip))]` correctly fences the hand-written bridge from alef regen.
- Test coverage (`run_ocr_on_images=false`, `ocr_text_only` on table-cell path, fallback-not-suppressed-without-images-config, HTML render flags) looks proportional to the surface added.

Blocker — CI is red because the FRB-generated Dart bridge wasn't regenerated:

```
error[E0063]: missing fields `append_ocr_text`, `ocr_text_only` and `run_ocr_on_images`
 in initializer of `ImageExtractionConfig`
   --> packages/dart/rust/src/frb_generated.rs:8716:16
```

All 7 failures (Rust ubuntu-arm + macos, Mobile Android x64/arm64, iOS aarch64 + sim, Lint) cascade from that single exhaustive struct literal in `frb_generated.rs`.

Fix:

```bash
dart pub global activate flutter_rust_bridge_codegen
task alef:generate   # picks up the new fields
# or run the FRB codegen step directly per your local setup
```

Then re-push. Alternative minimal patch if you can't run FRB codegen locally: add `..Default::default()` to the one offending site so the struct literal is no longer exhaustive — but the proper fix is FRB regen so the generated bridge stays in sync.

Happy to merge once CI is green.

kh3rld (approved):
looks good!

Goldziher (approved):

Goldziher:
@kh3rld rebase and merge

PR #1020, opened 2026-05-20
fix: Replacing meminfo with cgroup check

Luishfs:
## Description
While using kreuzberg-server docker image in k8s, we noticed that the pod would constantly return OOMKilled status.
By tracing the logs to the code, i noticed that kreuzberg uses `/proc/meminfo` by default to stipulate memory availability, which in our case led to the memory issue.

### Traceback

kreuzberg-server returned this log before crashing:
```
2026-05-15T13:55:05.967730Z DEBUG request{method=POST uri=/extract version=HTTP/1.1}:api.extract{files_count=1}:kreuzberg.service{kreuzberg.operation="extract_bytes" kreuzberg.document.mime_type=application/pdf kreuzberg.document.size_bytes=257060}:extract_bytes{mime_type="application/pdf" kreuzberg.operation="extract_bytes" kreuzberg.document.mime_type="application/pdf" kreuzberg.document.size_bytes=257060}:kreuzberg.extract{kreuzberg.extractor.name="pdf-extractor" kreuzberg.document.mime_type="application/pdf" kreuzberg.document.size_bytes=257060}: kreuzberg::extractors::pdf::ocr: OCR batch size adaptation available_mb=6933 usable_mb=6420 document_mb=0 memory_limited_batch=42 configured=2 result=2
```
The kreuzberg pod was configured within a node running a c7i.2xlarge, with 16Gb availability, but its limit was set to 4. The logs show 6.4Gb of usable memory, which led to the OOM and the liveness probe failing.

### Fix

To fix this, i've included the cgroup (both 1 & 2 ) check under  the `get_available_memory()` function. 
The function will compare both cgroup and /proc/meminfo, and select the lesser one ( since cgroup will represent a slice of the host memory, cgroup will always win if available )

## Checklist

- [ ] CI passing
- [ ] Tests added where applicable


Goldziher (changes requested):
Codex agent review: requesting changes.

The direction is right, but this implementation is not ready to merge.

1. `cgroups-rs` is a large new dependency for reading two cgroup files, and it brings in a broad transitive stack including zbus/async-process. This code runs in OCR batch sizing and does not need DBus/cgroup management dependencies. Please replace it with a small direct reader for cgroup v2 (`memory.max`, `memory.current`) and cgroup v1 (`memory.limit_in_bytes`, `memory.usage_in_bytes`) based on `/proc/self/cgroup` + `/proc/self/mountinfo`, with explicit handling for `max`/unlimited limits.

2. `Cgroup::load(hierarchies::auto(), "")` appears to inspect the root/default cgroup rather than the current process cgroup. In Kubernetes the relevant limit is the pod/container cgroup for the running process, so this can still return host/root values and fail to fix the reported OOM case.

3. There are no tests for the parser/fallback behavior. Please add unit tests that cover cgroup v2 limited/unlimited values, cgroup v1 limited/unlimited values, malformed files, and the final `min(host_mem_available, cgroup_headroom)` selection.

Goldziher:
Thanks for the writeup — the OOMKilled traceback is exactly the right diagnostic, and `min(host_meminfo, cgroup_headroom)` is the correct composition. The fix is needed. Re-reviewed end-to-end; a few changes before merge.

**1. Drop `cgroups-rs`; implement direct file reads.**

`cgroups-rs` pulls in `zbus`, `async-process`, and a sizable transitive tree to read two files (`memory.{limit,usage}_in_bytes` on v1, `memory.{max,current}` on v2). The Cargo.lock diff is 550+ lines from this one dependency, and it gets compiled on every target — macOS, Windows, WASM, Android — even though the code is `#[cfg(target_os = "linux")]`. For a hot-path memory-sizing helper, the dep weight isn't justified.

A pure-stdlib implementation is ~30 lines:

```rust
fn cgroup_headroom() -> Option<usize> {
    // cgroup v2 (unified): /sys/fs/cgroup/memory.{max,current}
    if let (Ok(max), Ok(cur)) = (
        std::fs::read_to_string("/sys/fs/cgroup/memory.max"),
        std::fs::read_to_string("/sys/fs/cgroup/memory.current"),
    ) {
        let max = max.trim();
        if max == "max" { return None; }  // unlimited
        let limit = max.parse::<usize>().ok()?;
        let usage = cur.trim().parse::<usize>().ok()?;
        return Some(limit.saturating_sub(usage));
    }
    // cgroup v1 fallback
    let limit = std::fs::read_to_string("/sys/fs/cgroup/memory/memory.limit_in_bytes")
        .ok()?.trim().parse::<usize>().ok()?;
    let usage = std::fs::read_to_string("/sys/fs/cgroup/memory/memory.usage_in_bytes")
        .ok()?.trim().parse::<usize>().ok()?;
    // v1 limit_in_bytes returns ~9.2e18 when unlimited
    (limit < (isize::MAX as usize)).then(|| limit.saturating_sub(usage))
}
```

**2. If you do keep an external crate**, target-gate it so it only compiles on Linux:

```toml
[target.'cfg(target_os = "linux")'.dependencies]
cgroups-rs = "0.5.0"
```

The current unconditional entry adds Linux-only deps to every target's build graph.

**3. Add tests.**

The helper functions are pure file readers — easily testable. Please add unit tests that:

- Parse a sample `memory.max` containing the literal `max` (unlimited) and assert `cgroup_headroom()` returns `None`.
- Parse a numeric `memory.max` + `memory.current` and assert the saturating subtraction.
- Same for the cgroup v1 path (`memory.limit_in_bytes` with the "near-MAX means unlimited" sentinel).
- `read_meminfo_available()` with a synthetic `/proc/meminfo` string.

A small helper that takes the file contents as `&str` (instead of reading from disk) makes the unit tests trivial — the disk-reading wrapper stays thin and integration-tested.

**4. Logging.** Once the change lands, the existing `kreuzberg::extractors::pdf::ocr: OCR batch size adaptation` log line will naturally show the constrained value — no change needed there, but please verify in your test cluster that the `usable_mb` field now reflects the cgroup limit (e.g. 4096 instead of 6420 in your traceback).

The behaviour change is exactly right; we just need a lighter implementation and coverage before pulling it into the v5 line.

Luishfs:
@Goldziher Sorry for the wait, just addressed your points and removed the cgroups-rs lib requirement & added tests.
Also just merged my branch with all the incoming changes form the RC.3

PR #1050, opened 2026-05-23
feat(pptx): extract speaker notes and section names per slide

### LINKED ISSUE #960: feat(pptx): extract speaker notes and section names
## Summary

Today PPTX extraction returns the slide bodies and (where applicable) tables/images, but drops two fields that are first-class in the OOXML schema:

1. **Slide speaker notes** — the per-slide \`notesSlide\` content that authors write for verbal delivery. For accessibility, recap docs, talk-from-slides workflows, and meeting prep RAG, the notes are often more information-dense than the slide body.
2. **Section names** — PowerPoint's section grouping (the \`p:sectionLst\` element in the presentation part). Users use sections as logical chapters; surfacing them gives chunking and summarization a meaningful boundary signal.

## Scope

- Parse \`ppt/notesSlides/notesSlide{N}.xml\` and attach the notes text to each \`PageContent\` (a slide).
- Parse \`ppt/presentation.xml\` \`p:sectionLst\` and attach the containing section name to each slide.
- Surface in \`ExtractionResult\`:
  - \`pages[i].speaker_notes: Option<String>\`
  - \`pages[i].section_name: Option<String>\`
- Include in the markdown output as a recognisable block (e.g. \`> **Speaker notes:** …\`) so it round-trips through markdown consumers without being silently dropped.

## Why this matters

- Talk-from-slides and AV-assistant use cases need the notes.
- Section names are a free chunking signal that doesn't require re-running heading detection over slide bodies.
- Layout-aware chunking benefits from knowing which slides belong to which section.

## Open questions

- Markdown rendering convention for notes — fenced block, blockquote prefix, or separate output field. Lean toward a separate field plus an opt-in toggle to inline.
- Slide layout name (\`Title Slide\`, \`Section Header\`, etc.) — useful adjacent feature; in scope or follow-up?

kh3rld:
implements `speaker_notes` and `section_name` to `PageContent` for PPTX extraction.

Both fields are `Option<String>` with `skip_serializing_if = "Option::is_none"`. They are only populated when `pages` config is set and the source is a PPTX file; all other extractors receive `None`.

Also fixes a data-loss bug in `build_document_from_result` where `page_contents` was not transferred to `doc.prebuilt_pages`, causing both fields to be dropped through the full extractor path even when correctly populated by `extract_pptx_from_bytes`.

The legacy `.ppt` (OLE binary) path uses `InternalDocumentBuilder` and does not produce per-slide `PageContent` objects; speaker notes for `.ppt` remain in `metadata.additional["speaker_notes"]` as a JSON array (documented by `test_ppt_speaker_notes_in_metadata_not_page_contents`).

Closes #960

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- **pptx**: `PageContent` gains two optional fields populated during PPTX per-page extraction (requires `page_config` to be set): `speaker_notes` (text from `ppt/notesSlides/notesSlideN.xml`) and `section_name` (from `<p14:sectionLst>` in `ppt/presentation.xml`). Both serialize with `skip_serializing_if = "Option::is_none"` and are `None` for all non-PPTX formats. (`crates/kreuzberg/src/types/page.rs`, `crates/kreuzberg/src/extraction/pptx/`)
+
 ### Changed
 
 - **deps**: bump alef pin to v0.19.5 (bundles AHashMap binding fixes for FFI/Dart/Swift/Elixir/Ruby/PHP and WASM emitter dedup + collect + From-impl + sub-config deserialize fixes).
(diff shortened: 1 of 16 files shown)


Goldziher:
Integrated into `chore/merge-head` via cherry-pick (will be merged to `main` via the staging PR). Thanks for the contribution!

Commits on staging branch:
41d66d0fa5 chore: regenerate bindings with alef v0.19.21 (PageRaster + speaker_notes/section_name)
3c132df21b feat(pptx): extract speaker notes and section names per slide

PR #1052, opened 2026-05-24
chore: add .gitattributes marking generated dirs as linguist-generated

kh3rld:
Adds a root-level `.gitattributes` emitted by `alef scaffold` (alef #110). Marks all generated output directories (`packages/**`, `crates/*-{py,php,ffi,node,wasm}/**`, `e2e/**`) as `linguist-generated=true` so generated files collapse in GitHub PR diffs.

This is a create-once seed file — it will not be overwritten on re-scaffold unless `--clean` is passed.


Goldziher:
Heads up: `.gitattributes` was already added to main via the alef scaffold regen in commit `e5ea539ef4` (shipped in v5.0.0-rc.3), with broader coverage than this branch. The only remaining net diff here is a CHANGELOG line that targets the rc.1 section (already-released history).

If you'd like to keep this PR open, please move the CHANGELOG entry to `[Unreleased]` or rebase. Otherwise this can be closed. Thanks!

PR #1054, opened 2026-05-26
feat(chunking): populate image_indices for YAML/JSON section chunks

### LINKED ISSUE #963: feat(chunking): YamlSectionChunker should populate image_indices
## Problem

`YamlSectionChunker` does not track page provenance: `ChunkMetadata.first_page` and `last_page` are always `None` for YAML-section chunks. The post-chunking step in `features.rs` skips `image_indices` population for any chunk where either page bound is `None`, so YAML-section chunks always have empty `image_indices` regardless of whether images were extracted.

Introduced in #777 — acknowledged as a known gap.

## Desired behavior

YAML-section chunks should carry correct `image_indices` matching the images on their constituent pages.

## Approach

Teach `YamlSectionChunker` to derive page bounds from byte offsets against the extraction result's `page_boundaries`, then let the existing `features.rs` image-index population run normally.

kh3rld:
YAML/JSON section chunks now carry correct `first_page`/`last_page` values, enabling the existing `image_indices` population step in `features.rs` to run.

Root cause: `chunk_yaml_by_sections` never received `page_boundaries`, so all YAML section chunks had `first_page = last_page = None` and the image-index loop was unconditionally skipped for them.

Fix: thread `page_boundaries` into `chunk_yaml_by_sections`. Since serde_yaml_ng doesn't expose source positions, all section byte offsets are 0; `page_for_offset(0, boundaries)` returns the first page — correct for single-page YAML/JSON files (the common case) and conservative for the rare multi-page edge case.

Closes #963

### DIFF
--- a/crates/kreuzberg/src/chunking/core.rs
+++ b/crates/kreuzberg/src/chunking/core.rs
@@ -82,7 +82,7 @@ pub(crate) fn chunk_text_with_heading_source(
 
     // Yaml creates new content per chunk (key prefix), can't use generic &str splitter.
     if config.chunker_type == ChunkerType::Yaml {
-        return super::yaml_section::chunk_yaml_by_sections(text, config);
+        return super::yaml_section::chunk_yaml_by_sections(text, config, page_boundaries);
     }
 
     // Semantic chunker has its own pipeline (segment → topic detect → merge).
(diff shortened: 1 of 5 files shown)


Goldziher:
Integrated into `chore/merge-head` via cherry-pick (will be merged to `main` via the staging PR). Thanks for the contribution!

Commits on staging branch:
aa7800e29a feat(chunking): populate image_indices for YAML/JSON section chunks

PR #1062, opened 2026-05-28
fix: added per-page PDF ocr

Luishfs:
## Description
While running kreuzberg server on a file processing pipeline, we noticed that kreuzberg memory footprint had a abnormal increase when allowing for PDF files to be OCRed in case they contained images and/or pages that contained text, but were not extractable with the default tools.

The behavior that was describe by both AI agents ( mainly claude code ) and that were also found inside [kreuzberg comments](https://github.com/kreuzberg-dev/kreuzberg/blob/707cdc2dedcbe9cd9e496d8f8474666ca82e19ed/crates/kreuzberg/src/extractors/pdf/ocr.rs#L52), is that kreuzberg would look into each page and decide whether to OCR that page or use the default text extraction.

Checking the code, i noticed that the evaluation [function](https://github.com/kreuzberg-dev/kreuzberg/blob/707cdc2dedcbe9cd9e496d8f8474666ca82e19ed/crates/kreuzberg/src/extractors/pdf/ocr.rs#L291) below:

```rust
#[cfg(feature = "ocr")]
pub(crate) fn evaluate_per_page_ocr(
    native_text: &str,
    boundaries: Option<&[crate::types::PageBoundary]>,
    page_count: Option<u32>,
    thresholds: &OcrQualityThresholds,
) -> OcrFallbackDecision {
    let boundaries = match boundaries {
        Some(b) if !b.is_empty() => b,
        _ => return evaluate_native_text_for_ocr(native_text, page_count, thresholds),
    };

    let mut document_decision = evaluate_native_text_for_ocr(native_text, page_count, thresholds);

    for boundary in boundaries {
        if boundary.byte_end > native_text.len() || boundary.byte_start > boundary.byte_end {
            continue;
        }
        let page_text = &native_text[boundary.byte_start..boundary.byte_end];
        if evaluate_native_text_for_ocr(page_text, Some(1), thresholds).fallback {
            document_decision.fallback = true;
            return document_decision;
        }
    }

    document_decision
}
```
Actually returns document_decision whenever a single page fails, making so that the whole PDF gets scanned because of one faulty page.

This PR adds the per-page intended behavior for kreuzberg pdf OCR and should prevent this issue.
## Checklist

- [ ] CI passing
- [X] Tests added where applicable


Goldziher:
@kh3rld please review this PR

crates/kreuzberg/src/extractors/pdf/mod.rs
```diff
@@ -332,6 +332,32 @@ impl PdfExtractor {
                         (native_text, ExtractionMethod::Native)
                     }
                 },
+                ocr::OcrGateOutcome::RunFallbackOnPages(pages) => match boundaries.as_deref() {
+                    Some(bounds) if !bounds.is_empty() => {
+                        match ocr::extract_mixed_ocr_native(&native_text, bounds, &pages, content, config, path).await {
```
kh3rld, line 370: Really clean fix — not short-circuiting on the first bad page is exactly the right call. One thing worth thinking about: the `force_ocr_pages` path is user-directed, so callers knowingly accept any OCR vs native text tradeoff. Here, the pages list comes from the automatic quality detector. Do we want a confidence floor before committing to replacing native text with OCR output? A false positive from the detector would silently downgrade perfectly good text. Not a blocker at all — just worth capturing as a follow-up issue once we have production data on how often the detector misfires.

crates/kreuzberg/src/extractors/pdf/ocr.rs
```diff
@@ -38,6 +38,8 @@ pub struct OcrFallbackDecision {
     pub avg_non_whitespace: f64,
     pub avg_alnum: f64,
     pub fallback: bool,
+    pub failing_pages: Vec<u32>,
+    pub whole_doc_failure: bool,
```
kh3rld, line 45: Love the `whole_doc_failure` field — it's the right way to distinguish document-level degradation from per-page spot failures. Could we make the invariant explicit with a quick doc comment? Something like: *set to `true` when the aggregate document quality check triggered `fallback`, independently of any per-page analysis*. Future contributors hitting this for the first time will know immediately why it exists alongside `failing_pages` rather than wondering if the two can get out of sync.

crates/kreuzberg/src/extractors/pdf/ocr.rs
```diff
@@ -79,10 +83,14 @@ pub(crate) fn evaluate_ocr_skip_gate(
 
     if skip_for_non_text {
         OcrGateOutcome::SkipNonText
-    } else if has_substantive_doc && !decision_fallback {
+    } else if has_substantive_doc && !decision.fallback {
         OcrGateOutcome::SkipSubstantive
-    } else if decision_fallback {
-        OcrGateOutcome::RunFallback
+    } else if decision.fallback {
+        if decision.whole_doc_failure || decision.failing_pages.is_empty() {
```
kh3rld, line 96: Smart guard — the `decision.whole_doc_failure || decision.failing_pages.is_empty()` condition correctly handles both cases. Worth a tiny comment here explaining why the second clause is needed: when `evaluate_native_text_for_ocr` fires (no page boundaries available), `fallback` can be `true` but `failing_pages` is empty because there were no boundaries to enumerate. Without the comment a reader might wonder if the second clause is dead code.

crates/kreuzberg/src/extractors/pdf/ocr.rs
```diff
@@ -300,18 +312,31 @@ pub(crate) fn evaluate_per_page_ocr(
     };
 
     let mut document_decision = evaluate_native_text_for_ocr(native_text, page_count, thresholds);
+    document_decision.whole_doc_failure = document_decision.fallback;
```
kh3rld: The per-page scan still runs even when `whole_doc_failure` is already true after `evaluate_native_text_for_ocr`. That's fine for collecting the `failing_pages` list for logging — just want to make sure it's intentional rather than accidentally paying the per-page scan cost on fully degraded documents. A brief comment like *// still scan to populate failing_pages for telemetry* would make this crystal clear for anyone profiling later.

crates/kreuzberg/src/extractors/pdf/mod.rs
```diff
@@ -1576,4 +1614,135 @@ mod tests {
...
+    #[test]
+    fn test_ocr_gate_whole_doc_failure_dominates_per_page_list() {
+        let thresholds = OcrQualityThresholds::default();
+        let outcome =
+            ocr::evaluate_ocr_skip_gate(false, 500, 0.9, &mk_decision(true, true, vec![1, 2, 3]), &thresholds);
+        assert_eq!(
+            outcome,
+            ocr::OcrGateOutcome::RunFallback,
+            "whole-doc failure must trigger full OCR even when per-page list is populated"
+        );
```
kh3rld, line 1805: Great test — the four-page layout with alternating good/bad text makes the assertion easy to follow. One small nit: the `good` string is 72 chars and `bad` is 5 chars. Given thresholds can change, would it be worth adding a brief comment explaining why these specific strings reliably pass/fail the quality check? Something like *// 'bad' is sparse punctuation: alnum_ratio ≈ 0, triggers fallback deterministically* would make this test self-documenting and help whoever tunes thresholds in future not accidentally break the fixture.

Goldziher:
@kh3rld please take over this PR. get it merge readyy and let me know.

kh3rld:
nice!

kh3rld (commented):
Addressed all review comments in d5b55125 (`docs(pdf): address review comments on per-page OCR fallback`):

- `whole_doc_failure` now has an explicit doc comment documenting its invariant: set when the aggregate document quality check triggered fallback, independently of per-page analysis.
- Added a comment to the `failing_pages.is_empty()` guard in the OCR gate explaining why the clause is needed: when `evaluate_native_text_for_ocr` fires with no boundaries, `fallback` can be true but `failing_pages` is empty — there were no boundaries to enumerate.
- Added a comment before the document-level check in `evaluate_per_page_ocr` clarifying that the per-page scan still runs on whole-doc failures intentionally, to populate `failing_pages` for telemetry.
- Added fixture comments in the test explaining why `good`/`bad` strings deterministically pass/fail (alnum_ratio ≈ 0 on `bad`, well above threshold on `good`).

The comment on `mod.rs:337` about the `force_ocr_pages` confidence floor is a valid follow-up point; captured separately as a future issue.

PR #1070, opened 2026-05-30
feat(transcription): impl audio/video speech-to-text foundation

kh3rld:
Introduces `transcription` and `transcription-types` feature gates for Whisper ONNX-based speech-to-text extraction. Follows the exact pattern of `embeddings`, `layout-detection`, and `paddle-ocr`.

## What's in this PR

- `TranscriptionConfig` + `WhisperModel` enum — pure-Rust config structs under `transcription-types`, WASM/Android-safe
- MIME type registrations for mp3, mp4, m4a, wav, ogg, webm, flac, opus, aac, avi, mov, mkv and family aliases
- `TranscriptionExtractor` plugin adapter wired into the default registry at priority 50
- `decode_audio_to_pcm` scaffold (Symphonia-backed) — resample to 16kHz mono f32 for Whisper input
- All heavy deps (ort, hf-hub, sha2, symphonia) are optional; zero impact when `transcription` is disabled
- deny.toml MPL-2.0 exceptions for all symphonia sub-crates
- `transcription` added to `server` aggregate; excluded from `no-ort-target`, `wasm-target`, `android-target`

## What's next (follow-up PR)

- Model download via `hf-hub` + SHA-256 verification
- Whisper ONNX inference (mel-spectrogram → encoder → decoder → text)
- VAD pre-pass to skip silence, batched segment processing
- Segment timestamps in result metadata

## Test plan

- [ ] `cargo check --features transcription` passes
- [ ] `cargo check` (no feature) has zero size impact
- [ ] MIME detection tests cover new audio/video extensions
- [ ] CI green on Linux amd64/arm64 and macOS


Goldziher:
great. please test it locally -- make sure its working end to end.

Also, i am testing on a branch using `candle` or `mistral.rs` for some models. It might be its a better option than using ort for these. not sure though.

kh3rld:
> great. please test it locally -- make sure its working end to end.
> 
> Also, i am testing on a branch using `candle` or `mistral.rs` for some models. It might be its a better option than using ort for these. not sure though.

tested this; 
on this implementation, the scaffold is intentionally backend agnostic;the inference stub doesn't commit to ORT yet

Goldziher (changes requested):
1. Critical: extractor returns Ok(empty doc) instead of Err

In crates/kreuzberg/src/extractors/transcription.rs both the async path (~L95-101)
 and sync path (~L158-164) do:

let _ = pcm;
tracing::warn!("transcription stub: returning empty document — Whisper inference
not yet wired (follow-up PR)");
let mut doc = InternalDocument::new("audio-transcript");
doc.mime_type = mime_type.to_string();
Ok(doc)

A user who enables the transcription feature, sends an mp3, and is not watching
warn-level logs gets Ok(empty_doc) with no transcript and no signal that anything
is missing. Production observability rarely catches this. Please change both paths
 to return Err(KreuzbergError::transcription("Whisper inference not yet
implemented — pending follow-up PR")) and flip
test_async_stub_succeeds_with_valid_config to expect an error. That test is
currently codifying the silent-success behavior.

2. Metadata extraction is entirely missing

The PR title is "audio/video speech-to-text foundation" but the foundation only
covers the decode-and-error-handling half. The metadata extraction half — which is
 squarely kreuzberg's purview, independent of any STT inference — is absent:

- FormatMetadata (types/metadata.rs:60-92) has discriminants for every existing
format (Pdf, Docx, Excel, Email, Image, Html, Code, …) but no Audio variant. The
transcription extractor returns a doc with default metadata.
- None of the common Metadata fields (title, authors, created_at, language,
subject) are populated from tags. For an mp3 with ID3, title should be the song
title, authors should be the artist, created_at should come from the
year/release-date tag, language from the language tag. For an m4a, the same fields
 are in MP4 atoms.
- Duration, codec, sample rate, channels, bitrate, container — all data we already
 know after decode — are dropped.

Even before Whisper inference lands, the foundation PR should:

- Add a FormatMetadata::Audio(AudioMetadata) variant carrying duration_ms, codec,
container, sample_rate_hz, channels, bitrate, plus any tag-derived fields.
- Populate the common Metadata fields from tags so audio files get the same
baseline document model every other format gets.

3. Please evaluate lofty-rs in place of symphonia for the tag-read path

lofty-rs is MIT, pure Rust, actively maintained, and covers ID3v1/v2, MP4 atoms
(iTunes-style), Vorbis comments, Opus tags, FLAC tags, WavPack, APE, AIFF, RIFF
INFO — strictly broader tag coverage than symphonia's metadata API, and read-write
 where symphonia is read-only. It is purpose-built for exactly this use case.

Concretely:
- For metadata extraction: lofty is the right tool. It would let you populate the
Metadata fields and a new FormatMetadata::Audio variant from one focused crate,
independent of whatever decoder is chosen for the inference pipeline.
- For PCM decode: symphonia stays. We've reviewed the MPL-2.0 question separately
— it is file-scoped and operationally a permissive license for our consumption
pattern (we are not modifying symphonia source), so MPL is not a reason to migrate
 the decode path. The decision about decode (keep symphonia / shell to ffmpeg /
WAV-only) can be made later and shouldn't block this PR.

Could you spike lofty for the tag path and report back whether it covers the
formats we register MIME types for? If yes, the foundation PR should land with
lofty wired up so the metadata half of the feature is real on day one — even while
 the inference half waits for the follow-up.

Notes that are not blockers but worth tracking

- The MIME registrations in core/mime.rs are unconditional and the in-code comment
 says "extraction fails gracefully with a clear error when the feature is absent".
 In practice a build without transcription returns generic
UnsupportedFormat("audio/mpeg") — accurate but not actionable. Consider routing to
 a clearer message hinting at the missing feature flag. Not blocking; same
behavior as office/pdf today.
- test_apps-style metadata tests for audio: please add a fixture covering at least
 one mp3 with ID3 tags and one m4a with MP4 atoms, asserting that
title/artist/duration flow through to ExtractionResult. Without these, the
metadata story is untested.

The deny.toml work and feature-isolation work in this PR are solid. Once (1)
honest error, (2) FormatMetadata::Audio + common-field tag population, and (3) the
 lofty evaluation are addressed, this is mergeable as the real foundation.

kh3rld:
All three blocking items addressed in 2b2dc659:

**1. Honest error instead of silent empty doc**
Both async and sync paths now return `Err(KreuzbergError::transcription("Whisper inference not yet implemented - pending follow-up PR"))`. `test_async_stub_succeeds_with_valid_config` is renamed and flipped to assert the error.

**2. `FormatMetadata::Audio(AudioMetadata)`**
New `AudioMetadata` struct under `transcription-types` (WASM/Android-safe, no ORT): `duration_ms`, `codec`, `container`, `sample_rate_hz`, `channels`, `bitrate`. Wired into `Display`, `PartialSchema` discriminator, and `ToSchema`.

**3. lofty for the tag path**
`lofty = "0.24"` added as an optional dep under the `transcription` feature (independent of the inference backend choice - candle/ORT/etc. won't affect tag reading). New `transcription/tags.rs` with `read_audio_tags()` covering ID3v1/v2 (MP3), MP4 atoms (M4A/MP4), Vorbis comments (OGG/FLAC/Opus), and RIFF INFO (WAV) — all formats we register MIME types for. Failures are non-fatal (debug-logged, returns `AudioTags::default()`).

Both extractor paths now call `read_audio_tags()` alongside PCM decode. Tags + pcm are in scope and ready to populate `Ok(doc)` with full `Metadata` + `FormatMetadata::Audio` when inference lands in the follow-up PR.

One note: the `AudioTags` fields have `#[allow(dead_code)]` for now since they are fully populated but not yet consumed (the `Ok(doc)` path doesn't exist yet). That annotation comes off in the inference PR.

Issue #1074, opened 2026-06-03
bug(chunking): first_page/last_page is None for chunks when output_format != Plain


kh3rld:
## Summary

When `output_format` is Markdown, Djot, HTML, or a custom format, `execute_chunking` now correctly chunks `formatted_content` instead of plain text (fixes #1073). As a consequence, page boundaries — which are computed against plain-text byte offsets — are not passed to the chunker for non-plain paths. `chunk.metadata.first_page` and `chunk.metadata.last_page` are `None` for all non-plain output formats.

## Expected

`chunks[].metadata.first_page` and `chunks[].metadata.last_page` are populated even when `output_format != Plain`.

## Current

Both fields are `None` for non-plain format chunks. The plain-text path is unaffected (page metadata continues to work correctly when `output_format == Plain`).

## Root Cause

Page boundaries (`PageBoundary { byte_start, byte_end }`) are computed by `recompute_boundaries_from_pages` against `result.content` (plain text). After the #1073 fix, chunks are split from `formatted_content` instead. The byte offsets from the plain-text derivation are not valid indices into the formatted string, so they are intentionally omitted.

## Fix Direction

Re-derive page boundaries against `formatted_content` using the same `recompute_boundaries_from_pages` approach but operating on the formatted string and the formatted page content. This requires either:

1. Re-rendering per-page content in the requested output format before boundary computation, or
2. Building a character-offset mapping from plain-text positions to formatted-string positions during rendering.

Option 1 is simpler but requires per-page rendering. Option 2 is more efficient but requires renderer cooperation.

## Affected Config

Any combination of:
- `output_format: markdown | djot | html | custom`
- `chunking` enabled

## Workaround

Use `output_format: plain` if chunk page metadata is required. The top-level `content` field can still be set to markdown post-hoc by callers if needed.

PR #1075, opened 2026-06-03
fix(pipeline): chunks[].content respects output_format

kh3rld:
When output_format=markdown (or html/djot/custom), execute_chunking always chunked result.content (plain text) and passed formatted_content only as a heading-context hint. Chunks were frozen as plain text before apply_output_format swapped formatted_content into the top-level content field, so chunks[].content and content disagreed regardless of output_format.

Fix: when output_format != Plain and formatted_content is available, chunk the formatted string directly. The Plain path is unchanged. Page metadata  (first_page/last_page) is not populated for non-plain chunks; byte offsets from the plain-text derivation are invalid for the formatted string; tracked in #1074.

  The second commit repairs five pre-existing hook failures unblocked by this branch: jni 0.22→0.21 version pin (comment/code mismatch), chunk_size test feature gates, ExtractedImage struct literal, cargo-machete optional-dep false positives, MIT-0 in deny.toml allow-list.
  
  resolves #1037


crates/kreuzberg-jni/Cargo.toml
```diff
@@ -17,7 +17,7 @@ name = "kreuzberg_jni"
 base64 = "0.22"
 # jni 0.22 split JNIEnv into Env + EnvUnowned; migration to with_env(|env| ...) pattern is deferred.
 # Pin to 0.21 until kreuzberg-jni is rewritten against the new API.
-jni = "0.22"
+jni = "0.21"
```
Goldziher, line 20: why?

PR #1087, opened 2026-06-09
feat(captioning): wire image bytes into OCR path for standalone files

### LINKED ISSUE #732: feat: VLM Images Captions
As the LLM/VLM integration have been added in #662 (thanks btw :) . There is great use of it in images/charts captioning. It is desired to be able to enable the VLM captioning independently of the backend engine, i.e. even if the selected engine is Paddle , the VLM image captioning pipeline would work with the images in that file.

A nice-to-have feature is to separate the configuration of the VLM image captioning options from the VLM backend pipeline options, as some VLM models are finetuned for OCR use-case (Dots.OCR/GLM OCR) while others are general VLM models with great captioning/describing capabilities such as the new Gemma4 series.

kh3rld:
`CaptioningProcessor` was a silent no-op for standalone image files processed via OCR. `ImageExtractor` built the raw image bytes but passed `None` as the images argument to `build_image_internal_document`, so `InternalDocument.images` was always empty on the OCR path,.. the processor had nothing to work with.

  This injects `extracted_image` into doc.images when config.captioning is set, completing the wiring in both the layout-OCR and regular-OCR branches. Also adds a `ProcessingWarning` when captioning is configured but `result.images` is `None`, so document-file users who forgot `config.images` get an actionable signal instead of silence.

  closes #732

### DIFF
--- a/crates/kreuzberg/src/extractors/image.rs
+++ b/crates/kreuzberg/src/extractors/image.rs
@@ -492,6 +492,9 @@ impl DocumentExtractor for ImageExtractor {
                     Ok(mut doc) => {
                         doc.metadata.format = Some(crate::types::FormatMetadata::Image(image_metadata));
                         doc.mime_type = mime_type.to_string();
+                        if config.captioning.is_some() {
+                            doc.images.push(extracted_image.clone());
+                        }
                         return Ok(doc);
                     }
                     Err(e) => {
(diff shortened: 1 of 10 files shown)


crates/kreuzberg/src/extractors/image.rs
```diff
@@ -814,4 +820,180 @@ mod tests {
...
+            doc.images.len(),
+            1,
+            "doc.images must contain the raw image (regression of #732)"
+        );
+        assert!(!doc.images[0].data.is_empty(), "image data must be non-empty");
+    }
+
+    /// Full-pipeline regression for #732: InternalDocument.images must survive the
+    /// derive.rs conversion so ExtractionResult.images is Some after run_pipeline.
+    #[cfg(feature = "ocr")]
```
Goldziher, line 920: one important note about scope - this should not be OCR only. captioning should be an option for all image paths. For example, we extract an image from a docx document, user wants captionoing.

Goldziher (commented):
see the comment i left. captionoing is not ocr specific.

Issue #1091, opened 2026-06-10
feat: expose the supported-format enumeration in the language bindings


tobocop2:
## Problem

There's no way to ask kreuzberg what formats it supports. The bindings can validate or detect a *given* MIME type, but they can't enumerate the supported set.

So any downstream tool that decides which files to ingest keeps its own hand-maintained extension whitelist, and that whitelist silently drifts every time kreuzberg adds a format. I got bitten by exactly this: gif/jp2/doc/eml were quietly dropped from my ingestion because my map predated their support.

## Proposed solution

Expose the existing `list_supported_formats()` — it already backs `kreuzberg formats` in the CLI, `GET /formats` in the REST API, and the MCP server — through the language bindings:

```python
from kreuzberg import list_supported_formats

supported = {f.extension: f.mime_type for f in list_supported_formats()}
```

Ingestion policy then becomes a small denylist over what kreuzberg actually supports, instead of a whitelist that has to chase releases.

## Open questions

- Extension/MIME pairs cover my use case. Is the full supported-MIME set (aliases, extension-less types) worth exposing too?

Happy to implement this and assign myself.

PR #1093, opened 2026-06-10
feat: expose list_supported_formats() in all language bindings

tobocop2:
Refs #1091.

## Problem

Bindings can validate a given MIME type or map one to extensions, but nothing enumerates what kreuzberg supports, so downstream tools keep hand-maintained extension whitelists that drift whenever kreuzberg adds a format. The enumeration already exists in core — `list_supported_formats()` backs the CLI, REST API, and MCP server — it just isn't exported to the bindings.

## Solution

Remove the `alef(skip)` attribute from `list_supported_formats()`, re-export it from the crate root, and regenerate with the pinned alef 0.23.75. All 15 bindings now expose the function, each with a generated e2e test asserting a non-empty result (new `plugin_api` fixture plus an e2e call block in `alef.toml`).

Notes for review:

- The e2e call block uses the `result_is_simple + result_is_array` shape; `result_is_vec` makes the Go emitter generate an error-pattern test that fails on success.
- The rustdoc was tightened while un-skipping: the list is what kreuzberg recognizes by extension (extension-less, dynamically detected formats aren't included), and it no longer references the private `FORMATS` static since alef copies doc text into binding docs.
- `types::document_structure::tests::test_skip_serializing_empty_fields` fails on main under `--features full` too; not introduced here.

Verified locally with cargo fmt, workspace clippy (`-D warnings`), the full `--features full` test suite, and `alef verify --exit-code`. Cross-language runtime suites run in the CI e2e matrix.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- **`list_supported_formats()` is now part of the public crate root and
+  every language binding.** Returns every file extension Kreuzberg
+  recognizes with its corresponding MIME type, so callers can derive
+  ingestion policy from the library instead of maintaining their own
+  extension whitelists. The function already backed the CLI
+  (`kreuzberg formats`), REST API (`GET /formats`), and MCP server; it
+  is now exported from the crate root and exposed in every binding via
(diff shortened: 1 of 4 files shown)


Goldziher:
@tobocop2 please rebase this.

tobocop2:
Rebased onto current main and regenerated with the new pinned alef 0.24.1.

PR #1106, opened 2026-06-12
fix(pdf): recalculate is_blank after OCR page texts are written

### LINKED ISSUE #1095: bug: result.pages is empty for scanned PDFs even when content and chunks exist
### Description
`result.pages[*].content` is empty for scanned / rasterized PDFs even when OCR succeeds

## Summary

For a fully scanned PDF (rasterized pages, no embedded text layer), Kreuzberg returns non-empty OCR output at the document and chunk levels, but `result.pages[*].content` is empty for every page.

In the same extraction result:

- `result.content` is populated
- `result.chunks[*].content` is populated
- but `result.pages[*].content` is empty for all pages
- and all pages are marked `isBlank: true`

This looks incorrect because OCR clearly succeeded at the document level.

## Observed behaviour

From one extracted 32-page scanned PDF:

- `len(result.content) = 48134`
- `len(result.chunks) = 57`
- `len(result.pages) = 32`
- every `result.pages[i].content == ""`
- every `result.pages[i].isBlank == true`

Example page entries:

```json
{
  "pageNumber": 1,
  "content": "",
  "isBlank": true
}
```

At the same time, the top-level content contains OCR text such as:

```text
# Investment and Subscription Agreement

dated 13.12.2021

entered into by and between
...
```

and chunks are also present and non-empty.

## What did you expect to happen?

If OCR succeeds for a scanned PDF, then page-level extraction should also contain page text, or at minimum pages should not all be marked blank.

Expected:

1. `result.pages[*].content` contains OCR text for each scanned page, or
2. if page extraction is intentionally unsupported for this case, docs should say so explicitly.

Current behaviour is inconsistent because OCR output exists, but the page layer is empty and flagged blank.

## Why this matters

Downstream systems often rely on `result.pages[*].content` for page-level classification, routing, preview, and auditability. When scanned PDFs produce only top-level OCR text but no page-level text, it becomes difficult to map extracted text back to specific pages.

## Environment

- Kreuzberg Docker image: `ghcr.io/kreuzberg-dev/kreuzberg:4.9.9-core`
- Input type: scanned / rasterized PDF (no embedded text layer)
- Config includes:
  - `output_format: markdown`
  - `include_document_structure: true`
  - `ocr.output_format: markdown`
  - `pages.extract_pages: true`

## Config snippet

```yaml
output_format: markdown
include_document_structure: true

ocr:
  output_format: markdown
  backend: vlm

pages:
  extract_pages: true
```

## Steps to reproduce

1. Start Kreuzberg with OCR enabled and `pages.extract_pages: true`.
2. Extract a fully scanned / rasterized PDF.
3. Inspect:
   - `result.content`
   - `result.chunks[*].content`
   - `result.pages[*].content`
   - `result.pages[*].isBlank`
4. Observe that OCR text exists in `result.content` and `result.chunks`, but every page content is empty and every page is marked blank.

## Additional signals from the payload

The payload also showed page metadata boundaries that looked degenerate for all pages, e.g. page 1 had:

```json
{
  "byteStart": 0,
  "byteEnd": 0,
  "pageNumber": 1
}
```

and subsequent pages advanced only by tiny increments, which may be related to the empty page content.

If helpful, I can provide the sample PDF and a sanitized raw extraction payload.

kh3rld:
For scanned/rasterized PDFs with no native text layer, native extraction sets is_blank=Some(true) on every page. When OCR subsequently writes content into those PageContent entries, is_blank was never recalculated,.. leaving  result.pages[*].isBlank=true even when page.content was fully populated.

  Four write sites in the OCR page-assembly block now call is_page_text_blank after every content mutation (zip loop, VLM single-string guard, scratch-page  construction, results_map targeted-OCR path).

  fixes #1095.

  Note: the degenerate byteStart/byteEnd==0 values reported in the same issue are a separate concern and not addressed here.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -304,7 +304,6 @@ packages/r/src/*.o
 
 # BEGIN ai-rulez (DO NOT EDIT - managed by ai-rulez)
 .agents/
-.ai-rulez/.generated-manifest.json
 .cursor/
 .github/agents/
 .github/commands/
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -73,6 +73,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- **pdf: `result.pages[*].isBlank` now reflects OCR content for scanned/rasterized PDFs.**
+  When OCR (including VLM) wrote text into existing `PageContent` entries, `is_blank` was
+  never recalculated — it retained the stale value from native text extraction, which is
+  always `Some(true)` for pages with no text layer. All four write sites in the OCR
+  page-assembly block now call `is_page_text_blank` after every content mutation.
+  ([#1095](https://github.com/kreuzberg-dev/kreuzberg/issues/1095))
+
 - **reranker: `RerankError` migrated to `thiserror`.** Matches the rest
   of the library and `rust-conventions`.
 
--- a/crates/kreuzberg/src/extractors/pdf/mod.rs
+++ b/crates/kreuzberg/src/extractors/pdf/mod.rs
@@ -425,6 +425,7 @@ impl PdfExtractor {
 
                     for (page, text) in pages.iter_mut().zip(pts) {
                         page.content = crate::pdf::text::fix_pdf_control_chars(&text).into_owned();
+                        page.is_blank = Some(crate::extraction::blank_detection::is_page_text_blank(&page.content));
                     }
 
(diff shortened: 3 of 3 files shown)


Goldziher:
@kh3rld -- please do a followup handling: "Note: the degenerate byteStart/byteEnd==0 values reported in the same issue are a separate concern and not addressed here."

Ok?

Goldziher (approved):

PR #1109, opened 2026-06-13
fix(ocr): prevent server abort when preprocessing config is set

### LINKED ISSUE #1089: bug: Enabling tesseract_config.preprocessing aborts /extract with ECONNRESET
## Summary
Enabling `ocr.tesseract_config.preprocessing` causes the Kreuzberg API process to abort during `/extract`, which resets the client connection.

## Environment
- Container image: `ghcr.io/kreuzberg-dev/kreuzberg:4.9.9-core`
- API mode: `serve -H 0.0.0.0 -p 8000 --config /app/kreuzberg.yml --log-level debug`
- Request: `POST /extract`
- Input seen in logs: PDF, 32 pages, about 8.8 MB

## Minimal config shape
```yaml
output_format: markdown

ocr:
  backend: tesseract
  language: eng
  output_format: markdown
  tesseract_config:
    psm: 3
    oem: 1
    output_format: markdown
    preprocessing:
      denoise: true
      contrast_enhance: true
```

## Steps to reproduce
1. Start the API server with Tesseract OCR enabled and the `preprocessing` block shown above.
2. Send a PDF to `/extract`.
3. Wait for OCR to start processing page images.

## Expected
The request should either complete successfully or return a normal extraction/configuration error.

## Actual
The server aborts, the container exits with code `133`, and the client receives a socket reset / hang up error.

Client side symptom:
```text
GaxiosError: request to http://localhost:8000/extract failed, reason: socket hang up
code: 'ECONNRESET'
```

Server log excerpt:
```text
... kreuzberg::ocr::processor::execution: stage="set_image" ...
fatal runtime error: Rust cannot catch foreign exceptions, aborting
kreuzberg-1 exited with code 133 (restarting)
```

## Notes
- The failure happens after Tesseract initialization and repeated `set_image` log entries.
- This looks like a hard crash in the preprocessing/Tesseract path rather than a handled request error.
- If useful, I can provide a smaller public PDF reproducer or test additional preprocessing combinations.

kh3rld:
`TessBaseAPIDetectOrientationScript` propagates a C++ exception when `osd.traineddata` is absent. The exception passes through Tesseract's `-fno-exceptions` frames and hits Rust's `extern "C-unwind"` boundary, triggering a runtime abort (exit 133 / SIGTRAP). Clients see ECONNRESET.

## What changed

**Behavioral change (`### Changed`):** `ImagePreprocessingConfig::default()` now returns `auto_rotate: false`. Callers that relied on the default enabling auto-rotation must set `auto_rotate: true` explicitly.

**Bug fix (`### Fixed`):** A C++ exception-barrier shim (`shim.cpp`, compiled with exception support) wraps `TessBaseAPIRecognize`, `GetHOCRText`, `GetUTF8Text`, `Clear`, and `DetectOrientationScript`. Any C++ exception from these calls is caught before it can reach Rust frames and converted to a graceful error return. `execution.rs` handles the graceful error with a warn-and-continue, so extraction succeeds even when OSD fails.

## Note

**System Tesseract (without `build-tesseract` feature):** the shim is compiled only when `build-tesseract` is active. On a system Tesseract install, the raw `TessBaseAPI*` calls remain exposed and the original exception propagation risk stands. This PR does not address that path.

**Stale binding defaults:** language binding packages (`python`, `java`, `csharp`, `swift`, `kotlin-android`, `php`, `wasm`) still default to `auto_rotate: true` because alef 0.24.13 (the pinned version) is not installed locally — regenerating with 0.23.70 would regress the C# `.csproj`. With the shim in place, a binding caller sending `auto_rotate=true` does not crash; OSD failure is caught, logged as a warning, and OCR continues. 

Closes #1089.

## Test plan

- [ ] `test_auto_rotate_true_does_not_abort` - explicit `auto_rotate: true` must return `Ok` (shim converts OSD exception to graceful warn+continue)
- [ ] `test_preprocessing_default_does_not_crash` - default config with preprocessing must not abort
- [ ] CI: confirm `build-tesseract` feature is active in the Tesseract CI job - otherwise both new tests are skipped and the shim is never exercised
- [ ] Follow-up: `chore: regen bindings` after installing alef 0.24.13, verify all binding defaults propagate to `false`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -443,6 +443,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   page-assembly block now call `is_page_text_blank` after every content mutation.
   ([#1095](https://github.com/kreuzberg-dev/kreuzberg/issues/1095))
 
+- **`ImagePreprocessingConfig` default `auto_rotate` changed from `true` to `false`; server
+  no longer aborts (exit 133 / SIGTRAP) when `preprocessing` is set in the server config.**
+  The previous default caused `TessBaseAPIDetectOrientationScript` to be called on every
+  image, which can propagate a C++ exception through Tesseract's `-fno-exceptions` frames
+  into Rust's `extern "C-unwind"` boundary, triggering a runtime abort. A C++ exception-
+  barrier shim now wraps the five most dangerous Tesseract C API calls on native builds so
+  that any escaping exception is converted to a graceful error return instead of aborting the
+  process. To keep auto-rotation, set `auto_rotate: true` explicitly.
+  ([#1089](https://github.com/kreuzberg-dev/kreuzberg/issues/1089))
+
 - **reranker: `RerankError` migrated to `thiserror`.** Matches the rest
   of the library and `rust-conventions`.
 
--- a/crates/kreuzberg-tesseract/build.rs
+++ b/crates/kreuzberg-tesseract/build.rs
@@ -719,9 +719,22 @@ mod build_tesseract {
         eprintln!("Bundled tessdata dir: {:?}", bundled_tessdata_dir);
 
         println!("cargo:rerun-if-changed=build.rs");
+        println!("cargo:rerun-if-changed=src/shim.cpp");
         println!("cargo:rerun-if-changed={}", third_party_dir.display());
         println!("cargo:rerun-if-changed={}", leptonica_dir.display());
         println!("cargo:rerun-if-changed={}", tesseract_dir.display());
+
+        // Compile the C++ exception-barrier shim (native builds only; `cc` is not
+        // available under build-tesseract-wasm). The shim must be compiled WITH
+        // exception support (no -fno-exceptions) so catch (...) is active. It wraps
+        // the dangerous Tesseract C API calls so C++ exceptions cannot propagate
+        // into Rust's extern "C-unwind" frames and abort the process.
+        #[cfg(feature = "build-tesseract")]
+        cc::Build::new()
+            .file("src/shim.cpp")
+            .cpp(true)
+            .include(tesseract_install_dir.join("include"))
+            .compile("kreuzberg_shim");
 
         println!("cargo:rustc-link-search=native={}", leptonica_lib_dir.display());
         println!(
--- a/crates/kreuzberg-tesseract/src/api.rs
+++ b/crates/kreuzberg-tesseract/src/api.rs
@@ -399,11 +399,12 @@ impl TesseractAPI {
     /// Returns `Ok(())` if recognition is successful, otherwise returns an error.
     pub fn recognize(&self) -> Result<()> {
         let handle = self.handle.lock().map_err(|_| TesseractError::MutexLockError)?;
-        // SAFETY: TessBaseAPIRecognize() is safe because:
-        // 1. *handle is a valid pointer to an initialized Tesseract engine with an image set
-        // 2. std::ptr::null_mut() is passed as the monitor parameter, which Tesseract accepts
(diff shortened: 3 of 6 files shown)


Goldziher (changes requested):
The fix direction (C++ exception barrier + default flip) is reasonable, but the current state has gaps that should land before merge.

## Blockers

**System-Tesseract path still aborts.** The shim is `#[cfg(feature = "build-tesseract")]`-only. Without that feature the raw `TessBaseAPIDetectOrientationScript` is called and a missing `osd.traineddata` still SIGABRTs. The stated goal — "prevent server abort when preprocessing config is set" — is not met for system-Tesseract users. Either gate `auto_rotate: true` as a runtime error / loud warning on non-`build-tesseract` builds, or compile a system-Tesseract variant of the shim. As-is the fix is silently partial.

**Binding defaults did not actually regenerate, despite the commit title.** `packages/csharp/src/Kreuzberg/ImagePreprocessingConfig.cs:33` still has `AutoRotate { get; init; } = true;`. The PR body acknowledges regen was deferred (alef 0.24.13 not installed), but commit `a8dc5130 chore: regenerate bindings for auto_rotate default change` claims otherwise. The two contradict each other. Either drop/rename that commit, or actually regenerate. Cross-language inconsistency (Rust default `false`, C#/Python/Java/Swift/PHP/WASM defaults `true`) is its own bug.

## Important

**`execution.rs` warn-and-continue is unverified.** The PR description and the test comment in `crates/kreuzberg/tests/ocr_errors.rs:64` cite `execution.rs`, but no such file is in the diff. The two new tests only assert `result.is_ok()` — they do not assert that text extraction proceeded after OSD failure, so the graceful-degradation claim isn't actually tested. Please point to the warn-and-continue site and add an assertion (non-empty text, expected metadata) that would fail if the path regressed.

**New tests are not feature-gated.** `test_auto_rotate_true_does_not_abort` and `test_preprocessing_default_does_not_crash` have no `#[cfg(feature = "build-tesseract")]` guard. Without that feature they exercise the unprotected raw API; if `osd.traineddata` is missing in that environment the test process aborts instead of failing cleanly. The PR's own test plan flags this risk. Add the cfg gate, or document the required feature in the test module.

**`shim.cpp` over-wraps.** `kreuzberg_tess_get_utf8_text` and `kreuzberg_tess_clear` are wrapped but neither touches OSD nor throws in normal operation. Wrapping every API call adds per-call overhead and obscures which calls are genuinely at risk. Keep `Recognize`, `GetHOCRText`, `DetectOrientationScript`; drop `GetUTF8Text` and `Clear` unless there's a documented throw case.

## Warnings / nits

- **MSVC exception model unset.** `build.rs`'s `cc::Build` for `shim.cpp` sets no `/EHsc`. Works on Linux/macOS via Itanium EH but is fragile on MSVC where exception unwinding across `extern "C"` frames is a different ABI. Add `.flag_if_supported("/EHsc")` and call out the cross-`extern "C"` unwind as ABI-dependent (technically UB per standard; works in practice on Itanium EH).
- **CHANGELOG splits one fix across `### Changed` and `### Fixed`.** Merge into a single `### Fixed` entry with the breaking-default callout inline so readers don't have to cross-reference two sections.
- **Default flip belongs under `### Breaking` rather than `### Changed`.** Silent semantic change for every caller using the default. v5 RC permits breaking changes — please mark it that way.
- **Shim compile bypasses the project's CXX-compiler selection.** `resolve_cxx_compiler` is used for the CMake path but not for `cc::Build` in the new shim branch; on Android cross-builds the `cc` crate may pick the wrong compiler. Pipe the resolved compiler in via `cc::Build::compiler(...)`.

PR #1111, opened 2026-06-13
fix(pdf): recompute page boundaries in metadata after OCR on scanned PDFs

### LINKED ISSUE #1110: bug(pdf): recompute page boundaries in metadata after OCR on scanned PDFs
### Description

  pdf/oxide/text.rs: computes PageBoundary offsets by measuring content.len() before and after appending each page's native text:
```rust 
  let byte_start = content.len();
  content.push_str(&cleaned);
  let byte_end = content.len();
```
  For a fully scanned PDF, cleaned is "" for every page (no embedded text layer). The resulting boundaries are `{byte_start: 0, byte_end: 0, page_number: 1}, {2, 2, 2}, {4, 4,3}` …  degenerate zero-length spans separated only by the `"\n\n"` separators. This is what the reporter sees as "tiny increments."

  pdf/oxide/metadata.rs:247 stores these stale boundaries into PageStructure.boundaries, which ends up in result.metadata.pages.boundaries (mod.rs).

  OCR then runs (mod.rs) and correctly writes real text into every PageContent.content via the zip loop (PR #1106 also recalculates is_blank here). But result.metadata.pages.boundaries is never touched; it still holds the degenerate positions computed against the empty native text.

### Steps to reproduce

` core/pipeline/features.rs` has recompute_boundaries_from_pages(), which searches for each page's content in result.content and returns correct byte offsets. It is called but only when chunking is configured, and the result is only used locally for chunk page metadata. It is never written back to result.metadata.pages.boundaries.

```text
crates/kreuzberg/src/core/pipeline/features.rs
```

kh3rld:
For fully scanned/rasterized PDFs (no embedded text layer), `metadata.pages.boundaries` reported degenerate `byteStart == byteEnd` for every page. PageBoundary offsets were computed against the empty native text before OCR ran, and never updated after OCR wrote real content into result.pages.

  The pipeline now runs `refresh_page_boundaries` unconditionally before chunking. It re-derives boundaries against `result.content` (the fully rendered string returned by the API) and writes them back to `metadata.pages.boundaries`. Chunking then reads the already-correct metadata instead of recomputing a second time.

  **Side-effect for native PDFs:** boundaries previously pointed into the raw extractor string (never returned by the API). They now consistently point into `result.content`, which is the correct invariant. This is a silent but intentional improvement.

  closes  #1110 the second part of #1095 (follow-up to #1106 which fixed  `isBlank` in the same write path).

  ## Tests
  - `refresh_page_boundaries_fixes_degenerate_offsets_after_ocr` - main regression
  - `refresh_page_boundaries_normalises_native_pdf_offsets_to_result_content` - native PDF side-effect
  - `refresh_page_boundaries_is_idempotent` - calling twice yields identical boundaries
  - `refresh_page_boundaries_no_op_when_pages_absent` / `_when_metadata_pages_absent` - guard clauses

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -78,12 +78,23 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   implicit `Item` wrapper instead of falling back onto the bare `List`, which violated
   CommonMark's `List → Item-only` constraint and panicked in debug builds.
   ([#1096](https://github.com/kreuzberg-dev/kreuzberg/issues/1096))
-  
+
 - **pdf: `result.pages[*].isBlank` now reflects OCR content for scanned/rasterized PDFs.**
   When OCR (including VLM) wrote text into existing `PageContent` entries, `is_blank` was
   never recalculated — it retained the stale value from native text extraction, which is
   always `Some(true)` for pages with no text layer. All four write sites in the OCR
   page-assembly block now call `is_page_text_blank` after every content mutation.
+  ([#1095](https://github.com/kreuzberg-dev/kreuzberg/issues/1095))
+
+- **pdf: `metadata.pages.boundaries` now reflects `result.content` offsets after OCR.**
+  For scanned/rasterized PDFs with no embedded text layer, `PageBoundary` offsets were
+  computed against the empty native text, producing degenerate `byte_start == byte_end`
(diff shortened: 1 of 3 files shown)


Goldziher:
Superseded by #1127 (which is now being re-implemented directly on main as the original commits were empty).

PR #1126, opened 2026-06-16
fix(config): default ImageExtractionConfig.classify to false

### LINKED ISSUE #1116: ImageExtractionConfig.classify default-on emits opaque "Vertex AI requires a project ID" warning when no VLM provider is configured
## Title
`ImageExtractionConfig.classify` default-on emits opaque "Vertex AI requires a project ID" warning when no VLM provider is configured

## Summary
`ImageExtractionConfig::default()` sets `classify: true`
(`crates/kreuzberg/src/core/config/extraction/types.rs:92,166`). With the
default value left untouched, kreuzberg's image classifier path runs
through liter-llm's VLM provider chain even when the deployment has not
configured a VLM provider. The provider chain then falls through to the
Vertex AI provider's auth check and surfaces

```
Vertex AI requires a project ID
```

as a warning attached to the extraction result. Callers cannot tell
whether this means "we tried Vertex but cannot use it, falling back to
nothing" or "your config is wrong" — and there is no path to make the
classifier go away short of explicitly setting `images.classify = false`
on every request.

## Reproduction
- Set `extract_images: true` on an extract request.
- Do not configure a VLM provider (no `VERTEXAI_PROJECT`, no
  `OPENAI_API_KEY`, no `GEMINI_API_KEY`, no per-request `ocr.vlm_config`).
- Run extraction against any document with embedded images.
- Observe the "Vertex AI requires a project ID" warning in the result.

## Suggested fix
Default `classify` to `false` and only flip it to `true` when at least
one VLM provider auth check succeeds at config-load time. Alternative:
keep the default at `true` but short-circuit the classifier with a typed
`ClassifyDisabled { reason: NoVlmProviderConfigured }` warning instead
of leaking the provider's auth error.

## Cloud-side workaround
For kreuzberg-cloud we are shipping a worker-side guard that forces
`classify=false` when neither `KREUZBERG_VLM_OCR_MODEL` is set nor a
per-request `vlm_config` is supplied. That mirrors what would be ideal
behaviour upstream.

Goldziher:
## Summary

Changes the `classify` field in `ImageExtractionConfig` to default to `false` instead of `true`. Image classification now must be explicitly enabled by users who need it, reducing unexpected overhead for users who don't require this feature.

## Changes

- Updated serde default for `classify` field
- Updated `Default` impl to set `classify: false`
- Updated existing test assertion
- Added 5 regression tests covering default, serialization, and backward compatibility

## Test Plan

- Unit tests pass: `cargo test -p kreuzberg --lib core::config::extraction::types::tests` (23 passed)
- Integration tests pass: `cargo test --test issue_1116_classify_default` (5 passed)
- All pre-commit hooks pass

Fixes #1116


Goldziher:
Closing: original commits were empty. Fix being re-implemented directly on main.

PR #1127, opened 2026-06-16
fix(pdf): recompute page boundaries in metadata after OCR

### LINKED ISSUE #1110: bug(pdf): recompute page boundaries in metadata after OCR on scanned PDFs

Goldziher:
After OCR fills page_contents with new text, the original page_structure boundaries (computed against native extraction) become invalid. Recompute boundaries against the OCR-filled content to ensure byte offsets remain valid for downstream consumers like the chunker.

Fixes #1110

## Test plan
- Unit test added: `recompute_boundaries_after_ocr_fills_scanned_pdf` verifies boundaries are recomputed when OCR fills initially-empty pages
- Existing boundary recomputation tests still pass
- `cargo test -p kreuzberg --lib --features chunking recompute_boundaries` all pass


Goldziher:
Closing: original commits were empty. Fix being re-implemented directly on main.

PR #1136, opened 2026-06-19
fix(derive): apply output_format to per-page content (#1094)

### LINKED ISSUE #1094: bug: result.pages content is plain text even with markdown output
### Description

# Bug report: `result.pages[*].content` is plain text even when markdown output is enabled

## Summary

When Kreuzberg is configured for Markdown output and page extraction is enabled, `result.content` is emitted as Markdown, but `result.pages[*].content` is still plain text / tab-separated text rather than Markdown.

In our case:

- `result.content` contains Markdown structure
- `result.chunks[*].content` contains Markdown structure
- `result.tables[*].markdown` contains Markdown table output
- `result.document.nodes` also preserves structure
- but `result.pages[*].content` loses the Markdown formatting and returns flattened text

This is surprising because the same extraction run is configured for Markdown output.

## What happened?

Example observed behaviour on a PDF containing a cap table:

- `result.content` includes Markdown headings and a Markdown table
- `result.chunks[0].content` includes Markdown headings and a Markdown table
- `result.pages[0].content` is plain text paragraph content plus tab-separated rows like:

```text
ID\tShareholder\tShare Class\tSecurity Type\tJurisdiction\tShares / Options\tOwnership %\tFully Diluted %\tNotes\tIssue Date
SH-001\tAlexander Rossi\tCommon\tOrdinary/Common\tSwitzerland\t3,500,000\t21.97%\t21.97%\tFounder grant\t2021-03-15
...
```

The same extraction result also contains a Markdown table in `result.tables[0].markdown`, so the structure is known but not preserved in page content.

## What did you expect to happen?

One of these should be true:

1. If `output_format = markdown`, `result.pages[*].content` should preserve Markdown structure.
2. Or, if page content is intentionally plain text, the docs / API contract should explicitly say that `result.pages[*].content` is normalized plain text even when the document output format is Markdown.

Right now the behaviour looks inconsistent with the configured output mode.

## Environment

- Kreuzberg Docker image: `ghcr.io/kreuzberg-dev/kreuzberg:4.9.9-core`
- Local config includes:
  - `output_format: markdown`
  - `include_document_structure: true`
  - `ocr.output_format: markdown`
  - `pages.extract_pages: true`

## Config snippet

```yaml
output_format: markdown
include_document_structure: true

ocr:
  output_format: markdown
  backend: vlm

pages:
  extract_pages: true
```

## Steps to reproduce

1. Start Kreuzberg with Markdown output enabled.
2. Enable page extraction with `pages.extract_pages: true`.
3. Extract a PDF that visibly contains headings and a real table.
4. Compare:
   - `result.content`
   - `result.tables[*].markdown`
   - `result.chunks[*].content`
   - `result.document.nodes`
   - `result.pages[*].content`
5. Observe that page content is plain text / tab-separated text instead of Markdown.

## Why this matters

Downstream consumers often use `result.pages[*].content` for page-level classification, routing, and retrieval. If page content loses Markdown structure while `result.content` preserves it, heading and table semantics are lost exactly at the page layer where many downstream systems operate.

If helpful, I can provide a sample PDF and a full raw extraction payload.

kh3rld:
## What

`result.pages[*].content` was always plain text regardless of `output_format`. Only `result.content` was formatted. Callers requesting `Markdown`/`Djot`/`Html` received correctly formatted full-document content but raw element text on every individual page.

## How

Adds `apply_page_content_format`, called in `derive_extraction_result` while the `InternalDocument` still owns its element data (before `derive_document_structure_inner` moves them). The function:

1. Builds a `BTreeMap<page_number → element_indices>` from page-tagged elements.
2. For each page, constructs a sub-document from that page's elements only and renders it with the same renderer used for the full document.
3. Remaps table indices into sub-doc space rather than cloning the full `doc.tables` slice per page (`Vec<Vec<String>>` cells are not Arc-backed; per-page full clone is O(all cells × num_pages) on large documents).

**Extractor-prebuilt pages are preserved unchanged** in two cases:
- `elements_by_page.is_empty()`: no elements carry page numbers (native PDF extraction, image OCR, Excel, PPTX). The whole `apply_page_content_format` call returns the prebuilt content untouched.
- Per-page `else` branch: a page exists in `prebuilt_pages` but has no elements tagged with its page number. That page's content is returned as-is; other pages on the same doc still get formatted.

`OutputFormat::Json`, `Plain`, `Structured`, and `Custom` are no-ops. Json renders the entire document as a single structured object — splitting it into per-page sub-objects would produce malformed fragments. Pages keep raw extracted text.

## Tests

| Test | What it covers |
|------|----------------|
| `page_content_markdown_heading_is_formatted` | `# h1` in per-page content with Markdown format |
| `page_content_plain_format_unchanged` | Plain format is a no-op on pages |
| `page_content_per_page_isolation` | No content bleed between pages |
| `page_content_markdown_list_items_formatted` | List items rendered with `-` syntax |
| `page_content_html_format_renders_headings` | Html renderer path is exercised |
| `page_content_prebuilt_pages_no_page_elements_unchanged` | `is_empty()` early-return: native PDF/OCR/Excel pages untouched |
| `page_content_page_without_matching_elements_unchanged` | Per-page `else` branch: page with no tagged elements keeps content |
| `page_content_json_format_pages_stay_raw` | Locks the Json asymmetry: `result.content` = JSON, `pages[*].content` = raw text |

Tests gated on `#[cfg(any(feature = "ocr", feature = "office", feature = "pdf", feature = "paddle-ocr", feature = "xml", feature = "hwpx", feature = "quality", feature = "chunking"))]` to match the `with_page` method's own feature gate. `page_content_prebuilt_pages_no_page_elements_unchanged` has no gate — it uses no `with_page` calls and compiles under all feature sets.

closes #1094

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -306,7 +306,6 @@ packages/r/src/*.o
 
 # BEGIN ai-rulez (DO NOT EDIT - managed by ai-rulez)
 .agents/
-.ai-rulez/.generated-manifest.json
 .cursor/
 .github/agents/
 .github/commands/
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- **derive**: `result.pages[*].content` now respects `output_format`. Previously, per-page content was always plain text even when `OutputFormat::Markdown`, `Djot`, or `Html` was requested — only `result.content` was formatted by `apply_output_format`. The new `apply_page_content_format` step re-renders each page's element subset with the same renderer used for the full document. Pages built from prebuilt extractor content that have no page-tagged elements (native PDF, image OCR, Excel, PPTX) are returned unchanged (#1094).
+
 ## [5.0.0-rc.23] - 2026-06-18
 
 ### Changed
--- a/crates/kreuzberg/src/extraction/derive.rs
+++ b/crates/kreuzberg/src/extraction/derive.rs
@@ -610,7 +610,7 @@ pub fn derive_extraction_result(
     //    over reconstructing from element-level page numbers.
     let raw_pages = doc.prebuilt_pages.take().or_else(|| build_pages(&doc));
     // Apply the requested output format to per-page content. Must run while doc still
-    // owns its elements (before derive_document_structure_inner moves them). Issue #1094.
+    // owns its elements (before derive_document_structure_inner moves them).
     let pages = apply_page_content_format(raw_pages, &doc, &output_format);
     // Prefer pre-built OCR elements stored directly by the extractor (e.g. image OCR
     // via inject_ocr_elements_from_vec was replaced by prebuilt_ocr_elements to avoid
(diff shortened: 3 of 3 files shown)


Goldziher:

PR #1137, opened 2026-06-19
feat(chunking): add TableChunkingMode::RepeatHeader for split table context

### LINKED ISSUE #1100: bug: split table chunks lose header/context after the first chunk
## Summary

When a single table is split across multiple `result.chunks`, only the first chunk keeps the table header. Later chunks contain only raw rows, so downstream extraction loses the column context.

This makes chunk-level consumers much weaker for entity/event extraction, because rows in chunk `n+1` no longer carry the semantic meaning of each column.

## Current behavior

Using a local Kreuzberg output sample (`doc2.json` in my `tmp` dir):

- `tables[0].markdown` contains the full table in one piece.
- `chunks[0]` starts with the header row.
- `chunks[1]` and `chunks[2]` continue the same logical table, but without repeating the header.

Example:

- `chunks[0]` starts with:
  - `| ... Shareholder | ... Share Class | ... Jurisdiction | ... Shares / Options | ... Ownership % | ... Issue Date |`
- `chunks[1]` starts with:
  - `| SH-006 Nathan Brooks | Seed Preferred | Preferred | Shares United Kingdom | 300,000 | ... |`
- `chunks[2]` starts with:
  - `| SH-013 Summit Horizon Growth | Fund Series B | Preferred Preferred | Shares Netherlands | 760,000 | ... |`

At that point, consumers of `chunks[1]` or `chunks[2]` cannot reliably know which value belongs to which column unless they manually recover context from the previous chunk or from `tables[0]`.

## Expected behavior

Chunked output should preserve table context for every chunk that contains part of a table.

## Potential solutions

A configurable approach would be ideal. For example:

1. **Keep each table in a single chunk**, even if it exceeds the normal chunk character limit.
2. **Repeat the table header in every chunk** that contains rows from the same logical table.

Something like a chunking option could work well here, e.g.:

- `table_chunking: "single_chunk"`
- `table_chunking: "repeat_header"`
- default behavior can remain unchanged if needed for backwards compatibility.

## Why this matters

For retrieval/extraction pipelines, `chunks` are often the primary unit passed to downstream systems. If a table is split without header carry-over, later chunks lose the schema of the data:

- shareholder / investor name
- share class / security type
- jurisdiction
- amount / ownership
- issue date

That makes table-derived facts much harder to extract correctly.

## Additional context

In the sample I tested, the full table is already available in `tables[0].markdown`, so the information exists. The issue is specifically that the `chunks` representation drops the table schema once the table spans more than one chunk.

kh3rld:
Markdown tables that exceed the chunk size are split at row boundaries, leaving continuation chunks without a header row. Downstream consumers,... search indexes, LLM prompts, structured extractors,... then lose column context entirely.

  Add `TableChunkingMode` enum (`Split` | `RepeatHeader`) and a `table_chunking` field on `ChunkingConfig`. When set to `RepeatHeader` and `chunker_type` is `Markdown`, a post-split pass (`inject_table_headers`) prepends the header row + separator to every continuation chunk. Default is `Split` ; no behavior change for existing callers.

  fixes #1100

  ## What changed
  - `TableChunkingMode` enum with `#[serde(default)]` field on `ChunkingConfig`; fully backwards-compatible, existing JSON configs deserialize correctly
  - `inject_table_headers` post-pass: O(n chunks), resets context on non-table chunks, handles multi-table documents correctly
  - 6 unit tests including: split/no-split comparison, two-table isolation, Text chunker no-op, single-chunk no-duplication, overlap > 0 correctness, serde default round-trip
  - All language bindings regenerated via `alef generate`; 4 new type files added (C#, Elixir, Java, Kotlin Android)

  ## Note
  - `rustdoc-lint` fails pre-commit on this Linux dev machine (candle-metal-kernels pulls in `objc2` + `cudarc`; pre-existing platform issue, not caused by this PR). CI on macOS/windows should be clean.
  - Pre-existing unterminated block doc-comment in the generated PHP binding (MIME wildcards `image/*` inside `/** */`) fixed in the test commit; root fix belongs in the alef generator.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -306,7 +306,6 @@ packages/r/src/*.o
 
 # BEGIN ai-rulez (DO NOT EDIT - managed by ai-rulez)
 .agents/
-.ai-rulez/.generated-manifest.json
 .cursor/
 .github/agents/
 .github/commands/
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- **chunking**: `TableChunkingMode` enum and `ChunkingConfig::table_chunking` field. When set to `RepeatHeader`, the markdown chunker prepends the table header row and separator to every continuation chunk produced by splitting a large table, so each chunk is self-contained for extraction, search, and LLM consumption. Default is `Split` (previous behavior, no header injection). Only applies when `chunker_type` is `Markdown` (#1100).
+
 ## [5.0.0-rc.24] - 2026-06-19
 
 ### Changed
--- a/crates/kreuzberg/src/chunking/config.rs
+++ b/crates/kreuzberg/src/chunking/config.rs
@@ -3,7 +3,7 @@
 use serde::{Deserialize, Serialize};
 
 // Re-export ChunkingConfig and ChunkerType from core config (canonical location)
-pub use crate::core::config::processing::{ChunkSizing, ChunkerType, ChunkingConfig};
+pub use crate::core::config::processing::{ChunkSizing, ChunkerType, ChunkingConfig, TableChunkingMode};
 
 /// Result of a text chunking operation.
 ///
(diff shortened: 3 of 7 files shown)


Goldziher:

PR #1143, opened 2026-06-20
feat(images): add data_base64 opt-in field to ExtractedImage

### LINKED ISSUE #1117: feat: ExtractedImage.data JSON wire size: add base64 (or hex) encoding option
## Title
`ExtractedImage.data` JSON wire size: add base64 (or hex) encoding option

## Summary
`ExtractedImage.data: Bytes` is currently serialised by `serde` defaults
as an array of integers (0..=255 per byte). A 256×159 PNG that
compresses to ~6 KB becomes a JSON integer array roughly 3× larger than
a base64 string and ~20× larger than the raw bytes. Customers consuming
the API from a JSON-only client (no `Bytes`-aware deserialiser) end up
with arrays of tens to hundreds of thousands of integers per image,
which is awkward to parse and costly to transport.

## Suggested change
Either

1. Add an `image_encoding` knob to `ImageExtractionConfig` that accepts
   `"png" | "jpeg" | "webp" | "base64"` (and possibly `"hex"`), with
   `"base64"` flipping the field to a JSON string instead of an integer
   array, OR
2. Expose a companion `data_base64: Option<String>` field that callers
   can opt into via request flag, leaving `data` untouched for backward
   compatibility.

Option 1 is cleaner but more invasive (touches serde, all language
bindings, and OpenAPI schemas). Option 2 is the minimum viable
improvement that unblocks customers without a backward-compat break.

## Reference
Customer evaluation report (Dutch/EU government docs platform) flagged
this as a friction point — the integer-array format was a clear
surprise when iterating on integration code.

kh3rld:
## Summary

Add `ExtractedImage::data_base64: Option<String>` populated when `ImageExtractionConfig::include_data_base64 = true`. closes #1117.

The default `bytes::Bytes` serde serialization emits a JSON integer array, which is inefficient for JSON-only clients. This field carries a Base64-encoded copy of `data` for callers that opt in. The field is absent from output by default, so existing clients are unaffected.

**Memory note**: enabling `include_data_base64` allocates a second copy of every image's bytes in the same response (roughly +33% over the raw size due to Base64 expansion). This is expected for an opt-in field and is called out in the doc comment. Rate limiting and response size enforcement remain the caller's responsibility, consistent with all other image flags.

## Changes

- `ExtractedImage::data_base64: Option<String>` - `#[serde(default, skip_serializing_if = "Option::is_none")]`, absent by default
- `ImageExtractionConfig::include_data_base64: bool` - opt-in flag, defaults to `false`
- `apply_data_base64_pass` in `pipeline/mod.rs` - runs at both async and sync call sites after all other transforms
- All explicit `ExtractedImage` struct literals updated with `data_base64: None`
- All Alef-managed bindings regenerated (C FFI, Python, Node, PHP, WASM, C#, Dart, Elixir, Go, Java, Kotlin Android, R, Ruby, Swift, Zig, JNI)
- Pipeline integration tests: `pipeline_populates_data_base64_when_opted_in` and `pipeline_omits_data_base64_when_not_opted_in`
- `alef verify` passes; `cargo check --no-default-features` passes (base64 is an unconditional dep)

## Test plan

- [ ] `cargo test --manifest-path crates/kreuzberg/Cargo.toml data_base64` - 5 tests pass
- [ ] `alef verify --exit-code` - all bindings fresh
- [ ] `cargo check -p kreuzberg --no-default-features` - base64 unconditional


Goldziher:

Issue #133, opened 2025-09-16
Extraction output too noisy — Markdown escaping & symbols not clean


mohamedGamalAbuGalala:
When extracting content from PDFs [google-doc-document (1).pdf](https://github.com/user-attachments/files/22366603/google-doc-document.1.pdf) or any PDF with a table, the result is cluttered with escaped characters and inconsistent symbol handling. The output loses readability and makes the Markdown quality low.

Actual Output:

```
[{"content":"Example document\n\nBeautiful is better than ugly.\n\nExplicit is better than implicit.\n\nSimple is better than complex.\n\nComplex is better than complicated.\n\nFlat is better than nested.\n\nSparse is better than dense.\n\nReadability counts.\n\nSpecial cases aren't special enough to break the rules.\n\nAlthough practicality beats purity.\n\nErrors should never pass silently.\n\nUnless explicitly silenced.\n\nn the face of ambiguity, refuse the temptation to guess.\n\nThere should be one\\-\\- and preferably only one \\-\\-obvious way to do it.\nAlthough that way may not be obvious at first unless you're Dutch.\nNow is better than never.\n\nAlthough never is often better than *right* now.\n\nIf the implementation is hard to explain, it's a bad idea.\n\nif the implementation is easy to explain, it may be a good idea.\nNamespaces are one honking great idea \\-\\- let's do more of those!\n\n...\n...\n...\n...\n...\nIndonesia ™ \\| Germany ®® \\| Austria \\=\\= \\| France Vatican \\[\\+\nContinent \\| Asia Europe\nCapital Jakarta Berlin Vienna Paris Vatican City\nCurrency Rupia EUR (€) \\-\nPopulation 273\\.879\\.750' \\| 83,190,556*\\| 8,935,112°\\| 67,413,000 453\n\n...\n...\n‘2021 estimate 2020 estimate\n$ 2020 estimate","mime_type":"text/plain","metadata":{"title":"PDF Example Document","created_by":"Skia/PDF m103 Google Docs Renderer","width":596,"height":842,"summary":"PDF document with 1 page. PDF version 1.4. Document is printable, modifiable, extractable."},"tables":[],"chunks":[],"images":[],"image_ocr_results":[],"entities":null,"keywords":null,"detected_languages":null,"document_type":null,"document_type_confidence":null,"layout":null}]
```

Expected Output (clean Markdown):

```
|               | Indonesia   | Germany    | Austria   | France    | Vatican      |
[...]

PR #1156, opened 2026-06-23
feat(config): adopt alef string_shorthand for model presets

tobocop2:
String shorthand for model presets is now available: alef [#133](https://github.com/xberg-io/alef/pull/133) (`feat(pyo3,magnus): string_shorthand for data-variant enum constructors`) landed and shipped in alef 0.28.0, which `main` already pins. The earlier blocker is resolved.

## Problem

Passing a bare preset name to the embedding/reranker model config crashes in the bindings. `EmbeddingConfig(model="fast")` (and the reranker `model="balanced"`) fail because the generated constructor encodes the bare string as the enum tag — `{"type":"fast"}` — which serde rejects, since `fast` is not a variant. It should build `Preset { name: "fast" }`.

## Solution

Opt the two enums into alef's `string_shorthand` mechanism:

```rust
#[cfg_attr(alef, alef(string_shorthand(variant = "Preset", field = "name")))]
```

on `EmbeddingModelType` and `RerankerModelType`. Both have a single-field `Preset { name }`, which satisfies the mechanism's constraint (a bare string fills exactly one field; any others must be optional). After regeneration the constructor emits `{"type":"preset","name":"fast"}`, so a bare preset name resolves to the `Preset` variant.

## Test

`fixtures/embed_extra/embed_texts_shorthand_preset.json` constructs the embedding model from a bare preset name (`"balanced"`) instead of the explicit `{ type, name }` form. It generates to `EmbeddingModelType("balanced")` (Python) and `model: 'balanced'` (Ruby) — the two backends alef 0.28.0 implements `string_shorthand` for — and is skipped for the others. The test fails against the older bindings (bare string serializes as `{"type":"balanced"}`, which serde rejects) and passes once the pyo3/magnus constructors emit `{"type":"preset","name":<s>}`.

## Scope

Source opt-in plus the test fixture only — no regenerated bindings. The `cfg(alef)` attribute is inert in a normal build (it affects alef extraction, not the compiled crate), so both the binding behavior and the generated e2e test materialize on the next 0.28.0 binding regen, which picks up the opt-in and the fixture automatically. Verified locally: regenerating with 0.28.0 emits `{"type":"preset","name":<s>}` for pyo3 and magnus, the generated code compiles, and the fixture produces the shorthand test shown above. Other backends are unchanged until they adopt the mechanism.

### DIFF
--- a/crates/kreuzberg/src/core/config/processing.rs
+++ b/crates/kreuzberg/src/core/config/processing.rs
@@ -391,6 +391,7 @@ impl Default for EmbeddingConfig {
 /// Embedding model types supported by Kreuzberg.
 #[derive(Debug, Clone, Serialize, Deserialize)]
 #[serde(tag = "type", rename_all = "snake_case")]
+#[cfg_attr(alef, alef(string_shorthand(variant = "Preset", field = "name")))]
 pub enum EmbeddingModelType {
     /// Use a preset model configuration (recommended)
     Preset {
--- a/crates/kreuzberg/src/core/config/reranker.rs
+++ b/crates/kreuzberg/src/core/config/reranker.rs
@@ -88,6 +88,7 @@ impl Default for RerankerConfig {
 /// Since v5.0.0.
 #[derive(Debug, Clone, Serialize, Deserialize)]
 #[serde(tag = "type", rename_all = "snake_case")]
+#[cfg_attr(alef, alef(string_shorthand(variant = "Preset", field = "name")))]
 pub enum RerankerModelType {
     /// Use a preset cross-encoder model (recommended).
     Preset {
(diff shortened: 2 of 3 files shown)


tobocop2:
Closed — the bare-string preset shorthand is the wrong direction. It's stringly-typed, only ever helps the single-field `Preset` case, and leaves `Custom`/`Llm`/`Plugin` untouched. The real gap is that the dynamic backends lack the per-variant constructors the statically-typed backends already generate. The proper fix is tracked upstream in alef: https://github.com/xberg-io/alef/issues/147 (per-variant constructors for data enums, parity across all backends), which supersedes alef #135.

Issue #1158, opened 2026-06-24
bug: PDF render produces blank pages for JPEG 2000 (JPXDecode) images, silently breaking OCR


tobocop2:
## Problem

`render_pdf_page_to_png` (and the OCR rasterization path that feeds the OCR backends) renders pages whose content is a JPEG 2000 image XObject (`/Filter /JPXDecode`) as an all-white bitmap. The renderer (`pdf_oxide` 0.3.66 + tiny-skia) doesn't decode `JPXDecode` and returns a blank page rather than erroring.

This surfaced while OCR'ing a scanned PDF whose pages are each a single full-page JPEG-2000 image: every OCR backend (tesseract, paddle, vlm, custom plugins) received a blank page, extracted nothing, and the document was silently skipped as "no usable text" — with no error or warning that an unsupported image filter had been dropped.

Minimal repro via the public API (no OCR needed):

```python
from kreuzberg import render_pdf_page_to_png
from PIL import Image
import io

png = render_pdf_page_to_png(open("scanned.pdf", "rb").read(), 0, 150, None)
print(Image.open(io.BytesIO(png)).convert("L").getextrema())  # (255, 255) -> blank white
```

A reference decoder renders the same page with full content (luminance extrema `(10, 241)`), confirming the PDF is valid and the gap is in `pdf_oxide`'s image handling. Affected sample: a 595-page scanned guide where every page is one 2123x2720, 24bpp `JPXDecode` image at 300 DPI.

JPEG 2000 (and, by the same token, likely JBIG2 and CCITT Group 3/4) are common in scanned and archival PDFs, so this silently breaks a whole class of scanned documents under OCR.

## Solution

Add JPEG 2000 (`JPXDecode`) decoding to `pdf_oxide`'s image path — e.g. via an OpenJPEG / `jpeg2k`-backed decoder — and ideally cover the other scan-common filters (`JBIG2Decode`, `CCITTFaxDecode`) while at it.

Until a page's image filter can be decoded, the renderer should not silently return a blank page: detect an unsupported image filter (or a page that has image content but rendered fully blank) and surface a warning/error so OCR doesn't silently no-op on the whole document.

tobocop2:
this can likely be closed soon due to the fix being in pdf oxide available today as of a few hours ago. I'm validating it and will close the issue if it is no longer relevant https://github.com/yfedoseev/pdf_oxide/releases/tag/v0.3.68

tobocop2:
[...]

PR #1159, opened 2026-06-24
fix(pdf): render image-only pages pdf_oxide can't rasterize instead of blank (#1158)

tobocop2:
> **🚧 Draft / work in progress.** Implementation is in place and compiles; tests, CHANGELOG, and the full CI/self-review pass are still to come. Opened early for visibility — do not merge yet.

## Problem

Scanned PDFs whose pages are stored as full-page JPEG 2000 images come out blank, and every OCR backend silently extracts nothing.

The 5.x renderer is pure-Rust `pdf_oxide`, which has no JPEG 2000 decoder. For a `/JPXDecode` image XObject, `extract_image_from_xobject` returns `UnsupportedFilter`, so `pdf_oxide`'s rasterizer drops the image and the page rasterizes to white. The OCR pipeline rasterizes each page and feeds it to the backend, so tesseract/paddle/vlm/custom all receive a blank page, extract zero text, and the document is skipped as "no usable text" — with no error or warning.

Minimal repro (public API):

```python
from kreuzberg import render_pdf_page_to_png
from PIL import Image; import io
png = render_pdf_page_to_png(open("scanned.pdf","rb").read(), 0, 150, None)
print(Image.open(io.BytesIO(png)).convert("L").getextrema())  # (255, 255) == blank
```

This is a regression from the 4.x → 5.x renderer swap: 4.x rendered via **pdfium**, which decodes JPEG 2000 natively. pdfium was removed in `124ba418` ("remove pdfium, migrate to pdf_oxide as sole PDF backend"), so the JPEG 2000 capability went with it. `GitHub #1158`.

## Solution

`pdf_oxide` stays the sole backend (pure-Rust, WASM-safe — reintroducing pdfium would reverse that deliberate migration). kreuzberg already ships pure-Rust decoders for these formats (`hayro-jpeg2000`, `hayro-jbig2`) for standalone image extraction, so the fix reuses them at render time.

New module `pdf/oxide/image_page.rs` renders an image-only page directly from its dominant image XObject when `pdf_oxide` can't rasterize it:

1. **Proactive fast path** — before rasterizing, if the page's dominant image uses a filter `pdf_oxide` can't decode (`/JPXDecode`), decode the codestream directly and skip the rasterizer. No wasted render, and *faster* than today for scans.
2. **Safety net** — if `pdf_oxide` renders an image-bearing page blank anyway, retry the direct decode before handing OCR a blank page.
3. **Loud failure** — when neither path can decode an image-bearing page, log a warning instead of silently returning blank.

`pdf_oxide` keeps doing all page composition (CTM, masks, vector, text); this only substitutes the image-decode step it lacks. For the image-only/scan case that is exact; rare pages mixing a JPEG 2000 image with vector content are recovered best-effort (dominant image) with a warning.

Wired through every rasterization entry point: the public `render_pdf_page_to_png`, the OCR pipeline (force-OCR and batch), and the layout runner.

### Scope: JPEG 2000 is the only gap

`pdf_oxide` already matches pdfium on CCITT (G3 **and** G4), JBIG2, DCT/JPEG, Flate, and LZW. JPEG 2000 is the single image format it cannot decode, so this closes the image-rendering parity gap left by the pdfium removal. The safety net additionally recovers any future/edge decode failure where an alternative in-tree decoder succeeds.

## Still TODO (why this is a draft)

- [ ] Regression tests: JPEG 2000 page renders non-blank via the public API; OCR pipeline + layout runner paths; safety-net + loud-failure behaviour; parity that `pdf_page_count` + `render_pdf_page_to_png` cover everything the removed `PdfPageIterator` did.
- [ ] CHANGELOG `[Unreleased]` entry.
- [ ] Full `cargo fmt` / `clippy -D warnings` / test pass + self-review across the changed surface.
- [ ] Refresh `#1158` to reflect the post-pdfium-removal root cause and chosen architecture.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- **PDF**: JPEG 2000 (`/JPXDecode`) pages no longer render blank. `pdf_oxide` has no JPEG 2000 decoder, so it dropped the image and rasterized the page to white; scanned PDFs stored as full-page JPEG 2000 images then produced zero OCR output across every backend, with no error. Pages whose dominant image XObject uses a filter `pdf_oxide` cannot rasterize are now decoded directly via the in-tree decoder set (`hayro-jpeg2000`), with a safety net for images `pdf_oxide` renders blank and a warning when a page image cannot be decoded at all (instead of a silent blank). Applies to `render_pdf_page_to_png`, the OCR pipeline, and the layout runner. JPEG 2000 was the only image format pdfium handled that `pdf_oxide` could not, so this closes the image-rendering gap from the pdfium → pdf_oxide migration. (#1158)
 - **build/node-musl**: the Alpine musl Node binding build sets `CC_/CXX_<target>=gcc/g++` so `cc-rs` (used by `ring`, tesseract) finds a compiler on napi-rs's cross path. It previously failed with `error occurred in cc-rs: failed to find tool "aarch64-linux-musl-gcc"` — the cargo `*_LINKER` vars were set but `cc-rs` resolves the *compiler* via its own per-target `CC_`/`CXX_` lookup. (`docker/Dockerfile.musl-node`)
 - **Ruby gem now ships precompiled platform binaries.** `ruby-gem` only ran `rake build`, emitting a
   source-only gem on every matrix leg, so `gem install kreuzberg` compiled the native extension from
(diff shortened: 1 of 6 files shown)


Goldziher:
@tobocop2 please also open an upstream issue in pdf_oxide for pages that cannot be rasterized

tobocop2:
looks like pdf oxide fixed this upstream, i'm going to close this and update the original issue.

tobocop2:
it was fixed today, which is wild https://github.com/yfedoseev/pdf_oxide/releases/tag/v0.3.68

PR #1161, opened 2026-06-25
fix(ner): build GLiNER ONNX backend against ort 2.0.0-rc.12

### LINKED ISSUE #1160: xberg-gliner fails to build against ort 2.0.0-rc.12 (breaks ner-onnx / full / the Python wheel)
## Problem

On `main` (latest, `0ec4fb26fa`), the `xberg-gliner` crate fails to compile, which breaks any build that enables the `ner-onnx` feature. `ner-onnx` is pulled by `full`, and `xberg-py` requests `full`, so `cargo check -p xberg-py` and the Python wheel build (`maturin build`) both fail.

```
error[E0277]: `?` couldn't convert the error to `GlinerError`
   --> crates/xberg-gliner/src/lib.rs:785:62
    |
783 | fn build_session<P: AsRef<Path>>(model_path: P, runtime: &RuntimeConfig) -> Result<Session> {
    |                                                                             --------------- expected `GlinerError` because of this
784 |     let session = Session::builder()?
785 |         .with_optimization_level(GraphOptimizationLevel::All)?
786 |         .with_intra_threads(runtime.intra_threads)?
```

`GlinerError` already derives `Ort(#[from] ort::Error)` (`crates/xberg-gliner/src/lib.rs:56`), but under `ort = "2.0.0-rc.12"` the `SessionBuilder` methods (`with_optimization_level`, `with_intra_threads`, …) no longer return `ort::Error` — they return a different error type — so `?` can no longer convert through the existing `#[from] ort::Error`.

## Repro

```
cargo check -p xberg-gliner    # or -p xberg-py
```

`main` @ `0ec4fb26fa`, `ort = { version = "2.0.0-rc.12", ... }` (workspace `Cargo.toml:83`). Confirmed the crate source is unchanged from the rebrand; the `ort` bump appears to have landed without updating the gliner backend to the rc.12 builder API.

## Suggested fix

Update `build_session` (and any other `ort` builder usage in the crate) for the rc.12 API — add the builder's error type to `GlinerError` as another `#[from]` variant, or `.map_err(...)` the builder calls into the existing `Ort` variant.

Found while building a trial Python wheel against latest `main`; the failure is in the hand-written `xberg-gliner` core crate, unrelated to codegen/bindings.

tobocop2:
## Problem

`ner-onnx` — pulled by `full`, and therefore by the Python wheel — stopped compiling against `ort 2.0.0-rc.12`. Two things the crate rebrand left inconsistent:

**`xberg-gliner` doesn't build against rc.12.** `build_session` applies `?` to the `SessionBuilder` option methods (`with_optimization_level`, `with_intra_threads`). Under rc.12 those return `ort::Error<SessionBuilder>` instead of `ort::Error`, so the `?` conversion into `GlinerError` no longer compiles. (#1160)

**The consumer still targets the removed crate.** `text::ner::gline` (`#[cfg(feature = "ner-onnx")]`) still imports the external `gline-rs` crates (`gliner`, `orp`) that the rebrand replaced with the in-tree `xberg-gliner`. Neither is a dependency anymore, so the feature also fails with unresolved-crate errors — which the `cargo check -p xberg-gliner` repro in #1160 hides, since it stops at the first failing crate.

## Solution

**Builder error.** Map the rc.12 `SessionBuilder` error into the plain `ort::Error` the `Ort` variant already accepts (`.map_err(|e| ort::Error::new(e.message()))`) — the same idiom the embeddings, transcription, reranking, layout, and doc_orientation session sites already use.

**Consumer.** Port `gline` to the `xberg_gliner` API. It's a drop-in: same `inference() -> SpanOutput`, same `Span` accessors, so only the imports and the constructor change.

**Test suite.** Compiling the crate for the first time since the rebrand ran its (until now dark) tests and surfaced two problems:

- A dead `Prompt::text_length` field — it duplicates `PromptInput::text_lengths` and fails `clippy --all-targets -- -D warnings`. Removed; the builder test now asserts the live field.
- `span_tensors_include_valid_span_indices_and_masks` asserted an impossible slot. For a 4-word text with `max_width = 3`, the last valid span `(3, 3)` sits at dimension `start * max_width + width` = 9; dimension 11 is the out-of-range span `(3, 5)` that `make_span_tensors` correctly leaves zeroed and masked. The test is corrected to check the real last valid span plus the padding — `make_span_tensors` matches GLiNER span-mode and is unchanged.

Also adds a `build_session` regression test: it exercises the rc.12 builder chain and confirms a missing model surfaces as `GlinerError` rather than panicking.

Verified with `cargo check -p xberg-py` (full), `cargo test -p xberg -p xberg-gliner --features full,ner-onnx --lib` (xberg 4399 passed, xberg-gliner 10 passed), and `cargo clippy -D warnings` on the affected scopes.

Closes #1160.

### DIFF
--- a/crates/xberg-gliner/src/lib.rs
+++ b/crates/xberg-gliner/src/lib.rs
@@ -317,15 +317,13 @@ impl Tokenizer for HFTokenizer {
 #[derive(Debug)]
 struct Prompt {
     tokens: Vec<String>,
-    text_length: usize,
     entities_length: usize,
 }
 
 impl Prompt {
-    fn new(tokens: Vec<String>, text_length: usize, entities_length: usize) -> Self {
+    fn new(tokens: Vec<String>, entities_length: usize) -> Self {
         Self {
             tokens,
-            text_length,
             entities_length,
         }
     }
@@ -372,7 +370,7 @@ impl PromptInput {
             let mut prompt = Vec::with_capacity(entities_prompt.len() + tokens.len());
             prompt.extend(entities_prompt.clone());
             prompt.extend(tokens.iter().map(|token| token.text().to_string()));
-            prompts.push(Prompt::new(prompt, tokens.len(), entities_prompt.len()));
+            prompts.push(Prompt::new(prompt, entities_prompt.len()));
             text_lengths.push(tokens.len());
             num_words = num_words.max(tokens.len());
         }
@@ -781,9 +779,13 @@ impl Gliner {
 }
 
 fn build_session<P: AsRef<Path>>(model_path: P, runtime: &RuntimeConfig) -> Result<Session> {
+    // ort 2.0.0-rc.12 builder option methods return `ort::Error<SessionBuilder>`; flatten to the
+    // plain `ort::Error` the `Ort` variant accepts (matches the other ORT session sites in xberg).
     let session = Session::builder()?
-        .with_optimization_level(GraphOptimizationLevel::All)?
-        .with_intra_threads(runtime.intra_threads)?
+        .with_optimization_level(GraphOptimizationLevel::All)
+        .map_err(|e| ort::Error::new(e.message()))?
+        .with_intra_threads(runtime.intra_threads)
+        .map_err(|e| ort::Error::new(e.message()))?
         .commit_from_file(model_path)?;
     Ok(session)
 }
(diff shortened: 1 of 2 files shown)


tobocop2:
Superseded by `1f2c2fc48c` ("chore: added xberg-gliner") on `main`, which lands the full xberg-gliner implementation and already fixes #1160:

- `session.rs` maps the rc.12 `SessionBuilder` error into `GlinerError::Ort` (the build_session fix this PR made).
- `text::ner::gline` is ported to the in-tree `xberg_gliner` API (the consumer fix this PR made).
- It also independently corrects the `span_tensors` test (asserting dim 9, not the impossible dim 11).

Verified `cargo check -p xberg --features ner-onnx` builds green on current `main`. Closing as redundant.

Issue #1165, opened 2026-06-25
bug: Python: EmbeddingModelType.llm()/EmbeddingConfig(model=LlmConfig) rejects the public LlmConfig class


tobocop2:
## Summary

In Python, building an LLM embedding model with the **public** `LlmConfig` fails:

```python
from xberg import EmbeddingConfig, EmbeddingModelType, LlmConfig

EmbeddingModelType.llm(LlmConfig(model="openai/text-embedding-3-small"))
# TypeError: 'LlmConfig' object is not an instance of 'LlmConfig'
```

This is the form our own docs show (`docs/snippets/python/llm/vlm_embeddings.md`), so the documented usage is broken. Consumers currently have to import the private `xberg._xberg.LlmConfig`, or pass a dict.

## Root cause

Two alef-generated mismatches:

1. **Inconsistent public re-export.** `packages/python/xberg/__init__.py` re-exports `EmbeddingModelType` from the compiled `._xberg` module, but `LlmConfig` from the pure-Python `.options` wrapper. So `EmbeddingModelType.llm()` (compiled factory `_factory_llm(llm: LlmConfig)`) only accepts the *compiled* `LlmConfig`, while `xberg.LlmConfig` is the dataclass.

2. **Enum-variant payloads aren't coerced like struct fields.** alef generates `LlmConfig` coercion for struct *fields* — e.g. `vlm_config` goes through `_to_rust_llm_config(...)`, which accepts `LlmConfig | dict | str`. But the `EmbeddingModelType::Llm { llm }` enum-variant payload gets no such coercion: `_to_rust_embedding_config` (api.py) just calls `_rust.EmbeddingModelType(value.model)`, whose `#[new]` only handles `str`/`dict`, not a dataclass. So even the idiomatic `EmbeddingConfig(model=LlmConfig(...))` path fails.

## Recommended fix

In the alef generator, coerce enum-variant struct payloads the same way struct fields are coerced — i.e. when an `EmbeddingModelType` value is (or contains) a public `LlmConfig` dataclass, route it through `_to_rust_llm_config` and build `EmbeddingModelType.llm(...)`. Equivalently, make the generated `.llm()` factory / `EmbeddingModelType(...)` constructor accept `LlmConfig | dict`. The public `LlmConfig` and `EmbeddingModelType` re-exports should also be consistent (both wrapper, or both compiled).

This is generator-level; `alef.toml` has no per-type union-coercion knob to express it.

[...]

PR #1166, opened 2026-06-25
docs(llm): document Python dict workaround for LLM embedding config

tobocop2:
## Problem

The Python VLM-embeddings docs (`docs/snippets/python/llm/vlm_embeddings.md`) showed:

```python
EmbeddingConfig(model=EmbeddingModelType.llm(LlmConfig(model="openai/text-embedding-3-small")))
```

That raises `TypeError: 'LlmConfig' object is not an instance of 'LlmConfig'`. The public `xberg.LlmConfig` (the dataclass re-exported from `.options`) isn't the compiled type that the `EmbeddingModelType.llm()` factory accepts, and the enum-variant payload isn't coerced the way struct fields are — so the documented usage is broken. Full root cause and the proposed generator-level fix are in #1165.

## Solution

This is a docs-only change to unblock users until the generator fix in #1165 lands — it does not touch the alef-generated bindings.

- **Working example.** The Python snippet now uses the dict form, which is accepted today: `EmbeddingConfig(model={"type": "llm", "llm": {"model": "..."}})`.
- **Guide note.** A note under *VLM Embeddings → Configuration* explains the limitation, links #1165, and lists the fields the nested `llm` dict accepts (same as `LlmConfig`).

Verified the snippet compiles (`python -m py_compile`).

Refs #1165.

### DIFF
--- a/docs/guides/llm-integration.md
+++ b/docs/guides/llm-integration.md
@@ -243,6 +243,15 @@ Use provider-hosted embedding models when you need to match your vector database
       --text "Hello world"
     ```
 
+!!! note "Python: pass the LLM model as a dict"
+
+    The typed `EmbeddingModelType.llm(LlmConfig(...))` constructor currently rejects the
+    public `xberg.LlmConfig` class. Until that is fixed
+    ([#1165](https://github.com/xberg-io/xberg/issues/1165)), specify the model as a dict —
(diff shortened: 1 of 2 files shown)


tobocop2:
apologies for the noise again. I meant to just create an issue. 

The generator-level fix tracked in #1165 is the solution.

PR #1189, opened 2026-07-01
build: remove stray tree-sitter-language-pack patch override

### LINKED ISSUE #1187: bug: stray tree-sitter-language-pack [patch.crates-io] override breaks fresh builds
## Description

The workspace `Cargo.toml` commits a `[patch.crates-io]` override that points `tree-sitter-language-pack` at a local sibling checkout:

```toml
[patch.crates-io]
# DEV-ONLY (uncommitted): local tslp override while fixing wasm32 grammar build.
# Remove once the fix is published to crates.io and the version floor is bumped.
tree-sitter-language-pack = { path = "../tree-sitter-language-pack" }
```

The comment marks it DEV-ONLY and says to remove it once the wasm32 grammar fix ships — which it has, and the workspace version floor is already `1.12.1`. The override leaked into the commit.

A fresh checkout without that exact sibling repo can't resolve it, so local `cargo build` and `maturin` wheel builds fail. CI is unaffected because it resolves from crates.io.

```
error: failed to load source for dependency `tree-sitter-language-pack`
Caused by: found a virtual manifest at `.../tree-sitter-language-pack/Cargo.toml` instead of a package manifest
```

(the sibling repo root is a virtual workspace; the actual package lives at `crates/ts-pack-core`).

**Expected:** the override is removed so `tree-sitter-language-pack` resolves to the published `1.12.1` from crates.io, keeping the `xberg*` path patches.

## Steps to reproduce

1. Fresh checkout of `main` without a `../tree-sitter-language-pack` sibling repo.
2. Run `cargo metadata` (or `cargo build`, or a `maturin` wheel build).
3. It fails with the `found a virtual manifest` error above.

tobocop2:
> found during trial integration of latest xberg rc, super minor but easy fix and it's causing CI failures.

The DEV-ONLY `[patch.crates-io]` override pointed `tree-sitter-language-pack` at a local sibling checkout, so fresh `cargo build` / `maturin` wheel builds fail to resolve it. The wasm32 fix it existed for is published and the floor is already `1.12.1`, so this drops the override and resolves from crates.io.

Fixes #1187.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -123,9 +123,6 @@ web-sys = { version = "0.3", features = [
 ] }
 
 [patch.crates-io]
-# DEV-ONLY (uncommitted): local tslp override while fixing wasm32 grammar build.
-# Remove once the fix is published to crates.io and the version floor is bumped.
-tree-sitter-language-pack = { path = "../tree-sitter-language-pack" }
 xberg = { path = "crates/xberg" }
 xberg-gliner = { path = "crates/xberg-gliner" }
 xberg-rag = { path = "crates/xberg-rag" }


Goldziher:
Redundant: the stray tslp [patch] override is removed as part of the in-flight alef 0.30.4 consolidation on main (which also bumps the tslp floor to the published wasm-fixed version). Closing in favor of that.

PR #1201, opened 2026-07-05
test(pdf): vertical-text pages must return text, not a guarded per-page error

tobocop2:
## Problem

PDFs with vertical (tategaki) text can make pdf_oxide's reading-order sort panic. Our panic guard keeps extraction alive, but the affected page's text is silently lost — the page comes back as an error instead of content.

## Solution

The real fix landed upstream in yfedoseev/pdf_oxide#808 and shipped in pdf_oxide 0.3.73. This PR bumps the workspace pin to 0.3.73 and adds a regression test that builds a small vertical-text PDF triggering the sort and asserts the page's text actually comes back, not just that extraction survives. The test fails against 0.3.72 and passes with 0.3.73.


tobocop2:
my pdf oxide fix has landed in the latest release 🥳  so going to revisit this now

PR #1218, opened 2026-07-06
fix(candle-ocr): fix the failing GPU OCR jobs (PaddleOCR-VL + Hunyuan-OCR)

### LINKED ISSUE #1214: bug: PaddleOCR-VL candle backend crashes on any non-square page (mrope index broadcast)
## Problem

The PaddleOCR-VL GPU job fails on every real image (candle_paddleocr_vl_e2e_extraction): https://github.com/xberg-io/xberg/actions/runs/28776698885

```
PaddleOCR-VL inference failed: ... H broadcast: cannot broadcast [1, 8] to [8, 26]
    at xberg_candle_ocr::models::paddleocr_vl::model::PaddleOCRVLModel::get_rope_index
```

`get_rope_index` builds the vision-block position ids for the multimodal rope. It constructs the h index as a `[1, h]` tensor and broadcasts it to `(h, w)`, which is only a legal broadcast when the patch grid is square — any real page (here 8 × 26 after spatial merge) fails inference outright. The block also stacks only two rows (h, w) where the surrounding position concat expects three, so even square grids would fail one step later at the concat, and the h values themselves are laid out along the wrong axis.

## Solution

Build the block the way the reference Qwen2-VL mrope (and upstream aha) does: three rows stacked as (t, h, w) over the flattened grid — t constant per frame (images are single-frame), h varying per grid row, w varying per grid column.

### LINKED ISSUE #1215: bug: Hunyuan-OCR rejects checkpoints whose config.json omits tie_word_embeddings
## Problem

The Hunyuan-OCR GPU job fails at model load (candle_hunyuan_ocr_e2e_extraction): https://github.com/xberg-io/xberg/actions/runs/28776698885

```
Hunyuan-OCR engine initialisation failed: model load failed:
Parse config: missing field `tie_word_embeddings` at line 190 column 1
```

`HunYuanVLConfig` declares `tie_word_embeddings` as a required field, but later revisions of the released checkpoint's config.json simply omit it — transformers treats it as optional with a default of `true`. A benign omission in the shipped config turns into a hard failure of the entire model load. The value doesn't even influence loading: the checkpoint ships no separate lm_head weight and the engine always ties the head to the input embeddings.

## Solution

Deserialize `tie_word_embeddings` with a serde default of `true` (the transformers default and what the checkpoint actually does), keeping explicit values honored.

### LINKED ISSUE #1216: bug: GPU CI can no longer download Hunyuan-OCR weights (tencent/HunyuanOCR pulled from the HF Hub)
## Problem

The GPU CI hunyuan-ocr job downloads its weights from `tencent/HunyuanOCR` on every run (the actions cache for multi-GB weights is evicted immediately — the repo carries thousands of cache entries against the 10 GB limit). On 2026-07-06, between the morning CI run (which still fetched the repo unauthenticated) and the afternoon, Tencent pulled the repo from the Hugging Face Hub: it now returns 401/404 even for authenticated requests, and the `HF_TOKEN` org secret is empty. The next GPU run fails at the download step before any test executes.

## Solution

Point the download at a source that still exists and can't drift:

- `lvyufeng/HunyuanOCR` mirrors the release on the HF Hub; all four safetensors shards are sha256-identical to Tencent's official ModelScope release (`Tencent-Hunyuan/HunyuanOCR`), which remains publicly available.
- Pin the mirror to a fixed revision in the workflow matrix and scope the weight cache key to that revision, so neither mirror drift nor a stale cached snapshot can change what CI tests.

Longer term the weights could be re-hosted under an org-controlled HF repo; that needs an HF org token with write access, which CI currently doesn't have.

### LINKED ISSUE #1217: bug: docs claim Candle Hunyuan-OCR auto-downloads its model, but the backend requires model_path
## Problem

`docs/guides/ocr.md` (Candle Hunyuan-OCR section) says "The model downloads automatically on first use (~3.5 GB) and is cached at ~/.cache/huggingface/". The backend does the opposite: `HunyuanOcrBackend` errors unless `backend_options.model_path` points at a local model directory — there is no auto-download path (unlike TrOCR/GLM-OCR, which do fetch from the Hub).

This got worse on 2026-07-06: `tencent/HunyuanOCR` was pulled from the HF Hub (#1216), so a reader following the docs has neither an auto-download nor an obvious place to fetch weights from.

## Solution

Either document the real contract (local `model_path` required, plus where to obtain the weights — Tencent's official ModelScope release, or a checksum-verified HF mirror), or implement the auto-download the docs promise, mirroring the TrOCR/GLM-OCR loader path. Not addressed by the CI fix PR; needs its own change.

### LINKED ISSUE #1224: bug: layout-model download lock race fails GPU ORT tests when two tests fetch the same model
## Problem

In GPU CI run https://github.com/xberg-io/xberg/actions/runs/28815855237, `layout_detection_cuda::rtdetr_table_detection` failed with:

```
Failed to create layout engine: ModelDownload("Failed to download 'rtdetr/model.onnx' from xberg-io/layout-models:
Lock acquisition failed: ~/.cache/huggingface/hub/models--xberg-io--layout-models/blobs/<sha>.lock")
```

`rtdetr_table_detection` and `rtdetr_complex_document` run concurrently in the same test binary and both try to download `rtdetr/model.onnx` on a cold cache. One acquires the HF blob lock, the other errors out instead of waiting — the sibling test passed with the same model 30 seconds later. Cold-cache GPU runs flip a coin on this race.

## Solution

Make the layout-model download path wait on (or retry after) lock contention instead of failing — a bounded retry around the lock acquisition, or serializing first-download via a process-wide mutex keyed on repo+file. Alternatively pre-fetch layout models once before the test threads start.

tobocop2:
## Problem

The GPU OCR suite had two red jobs (paddleocr-vl and hunyuan-ocr, e.g. [this run](https://github.com/xberg-io/xberg/actions/runs/28776698885)), with three distinct causes:

**PaddleOCR-VL crashed on every real page.** The multimodal rope index built its height positions with a transposed shape that only broadcasts for square patch grids, so a normal page (an 8 × 26 grid after spatial merge) died in the forward pass with `cannot broadcast [1, 8] to [8, 26]`. Fixing that exposed two more problems that had been unreachable behind the crash. The generation loop had never actually decoded a token: it read the argmax at the wrong tensor rank, re-fed the whole sequence every step while the attention KV cache kept appending duplicate keys, and the branch that continues rope positions across decode steps had been lost during vendoring, so every step failed with a dtype mismatch. And once tokens flowed, the output was garbage — the decoder ran its multi-token prefill without a causal mask, so prompt tokens attended bidirectionally, the KV cache was built from contaminated hidden states, and generation collapsed into repeating a single token.

**Hunyuan-OCR refused to load the released checkpoint.** Newer revisions of its `config.json` no longer include `tie_word_embeddings`. Our parser treated the missing field as fatal, even though the loader never uses the value — the output head is always tied to the input embeddings, and transformers defaults the field to `true`.

**The Hunyuan weights vanished mid-day.** Tencent pulled `tencent/HunyuanOCR` from the Hugging Face Hub on 2026-07-06 — the morning CI run still downloaded it unauthenticated; by afternoon it returns 401 even with a token. Since the multi-GB weight caches evict immediately (the repo is far over the 10 GB actions-cache budget), every run re-downloads, so the hunyuan job would now fail at the download step no matter what the code does.

**Hunyuan-OCR required a manual `model_path` despite docs promising auto-download.** The docs said the weights download on first use, but the backend errored unless `backend_options.model_path` pointed at a pre-staged directory — and with `tencent/HunyuanOCR` pulled from the Hub, there was nowhere obvious to stage it from.

Fixes #1214, fixes #1215, fixes #1216, fixes #1217, fixes #1224.

## Solution

- **Rope index**: build the vision block as three stacked rows (t, h, w), matching the Qwen2-VL reference — t constant for the single image frame, h advancing per grid row, w per grid column. Works for any grid shape.
- **Generation loop**: prefill the prompt once, then feed each new token through the cached decode path at its absolute position — the same loop the Hunyuan engine already uses. The restored position-continuation branch makes those steps work, and the loop now returns only newly generated tokens, so prompt text can never leak into OCR output.
- **Causal mask**: the prefill now applies the standard additive causal mask, like the Hunyuan and DeepSeek decoders (single-token decode steps attend over the whole cache, which is exactly the visible past).
- **Honest tests**: the paddle and hunyuan GPU e2e tests now assert the fixture text is actually read — "produces text" and "reads the page" are different claims, and degenerate output can no longer pass as green.
- **Hunyuan config**: `tie_word_embeddings` defaults to `true` when absent; an explicit value is still honored.
- **Runtime auto-download**: the Hunyuan-OCR backend now fetches its weights on first use from Tencent's official ModelScope release and caches them under the xberg cache directory — `model_path` becomes optional (an explicit path still wins for offline/custom weights). The file list and checksums live in one checked-in manifest (`crates/xberg/src/candle_ocr/hunyuan-ocr.sha256`) that both the runtime (via `include_str!`) and CI (`sha256sum --check`) read — single source of truth, so updating the model is a one-file edit. Trust attaches to the manifest, not the host: a changed or tampered upstream file fails staging instead of feeding wrong weights into inference. This is the existing GLiNER checksum-manifest pattern; the hardcoded-checksum shape is already used by the layout, paddle, and doc-orientation model managers.
- **Shared manifest parser**: the Candle weight stager and the GLiNER checksum loader had near-identical parsers for the `sha256sum` manifest format; extracted one `model_download::parse_sha256_manifest` that both now use (behavior unchanged).
- **CI weights**: the hunyuan GPU job stages from the same manifest and same ModelScope source as the runtime, and its weight cache survives job failures so a red run doesn't re-download multi-GB shards.

## Validation

- **Unit tests**: 40 crate tests pass, including four new regression tests (non-square rope layout, decode-step position continuation, causal-mask semantics, config parse without the field). Each was verified to fail on the unfixed code with the exact error from CI.
- **CI**: [run 28830469560](https://github.com/xberg-io/xberg/actions/runs/28830469560) on the final HEAD is fully green — all 12 GPU jobs, under the strict content assertions. Both paddle and hunyuan read "Hello World" off the fixture (paddle was emitting 3875 chars of a single repeated token before the causal-mask fix; the old bare non-empty check passed it). Weights came through the checksum-verified ModelScope download.
- **Auto-download**: exercised end to end against the live ModelScope release (in-process `ureq` → checksum → atomic publish, plus the warm-cache no-op path); the four config files' hashes match the manifest exactly. Inference from the staged directory is identical to the already-GPU-verified `model_path` path.
- **GPU pod** (RunPod A40, CUDA 12.6, rust 1.95): `main` reproduces the paddle crash verbatim against the real released checkpoints, and hunyuan reads "Hello World" end to end with the official weights.

The ONNX/ORT GPU job also failed once during verification on a cold-cache model-download lock race (two GPU tests fetching the same layout model at once, racing hf-hub's blob lock). Fixed here by serializing same-file downloads above hf-hub — see #1224.


tobocop2:
@Goldziher 
Screenshot 2026-07-06 at 8 52 42 PM

💪

tobocop2:
@Goldziher — one intentional inconsistency to flag. **This can merge as-is; it just needs a follow-up from you.**

**What**
Hunyuan-OCR weights are staged from Tencent's official ModelScope repo (`Tencent-Hunyuan/HunyuanOCR`), not Hugging Face — the only non-HF download in the tree.

**Why**
`tencent/HunyuanOCR` was pulled from the HF Hub (401 even with a token; our `HF_TOKEN` is empty), so hf-hub can't reach it. ModelScope is Tencent's own release and works today with no publishing on our side. Intentional stopgap.

**Is it safe**
Yes — every file is verified against a committed sha256 manifest (`crates/xberg/src/candle_ocr/hunyuan-ocr.sha256`); a changed or tampered file fails staging. Runtime and CI both read that manifest.

**Follow-up (for you)**

- **Host the weights** — create `xberg-io/hunyuan-ocr` with these 8 files (hashes already in the manifest):
    - `config.json`
    - `generation_config.json`
    - `preprocessor_config.json`
    - `tokenizer.json`
    - `model-00001-of-00004.safetensors`
    - `model-00002-of-00004.safetensors`
    - `model-00003-of-00004.safetensors`
    - `model-00004-of-00004.safetensors`
- **`.github/workflows/ci-gpu.yaml`** — in the `hunyuan-ocr` matrix entry, swap `ms_repo: Tencent-Hunyuan/HunyuanOCR` → `hf_repo: xberg-io/hunyuan-ocr`.
- **`crates/xberg/src/candle_ocr/model_stager.rs`** — fetch via `hf_download("xberg-io/hunyuan-ocr", name)` + `verify_sha256`; delete the ModelScope helpers (`modelscope_url` / `download_to` / `try_download_to`).
- **`crates/xberg/src/candle_ocr/hunyuan-ocr.sha256`** — no change (same bytes, same hashes).

PR #1219, opened 2026-07-06
fix: extraction-engine audit fixes (#1223) — redaction, tables, encoding, config

### LINKED ISSUE #1212: bug: Multiple markdown output bugs in docx conversion
### Description

I'm using xberg to convert `.docx` files to markdown for use as LLM context. The issues below cause extra tokens, semantic loss, and reduced retrieval accuracy.

---

## Bug 1: Bold/italic escaped in tables

Bold and italic inside table cells and headers get backslash-escaped causing extra tokens if output is being passed to an LLM.

**Actual:**
```markdown
| \*\*Name\*\* | \*\*Status\*\* |
| \*\*REQ-001\*\* | \*See appendix\* |
```

**Expected:**
```markdown
| **Name** | **Status** |
| **REQ-001** | *See appendix* |
```

Body text bold/italic works fine outside of tables, this only happens inside tables.

---

## Bug 2: Lists flattened

Bullet and numbered lists lose their markers and line breaks.

**Actual:**
```
Connect the power cable Wait for the LED to turn green Open the browser
```

**Expected:**
```
- Connect the power cable
- Wait for the LED to turn green
- Open the browser
```

---

## Bug 3: Ampersands stripped

The `&` character is silently removed.

**Actual:**
```
Terms Conditions
Research Development
```

**Expected:**
```
Terms & Conditions
Research & Development
```

---

### Steps to reproduce

Use the attached .docx file to reproduce the issue. It contains examples covering all three reported bugs. [xberg-test-issues.docx](https://github.com/user-attachments/files/29692355/xberg-test-issues.docx)

```text

```

tobocop2:
Implements fixes for the extraction-engine audit in #1223. Landing as a sequence of focused, individually-tested commits on one branch, each with a failing-test-first reproduction and (for table/extraction changes) a corpus A/B check against `main`.

## Landed

- **Stroke-width-encoded table rules are detected.** Print-era generators draw a table's vertical rules as a tiny stroked segment whose line width equals the table height — geometrically a speck, but it renders as a full-height bar. Detection only looked at path geometry, so these tables had no detectable vertical rulings and their text poured out column by column. Detection now accounts for stroke-rendered geometry when a path has no usable geometric extent of its own. Byte-identical on all 310 corpus PDFs against `main`; an earlier broader version that disturbed a pdfplumber fixture was caught by the sweep and turned into a regression test.
- **Redaction masks every output field, not just `content`.** Redaction rewrote ~7 fields and left ~12 serialized text fields (page content, table cells, elements, URIs, form-field values, image captions, metadata, `structured_output`, …) carrying the original PII while the report claimed success. An exhaustive field walk now masks every text surface of `ExtractedDocument`, recursing into image OCR sub-documents and `structured_output`/`code_intelligence` JSON. Covered by a test that plants the same email in a table cell, page, URI, form field, metadata, and structured output and asserts none survive.
- **Language detection honors `min_confidence` and is deterministic.** Multi-language mode capped the caller's threshold at 0.35 (so a request for high-confidence detection still admitted 0.35 chunks) and ordered equal-frequency languages by HashMap iteration, so the primary language flipped between runs. Both fixed.

## In progress

The remaining root-cause fixes from #1223 land as further commits here: one span-aware grid→rows helper for merged cells across HTML/DOCX/PPTX/XLSX (also closes #1212's DOCX table bugs), a central charset-detection entry point, a pipeline rule against content mutation after chunking, `PdfConfig.passwords` plumbing, effective-config cache keys, and error-not-silence for OCR/archive paths. Findings that need runtime fixtures (OCR backends, embedding models, binary Office docs) get committed fixtures + tests in the same PR.

The running-footer suppression from the first draft is held: the corpus sweep showed its repetition key collides on digit-varying repeated paragraphs (a federal-register fixture), so it is being rebuilt with an exact-match + margin-band gate before re-landing.

## Upstream (pdf_oxide)

Three findings root-cause to the pdf_oxide backend and are filed there; this PR works around them where possible and will adopt the upstream fixes when released:
- yfedoseev/pdf_oxide#811 — words split inside single string literals ("module" → "m odu le").
- yfedoseev/pdf_oxide#812 — path bounding boxes ignore stroke width (the root cause of the stroke-width table miss above).
- yfedoseev/pdf_oxide#813 — rotated pages extract in portrait order with no rotation metadata on words.

Each change ships with unit and full-pipeline integration tests plus, for extraction changes, the corpus A/B check that already caught two bad iterations.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -49,6 +49,15 @@ The changelog starts fresh at `1.0.0-rc.1`. For the Kreuzberg v1–v4 history, s
   width — scanned pages with vertical OCR layers, typeset tategaki books. The panic guard kept
   extraction alive, but the affected page came back as a per-page error with its text lost.
   pdf_oxide 0.3.73 fixes the sort, so those pages now extract normally.
+- **Bordered tables with stroke-width-rendered rules are detected (#1213).** Some
+  print-era PDF generators draw a vertical table rule as a ~1pt segment stroked with a line
+  width equal to the table height, so the rule's geometric bounding box is a speck and the
+  Lines-strategy detector saw no vertical rulings — whole fuse-chart-style tables were missed
+  (their only detected "table" being a false-positive page footer) and their text flowed out
+  column-major, destroying row associations. Table detection now expands such stroke-width-encoded
+  rules to their rendered geometry before filtering, so these grids are detected with rows
+  intact. Expansion applies only to paths with no usable geometric extent of their own, so
+  documents that don't use this drawing trick extract identically.
 
 ## [1.0.0-rc.1] - 2026-06-26
 
--- a/crates/xberg/src/pdf/oxide/table.rs
+++ b/crates/xberg/src/pdf/oxide/table.rs
@@ -56,11 +56,14 @@ const PRIMITIVE_GEOMETRY_FLOOR: f32 = 5.0;
 /// 0×507pt/7.2pt column rules in pdfplumber's la-precinct fixture — is
 /// already a valid primitive and must keep its geometric bbox).
 fn expand_hairline_stroke(path: &mut pdf_oxide::elements::PathContent) {
-    if path.stroke_color.is_none() || path.stroke_width <= PRIMITIVE_GEOMETRY_FLOOR {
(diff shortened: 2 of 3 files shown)


tobocop2:
Superseded by the clean xberg-native fixes PR on fix/extraction-audit-xberg (stroke workaround removed; it duplicated pdf_oxide internals). See the new draft PRs.

Issue #1223, opened 2026-07-06
bug: extraction-engine audit — 20+ verified defects (table alignment, encoding, OCR, config) and a redaction PII leak across ~12 output fields


tobocop2:
## Description

This started with one document — a car owner's manual (2011 Ford Crown Victoria) whose fuse-chart tables extracted with their rows destroyed (#1213). That turned out to be a pdf_oxide-level defect, but investigating it prompted a **systematic audit of the whole extraction engine** to find out what else was wrong. This issue is the result of that audit.

10 review rounds covering every subsystem (PDF text/reading order, table detection, OCR, office formats, HTML/email/archives, chunking, output composition, caching, MIME routing, config plumbing, FFI/API), each round independently re-verified, and the highest-severity findings reproduced at runtime. The handful of table defects that started it turned out to be instances of engine-wide classes; a mechanical field-by-field enumeration of `ExtractedDocument` then showed the redaction PII leak spans ~12 output fields, not the two originally found.

**Scope:** this issue tracks the **xberg-side** defects, which are fixable in this repo and land as one PR. The original trigger — bordered stroke-width-ruled tables (the fuse charts) — is a pdf_oxide defect and is tracked separately in **#1213**, blocked on **yfedoseev/pdf_oxide#814**; it is not fixed here.

**Evidence tags**
- `[E]` — reproduced at runtime with a minimal input. Actual output shown; the runnable test is in the reproduction harness (see *Steps to reproduce*).
- `[R]` — confirmed by two independent code reads (one reviewer found it, a second tried and failed to refute it); not yet runtime-driven, usually because it needs an OCR runtime, an embedding model, or a binary Office fixture.

Runtime testing also **removed** two findings that passed code review (a claimed HTML→XML misroute — both probes detect `text/html`; and colspan header misalignment — simple colspan headers align, only rowspan misaligns), which is why the tags matter.

## Steps to reproduce

Reproduction harness (11 defects driven at runtime as `#[ignore]` integration tests) and the reference PDF are in this gist:
- https://gist.github.com/tobocop2/0d30905e2375e09b153ce50296316f91 — `audit_reproductions.rs`, `cv-manual.pdf`

Each `[E]` finding below states the exact input and the actual output observed.

## Relevant files and configuration

Default `ExtractionConfig` unless a finding names a config field. All code locations are `crates/xberg/src/...`.

---

# Security

## 1. Redaction leaves most output fields unredacted — PII leak `[E]`

[...]

Issue #1213, opened 2026-07-06
bug: bordered stroke-width-ruled tables not detected (fuse charts) — blocked on pdf_oxide #814


tobocop2:
## What happens

Fully ruled three-column tables — the fuse charts in a 2011 Ford Crown Victoria owner's manual (pages 194–200) — are not detected as tables, and their text extracts **column-by-column** (all locations, then all ratings, then all circuit descriptions), so every row association is destroyed. A question like "what amp rating is fuse 101?" can't be answered from the extracted text.

Reproducer: https://gist.github.com/tobocop2/0d30905e2375e09b153ce50296316f91 (`cv-manual.pdf`), page 198 is the clearest single page.

## Root cause: pdf_oxide path bounding boxes ignore stroke width

These tables draw each vertical rule as a ~1pt horizontal segment stroked with a line width equal to the table height:

```
430.05 w
q 1 0 0 1 236.154 402.163 cm
0 0 m  .998 0 l  S
```

Rendered, that's a full-height vertical bar. But `pdf_oxide::extract_paths` returns its bounding box as geometry only — `bbox = (236.2, 402.2, w=1.0, h=0.0)`, `stroke_width = 430.05` — so it fails `is_table_primitive`, the Lines-strategy detector sees a grid with no vertical rulings, and the whole table is missed.

This is a pdf_oxide-level defect: **yfedoseev/pdf_oxide#812** (path bboxes ignore stroke width), fixed in **yfedoseev/pdf_oxide#814**. Once that lands and xberg bumps the dependency, `extract_tables_with_config` detects these rules natively with no xberg-side workaround.

## Status: blocked on pdf_oxide #814

A stroke-aware detection workaround was prototyped (expanding stroke-width-encoded rule bboxes to their rendered geometry before the primitive filter), but it reimplements pdf_oxide's table pipeline inside xberg, which duplicates upstream internals and would drift as pdf_oxide evolves. That workaround is kept on a dedicated branch and will **not** be merged; the fix belongs upstream.

This issue is intentionally scoped to the pdf_oxide-blocked defect. The broader extraction-engine audit that this document triggered — redaction PII leaks, encoding handling, merged-cell alignment, and the rest — is tracked separately in #1223 and lands independently.

## Related upstream

- yfedoseev/pdf_oxide#812 — path bboxes ignore stroke width (root cause; PR #814).
[...]

PR #1225, opened 2026-07-06
fix: extraction-engine audit (#1223)

### LINKED ISSUE #8: Resolve CI Test Failures
See the following PR: https://github.com/Goldziher/kreuzberg/pull/10

The pipeline is failing. I would love to have some help addressing the test failures since I have a mac and I cant debug this easily. 

Any takers?

### LINKED ISSUE #1212: bug: Multiple markdown output bugs in docx conversion

tobocop2:
Implements the xberg-side fixes from the extraction-engine audit in #1223. Each fix is an individually-tested commit; nothing here touches pdf_oxide.

The audit started from a car owner's manual whose fuse-chart tables extracted with rows destroyed — that specific defect is a pdf_oxide issue tracked in #1213 (blocked on yfedoseev/pdf_oxide#814) and is **not** in this PR. Investigating it surfaced engine-wide classes; this PR fixes the ones that live in this repo.

## Fixes (14)

| # | Area | What was broken | Severity |
|---|------|-----------------|----------|
| 1 | Redaction | Rewrote ~7 fields, left ~12 (page content, table cells, elements, URIs, form-field values, image captions, metadata, `structured_output`, …) carrying raw PII while the report claimed success | High (security) |
| 2 | Merged tables | A cell under a `rowspan` shifted left into the spanning column in HTML (4 flatteners) and DOCX; merged headers misaligned every following row. Also closes #1212's DOCX bug. XLSX and PPTX were flagged on the same suspicion but live extraction confirms they already align correctly — regression guards added rather than a fix | High |
| 3 | Table heuristics | Dense unruled reference tables (>20 rows, ≤3 cols) were rejected by the density guard on row-count alone; now kept when their cells are short values, while columned prose stays rejected. All-text rosters are left conservatively rejected — indistinguishable from scanned columned prose | Med |
| 4 | Encoding | Plain text / XML lossily decoded as UTF-8 (Latin-1 → U+FFFD); XML ignored its `encoding=` declaration; CSV kept Excel's UTF-8 BOM and made gb18030/big5 unreachable | High |
| 5 | OCR error handling | OCR failures returned already-failed native text as a clean success (silent empty); an empty OCR result wiped a mixed page's native text | High |
| 6 | PDF passwords | `PdfConfig.passwords` had zero read sites — encrypted PDFs returned a silent empty success even with the correct password | Med-High |
| 7 | Cache key | `source_name` and `ocr.tessdata_bytes` were `#[serde(skip)]`, so a config change that alters output served a stale cached result | Med-High |
| 8 | Embeddings | Tokenizer truncation hardcoded at 512 tokens with no config — chunks over 512 tokens embedded only their prefix. Added `EmbeddingConfig.max_sequence_length` | Med-High |
| 9 | MIME routing | Path entry trusted the extension, bytes entry sniffed magic — a misnamed file (DOCX named `report.pdf`) routed to different extractors | Med |
| 10 | Token reduction | Computed then discarded for Markdown/HTML output (only `content` was reduced, then replaced by `formatted_content`) | Med |
| 11 | Archives | Non-UTF-8 text members (zip/tar/7z) silently dropped by a failed `read_to_string` | Med |
| 12 | Language detection | `min_confidence` silently capped at 0.35; equal-frequency languages ordered nondeterministically | Med |
| 13 | Diff | Shape-changed tables produced an information-free empty diff instead of removed+added | Med |
| 14 | CSV | `NaN`/`inf`/`infinity` treated as numeric, flipping header and column-type detection | Low-Med |

Each change ships with unit and/or integration tests (synthetic HTML/DOCX/PDF fixtures, encrypted-PDF builder, encoding fixtures). Fixes #5 (OCR) and #8 (embeddings) were developed and runtime-verified on a pod against real backends and are integrated from the companion branch `fix/audit-ocr-embeddings`.

## Runtime validation

Beyond unit tests, the two model-dependent fixes were verified on a pod against real backends:

- **OCR error handling (#5):** with real tesseract — a scanned image-only PDF OCR'd correctly; an unregistered backend produced empty content **with a `ProcessingWarning`** (no longer a silent success); a mixed native+scanned PDF kept page 1's native text *and* OCR'd page 2. Includes unit tests for the merge (empty OCR result keeps native text; non-empty replaces it).
- **Embeddings truncation (#8):** with real `jina-embeddings-v2-small-en` (8192-ctx) — a >512-token chunk embeds identically to its 512-token prefix at the old cap (cosine 1.000000) but diverges once `max_sequence_length` is raised to 1024 (cosine 0.994623), proving the truncated tail now contributes to the vector.
- **Table heuristics (#3):** an A/B extraction against `main` over the full `test_documents` PDF corpus (310 files) recovers 4 real dense reference tables (assembly and Intel manual instruction tables) with zero tables lost and zero extraction status changes; synthetic fixtures prove a dense ledger is now detected while columned prose and an all-text roster stay rejected. A broader relaxation was tried and rejected: it recovered ~400 more tables but corpus inspection showed they were overwhelmingly false positives (bibliographies and scanned two-column prose), so the change was scoped to the density guard only.

## Note for the maintainer — binding regen

Fix #8 adds a public config field (`EmbeddingConfig.max_sequence_length`). This PR is source-only, so the generated language bindings are **not** regenerated here. Several bindings (Python, PHP, dart, elixir, ruby) construct `xberg::EmbeddingConfig` exhaustively (`Self { ...all fields }`), so until they are regenerated their build/E2E jobs fail on `missing field max_sequence_length`. Regenerating the bindings picks up the new field and resolves those jobs. (node builds from a default-and-override and already compiles; wasm doesn't build the ORT-dependent embedding path at all.)

Two CI reds are pre-existing on `main`, unrelated to this PR: `pdf::structure::classify::tests::test_is_math_character` (the function classifies `'+'` as math but the test asserts it isn't) and the `is_missing_default_backend`/`ensure_defaults` dead-code under the reranker feature set (from `c2090c6704`).

## Not in this PR

- Bordered stroke-width-ruled tables (the fuse charts) — **#1213**, blocked on pdf_oxide#814.
- Word-fragmentation in cells (pdf_oxide#811) and rotated-page tables (pdf_oxide#813).


Goldziher:
Wow, amazing 🤩

PR #1226, opened 2026-07-06
fix(pdf): stroke-width bordered-table detection — BLOCKED on pdf_oxide#814 (#1213)

### LINKED ISSUE #1213: bug: bordered stroke-width-ruled tables not detected (fuse charts) — blocked on pdf_oxide #814

tobocop2:
Fixes #1213 — **blocked on yfedoseev/pdf_oxide#814**. Do not merge until that lands.

## What this is

Bordered three-column tables whose vertical rules are drawn as ~1pt segments stroked with a table-height line width (the fuse charts in a car owner's manual) are not detected, and their text extracts column-by-column with rows destroyed. The root cause is that pdf_oxide's path bounding boxes ignore stroke width, so each rule is a 1×0pt speck that the Lines-strategy detector can't see.

## Why it's blocked, not merged

This branch carries a workaround that expands stroke-width-encoded rule bboxes to their rendered geometry before the primitive filter. It works (the regression tests pass), but it **reimplements part of pdf_oxide's table pipeline inside xberg** — it duplicates upstream internals and would drift as pdf_oxide evolves.

The correct fix is upstream: **yfedoseev/pdf_oxide#812** (path bboxes should account for stroke width), fixed in **PR #814**. Once that lands and this repo bumps the dependency, `extract_tables_with_config` detects these rules natively and this whole workaround reverts to a dependency bump.

This PR exists to (a) capture the bug and its regression tests as history, and (b) hold the interim workaround in one reviewable place. **It should not merge to main** — when #814 lands, the reviewer swaps the workaround for the dep bump.

## Related

- Root cause: yfedoseev/pdf_oxide#812 (PR #814).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -49,6 +49,15 @@ The changelog starts fresh at `1.0.0-rc.1`. For the Kreuzberg v1–v4 history, s
   width — scanned pages with vertical OCR layers, typeset tategaki books. The panic guard kept
   extraction alive, but the affected page came back as a per-page error with its text lost.
   pdf_oxide 0.3.73 fixes the sort, so those pages now extract normally.
+- **Bordered tables with stroke-width-rendered rules are detected (#1213).** Some
+  print-era PDF generators draw a vertical table rule as a ~1pt segment stroked with a line
+  width equal to the table height, so the rule's geometric bounding box is a speck and the
+  Lines-strategy detector saw no vertical rulings — whole fuse-chart-style tables were missed
+  (their only detected "table" being a false-positive page footer) and their text flowed out
+  column-major, destroying row associations. Table detection now expands such stroke-width-encoded
+  rules to their rendered geometry before filtering, so these grids are detected with rows
+  intact. Expansion applies only to paths with no usable geometric extent of their own, so
+  documents that don't use this drawing trick extract identically.
 
 ## [1.0.0-rc.1] - 2026-06-26
 
--- a/crates/xberg/src/pdf/oxide/table.rs
+++ b/crates/xberg/src/pdf/oxide/table.rs
@@ -38,6 +38,13 @@ const PRIMITIVE_GEOMETRY_FLOOR: f32 = 5.0;
 
 /// Expand a stroke-width-encoded rule's bounding box to its rendered geometry.
 ///
+/// INTERIM WORKAROUND — remove once yfedoseev/pdf_oxide#812 lands (PR #814).
+/// This reimplements part of pdf_oxide's table pipeline to inject a bbox
+/// expansion; once pdf_oxide's path bounding boxes account for stroke width,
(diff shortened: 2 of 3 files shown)


tobocop2:
Superseded by #1252. pdf_oxide 0.3.74 lands the upstream stroke-width bounding-box fix this branch was blocked on, so the fix reduces to a dependency bump plus the regression test — no in-tree workaround. #1252 carries that on an upstream branch so CI runs on it.

Issue #1235, opened 2026-07-09
bug: test_is_math_character fails on main — a stale assertion contradicts the classifier


tobocop2:
## Problem

`main` is red:

```
cargo test -p xberg --features pdf --lib pdf::structure::classify::tests::test_is_math_character

thread '...' panicked at crates/xberg/src/pdf/structure/classify.rs:2072:9:
assertion failed: !is_math_character('+')
```

`is_math_character` (`classify.rs:1230`) counts `'+'`, `'='` and `'^'` as math characters, while `test_is_math_character` asserts that `'+'` is not one.

The implementation is the correct side. Two other tests depend on it:

- `test_formula_detection_ascii_operators` — `a2 + 8 = 12` must classify as a formula
- `test_formula_detection_superscript_exponent` — `a² + 8 = 12` must classify as a formula

Remove the ASCII operators from `is_math_character` and both of those fail, because the paragraph then contains zero math characters. So the single `assert!(!is_math_character('+'))` line is a leftover from before ASCII operators were added, and was never updated.

## Solution

Delete the stale assertion. Optionally assert the real contract instead — that `'+'`, `'='` and `'^'` *are* math characters — so the two formula-detection tests cannot silently lose their premise.

## Follow-up, not this bug

Counting bare ASCII operators does produce false positives. The only consumer (`classify.rs:133`) marks a paragraph of at most 30 words as a formula once it holds three or more math characters:

```rust
let math_char_count = para_text.chars().filter(|c| is_math_character(*c)).count();
if total_chars > 0 && (math_char_count >= 3 || (math_char_count as f64 / total_chars as f64) >= 0.15) {
    para.is_formula = true;
```

`C++ and C++17 are supported` contains four `'+'` characters and is therefore classified as a formula, losing its heading level. Fixing that needs a smarter rule than a flat character count — for instance not counting repeated identical operators, or requiring an operator to sit between operands. Worth a separate issue rather than folding into the red-test fix.

PR #1236, opened 2026-07-09
feat(pdf): detect scanned PDFs and OCR only their pages (#1229)

### LINKED ISSUE #1229: feat: Add option to detect scanned PDFs with textlayers and run ocr conditionally

Add some sort of flag/guess score for PDFs indicating if they are result of a scan.

Perhabs it could be done based on some factors like fullscreen images being present on all pages in combination with the already present quality score for the textlayer.

I am dealing with some scanned pdfs where the scanner includes a hiddden, poor quality textlayer.
The issue i am facing with those, is that i have no way to tell beforehand.

`force_ocr=true` produces much better ouput but on clean pdfs results are obviosuly worse.

It would prevent the need for additional pdf analytics steps outside.
Xberg already has all metdata, etc. of the pdf file.
`force_ocr=true` seems all or nothing

### LINKED ISSUE #1234: bug: cargo test --features pdf,ocr fails to compile because the auto-rotate tests aren't feature-gated
## Problem

`rotate_rgb_image_data` in `crates/xberg/src/ocr/processor/execution.rs` is defined under `#[cfg(feature = "auto-rotate")]` (line 116), but the seven unit tests that exercise it are not gated. Building the test target with the `ocr` feature but without `auto-rotate` fails to compile.

Reproduce on a clean checkout of main (191863eba0):

```
cargo clippy -p xberg --features pdf,ocr --lib --tests -- -D warnings
```

```
error[E0425]: cannot find function `rotate_rgb_image_data` in this scope
    --> crates/xberg/src/ocr/processor/execution.rs:1434:27
```

Seven occurrences, at lines 1434, 1443, 1452, 1460, 1468, 1469 and 1478.

## Solution

Gate the affected test functions on `feature = "auto-rotate"`, matching the item under test. The tests begin at `test_rotate_rgb_image_data_identity` (line 1431).

Found while working on scanned-PDF detection, where `--features pdf,ocr` is the natural combination to lint.

tobocop2:
Closes #1229. Also closes #1234 and #1235, two reds on `main` that this branch had to get past.

Consumer scanners embed an invisible OCR text layer under the page image. Its words are well-formed but wrong — `Tbe` for `The`, `rn` for `m`, `$l2,45O.OO` for `$12,450.00`. Every quality check xberg applied to a native PDF text layer was lexical (replacement characters, alphanumeric ratio, word fragmentation, consecutive repeats), so that text passed all of them and OCR never ran. The reporter's only escape was `force_ocr`, which is all-or-nothing and makes clean PDFs worse.

Nothing in the pipeline asked the question that settles it: is this page a picture of a document? This PR adds that question.

## What it does

`PdfMetadata` gains `scanned_confidence` (0.0–1.0) and `scanned_pages`, so a caller can tell whether a document is a scan *before* deciding how to extract it. That was the literal ask in the issue: "I have no way to tell beforehand."

`ExtractionConfig` gains `ocr_strategy`. The default `auto` is exactly today's behaviour. `scanned_pages { min_confidence }` additionally OCRs pages that look like scans and leaves every other page on native text. On the CLI: `--ocr-scanned-pages [--scanned-min-confidence 0.7]`.

A page is graded on how much of it is covered by raster images, whether its text layer is invisible or absent, its image codec (CCITT and JBIG2 are bilevel fax formats, never emitted by authoring tools), and the document producer. The two failure modes a naive rule hits are both avoided:

- **Coverage alone is not enough.** A slide with a full-bleed background image is entirely raster, but its text is drawn *visible* on top. It scores 0.50 — below the 0.70 default — and is never OCR'd.
- **Text quality alone is not enough.** A scanned page under a hidden text layer has lexically perfect text. It scores 0.85, because the text is invisible and the page is a raster.

Detection is advisory and infallible: a page that cannot be read contributes no evidence rather than failing the extraction, and the classification call runs under the existing panic guard.

Cost is two-stage, so it tracks how image-like a document is. A prefilter reads image placements without decoding pixels (~130 µs/page); only image-dominated pages pay for content-stream inspection. A 455-page text book inspects one page rather than all 455 — 59 ms instead of 3.6 s.

## Does it work?

On the reporter's case — a scanned page carrying a scanner's hidden text layer — extracted twice with a real tesseract backend and scored against ground truth:

| | extraction method | character error rate | word error rate |
|---|---|---:|---:|
| **before** (`ocr_strategy = auto`, today's default) | `native` | 10.6% | **48.4%** |
| **after** (`ocr_strategy = scanned_pages`) | `mixed` | **0.0%** | **0.0%** |

The page is detected as a scan with confidence 0.85. Ground truth is the text rendered into the page image when the fixture was built — authored input, never an extractor's own output.

Worth noting that the PDF backend's own auto-routing does not catch this either. Asked to route text-vs-OCR on the same file, it returns an empty `pages_needing_ocr`, classifies both pages as a high-confidence text layer, and emits the same corrupt sidecar — `CER 10.6%`, `WER 48.4%`, identical to the `auto` row above. It keeps a lexically clean sidecar on purpose. So this is not a bug that can be fixed one layer down; the detection has to ask a question that layer never asks.

Before, the native text layer is returned verbatim:

```
Tbe quarterly financiaI report sbows revenue
of $l2,45O.OO for tbe period ending Marcb 3l.
Operating expenses rernained flat conipared
```

After, the page is OCR'd instead:

```
The quarterly financial report shows revenue
of $12,450.00 for the period ending March 31.
Operating expenses remained flat compared
```

Every word in the "before" text is well-formed — no replacement characters, no fragmentation, plausible word lengths. That is exactly why every lexical quality check accepted it and OCR never ran. Nearly half the words are wrong.

## Does it break anything?

A per-page differential sweep over a 281-document corpus (arXiv, the pdf.js test suite, IRS forms, multilingual, PoC‖GTFO), run on a pod: **6,269 pages**, each extracted independently by a probe built from `main` and a probe built from this branch. Harness: https://gist.github.com/tobocop2/36b19e45dd339dda3f5f41bbdbdb6fc8

Per-page rather than per-document on purpose: a document's `scanned_confidence` is a max over its pages, so one misclassified page inside a 30-page document is invisible in any aggregate.

```
pages=6269 scored=6266 missing_branch=0
detected_scans=6
fatal=0
```

- **The default path does not move.** Under `auto`, the SHA-256 of every page's extracted text is byte-identical between `main` and this branch — 6,269 of 6,269.
- **Every score is reproducible.** The harness independently reimplements the coverage formula and the scoring rule, never calling the shipped code, and asserts the shipped `scanned_confidence` matches to 1e-4. All 6,266 scored pages agree.
- **Invariants hold.** No page scores above threshold while carrying visible text; no image-dominated page with a hidden text layer scores below it. Zero violations.

Three pages fail to parse on both sides identically, so they are scored `NA` rather than silently counted as clean.

What detection actually finds in a born-digital corpus:

| | pages | share |
|---|---:|---:|
| detected as scans (confidence ≥ 0.70) | 6 | 0.1% |
| full-page raster, but visible text (0 < conf < 0.70) | 5 | 0.1% |
| not a scan (conf = 0) | 6,255 | 99.8% |

All 6 detected pages are the synthetic scan fixtures. Nothing in arXiv, pdf.js or the IRS forms was flagged. The middle row is the one that matters for clean PDFs: those pages are entirely covered by a raster image, so a coverage-only rule would OCR them — but their text is *visible*, drawn over the image, so they are born-digital and left alone.

## What detection does not do

It identifies that a scanner produced a text layer. It cannot judge whether that text is *accurate*. A page carrying a good hidden sidecar is indistinguishable from one carrying a bad sidecar — identical coverage, identical invisible-text ratio — and is OCR'd too. Deciding accuracy would need a dictionary, which is language-dependent and would regress every non-English corpus. The tests encode this rather than pretend otherwise.

This is why the strategy is opt-in and `auto` remains the default.

## CI is red here on purpose — #1239 fixes it

This PR is source-only, so it carries no generated files. Removing `HierarchyConfig.ocr_coverage_threshold` therefore orphans the accessors alef emits for it — one getter per core field — and four jobs fail with a single error:

```
error[E0609]: no field `ocr_coverage_threshold` on type `&xberg::HierarchyConfig`
```

- [Build FFI (aarch64-unknown-linux-gnu)](https://github.com/xberg-io/xberg/actions/runs/29054372257/job/86242331659)
- [Build FFI (x86_64-unknown-linux-gnu)](https://github.com/xberg-io/xberg/actions/runs/29054372257/job/86242331625)
- [Rust (macos-latest)](https://github.com/xberg-io/xberg/actions/runs/29054372218/job/86242375505)
- [Rust (ubuntu-24.04-arm)](https://github.com/xberg-io/xberg/actions/runs/29054372218/job/86242375504)

One root cause, four jobs. **#1239 regenerates the bindings on top of this branch and clears all four.** Merge it into `feat/scanned-pdf-detection` and this PR's CI re-runs green.

CI cannot report on #1239 itself: every workflow is declared `pull_request: branches: [main]`, so a PR into a feature branch runs only `validate`. The green run has to appear here, after it lands.

[`Live HF preset tests`](https://github.com/xberg-io/xberg/actions/runs/29054372218/job/86242375490) (`late_interaction_preset_live`, colbert query augmentation) is already red on `main` and untouched by this branch — no file here goes near embeddings.

## The bindings themselves are fine

Regeneration was run and checked, not assumed. With the pinned alef 0.34.5 (`alef all --clean`) plus `cbindgen` for the C header, which alef does not own:

- `cargo clippy --workspace --lib -- -D warnings` — passes
- `cargo check` on `xberg-ffi`, `xberg-py`, `xberg-node`, `xberg-php`, `xberg-wasm`, `xberg-jni` — all pass

**No alef change is needed.** `OcrStrategy` is a data-carrying enum, which was the thing worth checking; alef already handles that shape (`VlmFallbackPolicy` is the precedent) and emits idiomatic types everywhere, creating new files where a language needs one — `OcrStrategy.cs`, `ocr_strategy.ex`, `OcrStrategy.java`, `OcrStrategy.kt`. For example, C#:

```csharp
[JsonConverter(typeof(OcrStrategyJsonConverter))]
public abstract record OcrStrategy
{
    public sealed record Auto() : OcrStrategy;
    ...
}
```

The metadata fields land in each language's idiom:

```ts
readonly scannedConfidence?: number
readonly scannedPages?: Array<number>
```
```python
scanned_confidence: float | None
scanned_pages: list[int] | None
```
```c
float xberg_pdf_metadata_scanned_confidence(const XBERGPdfMetadata *ptr);
char *xberg_ocr_strategy_to_json(const XBERGOcrStrategy *ptr);
```

The *additions* alone break nothing: every binding builds `xberg::ExtractionConfig` and `xberg::pdf::metadata::PdfMetadata` from its mirror struct through `Default`, so the currently-committed generated code still compiles. It is only the *removal* that forces the regen.

## Also removed: a config field that did nothing

`HierarchyConfig.ocr_coverage_threshold` promised "trigger OCR if less than 50% of page has text", and the reference docs for all ten languages said so. Nothing ever read it. It was declared, defaulted, and round-tripped by its own unit test; that was the entire lifetime of the value. Its only real coverage, `tests/pdf_ocr_triggering.rs`, has been disabled behind `#![cfg(any())]` since the alef migration.

It described the feature this PR actually implements, so leaving it in place next to `ocr_strategy` would have shipped two knobs for one job, one of them fake. `HierarchyConfig` has no `deny_unknown_fields`, so config files that still set the key keep loading and the key is ignored.

## Two reds on main, fixed here

Neither is related to scanned PDFs; both were in the way. One commit each.

**#1235** — `test_is_math_character` asserts `'+'` is not a math character, but the implementation counts it, and two formula-detection tests depend on that (`a2 + 8 = 12` must be a formula). The assertion is stale; fixed to assert the real contract.

**#1234** — `cargo clippy -p xberg --features pdf,ocr --lib --tests -- -D warnings` fails on four `cfg` gaps that CI never builds: ungated `auto-rotate` tests, `with_rotation` reachable only from a `paddle-ocr` test, `detect_and_rotate` gated on the wrong feature, and a missing test import. Each is now gated on the feature that reaches it.

No runtime behaviour changes — every touched item is `cfg(test)` or dead in the failing cells. `cargo test -p xberg --features pdf --lib` goes from 1760 passed / 1 failed to 1762 passed / 0 failed.


tobocop2:
@Goldziher — flagging a design question. Your call.

**Some of this belongs in pdf_oxide, but not the part that fixes #1229 so I think this is good for merge**

I filed yfedoseev/pdf_oxide#840: `classify_page` calls born-digital pages `Scanned`. Over a 6,269-page corpus it flags 860 of them — 13.7% — including plain LaTeX text pages with no image at all. Anything routing `pages_needing_ocr` into OCR would wreck those.

But the reporter's case is the opposite failure, and it is deliberate upstream. Asked to auto-route the same file, pdf_oxide returns an empty `pages_needing_ocr`, calls both pages a high-confidence text layer, and emits the corrupt sidecar — `CER 10.6%`, `WER 48.4%`, exactly what we get today. `auto.rs:801-805` keeps a lexically clean sidecar on purpose. **A fully fixed pdf_oxide still would not fix #1229.**

**One piece of duplicated code.** `PageSignals` is only reachable via `classify_page()`, which also grades the page text (~15 ms/page). Too slow to run on everything, so `scan_detect` re-derives image coverage from `page_image_handles()` + `get_page_media_box()` at ~130 µs/page — a second copy of pdf_oxide's `image_area_ratio`. On the corpus the two agree everywhere it matters. I asked upstream for a cheap `page_signals()` accessor; if it lands, that code deletes itself.

**What stays here regardless:** `scanned_confidence`, `ocr_strategy`, the CLI flags, the OCR gate.

Not blocking. Happy to hold and rebase if you would rather wait for upstream.

tobocop2:
## CI: green once the bindings are regenerated (#1239)

The four red jobs here all fail with one error — `no field `ocr_coverage_threshold` on type `&xberg::HierarchyConfig`` — because this PR is source-only and the removal orphans the accessors alef emits per core field.

On the same tree with #1239's regen commit applied ([full run](https://github.com/xberg-io/xberg/actions/runs/29068994861), commit `897c6e261f`):

| job | here | with #1239 |
|---|---|---|
| Build FFI (aarch64-unknown-linux-gnu) | fail | **pass** |
| Build FFI (x86_64-unknown-linux-gnu) | fail | **pass** |
| Rust (macos-latest) | fail | **pass** |
| Rust (ubuntu-24.04-arm) | fail | **pass** |

The 14-language e2e matrix then runs: **13 of 14 pass** (csharp, elixir, go, java, kotlin-android, node, php, python, ruby, rust, swift, wasm, zig).

`dart` fails on `no such command: `expand`` — `cargo-expand` missing on the runner, [identical on `main`](https://github.com/xberg-io/xberg/actions/runs/29036375043/job/86190556345). `Live HF preset tests` is red on `main` too. Neither is from this branch.

## Two bugs the cross-language matrix caught

The 6,269-page corpus sweep could not see either, because it only exercises the Rust API. Both are fixed on this branch.

**`b784cfce63` — `ocr_strategy` rejected an explicit `null`.** Bindings serialize a zero-valued mirror struct with every field present, so Go sent `"ocr_strategy": null`. `#[serde(default)]` only covers a *missing* key, and an internally-tagged enum rejects `null`. Every Go batch call failed with `invalid type: null, expected internally tagged enum OcrStrategy`. Fixed with `deserialize_null_default`, which already existed for exactly this case, plus a test that reproduces the Go error.

**`de442b2d0d` — the `Default` impl hid the variant name.** It read `ocr_strategy: OcrStrategy::default()`. alef reads that literal to emit per-language defaults, so the generated Kotlin declared `val ocrStrategy: OcrStrategy` with no default and every `ExtractionConfig()` in its e2e suite stopped compiling. Now `OcrStrategy::Auto`, matching `vlm_fallback: VlmFallbackPolicy::Disabled` in `OcrConfig`.

Both come from copying `VlmFallbackPolicy`'s shape without its two cross-binding conventions: it sits behind an `Option<OcrConfig>` so it never receives a bare `null`, and its `Default` names the variant. alef behaved correctly in both cases.

PR #1239, opened 2026-07-10
chore(bindings): regenerate for scanned-PDF detection

tobocop2:
Stacked on #1236 — merge this into `feat/scanned-pdf-detection` first, and #1236 goes green.

## Why this exists

#1236 removes `HierarchyConfig.ocr_coverage_threshold`, a config field nothing ever read. alef emits one accessor per core field, so the committed bindings still call `xberg_hierarchy_config_ocr_coverage_threshold` on a field that no longer exists. That single removal is the only cause of every red job on #1236:

```
error[E0609]: no field `ocr_coverage_threshold` on type `&xberg::HierarchyConfig`
```

Build FFI (aarch64), Build FFI (x86_64), Rust (macos-latest) and Rust (ubuntu-24.04-arm) all fail on it, and nothing else does. After this commit they compile.

Verified locally: `cargo clippy --workspace --lib -- -D warnings` passes, as does `cargo check` on `xberg-ffi`, `xberg-py`, `xberg-node`, `xberg-php`, `xberg-wasm` and `xberg-jni`.

Note that CI cannot report on this PR itself: every workflow is declared `pull_request: branches: [main]`, so a PR into a feature branch only runs `validate`. The green run appears on #1236 once this lands.

## How it was generated

The pinned alef 0.34.5 (`alef all --clean`), plus `cbindgen` for `crates/xberg-ffi/include/xberg.h` and its three copied siblings, which alef does not own.

The new surface — `OcrStrategy`, `ExtractionConfig.ocr_strategy`, `PdfMetadata.scanned_confidence`, `scanned_pages` — lands in every language, including four new files (`OcrStrategy.cs`, `ocr_strategy.ex`, `OcrStrategy.java`, `OcrStrategy.kt`).

## Churn that is not from #1236

Regenerating an untouched `main` produces both of these, so they are pre-existing staleness rather than anything this branch did:

- **API reference version badges** move from `v1.0.0-rc.18` to `v1.0.0-rc.19`. The committed docs lag `Cargo.toml`.
- **alef content hashes** refresh across the generated tree.

## Deliberately not regenerated

Generated `Cargo.toml` manifests are reverted. Regeneration wants to downgrade `crawlberg` from 1.0.5 to 1.0.3 and delete the `macos-intel-target` blocks — both hand-maintained on `main`, and both regressions if committed. Worth deciding whether alef should stop owning those files, or whether its templates need updating.


tobocop2:
## CI status

CI cannot run on this PR directly — every workflow is declared `pull_request: branches: [main]`, so a PR into a feature branch only triggers `validate`.

So the same tree (`feat/scanned-pdf-detection` + this regen commit, `47da2f33de`) was pushed to a throwaway branch and run against `main` in #1240. That is byte-identical to what merging this PR produces.

The four jobs that fail on #1236 — all with `error[E0609]: no field `ocr_coverage_threshold` on type `&xberg::HierarchyConfig`` — all pass with the bindings regenerated:

| job | #1236 (source-only) | with this regen |
|---|---|---|
| [Build FFI (aarch64-unknown-linux-gnu)](https://github.com/xberg-io/xberg/actions/runs/29060801956/job/86262023421) | fail | **pass** (36m53s) |
| [Build FFI (x86_64-unknown-linux-gnu)](https://github.com/xberg-io/xberg/actions/runs/29060801956/job/86262023413) | fail | **pass** (43m5s) |
| [Rust (macos-latest)](https://github.com/xberg-io/xberg/actions/runs/29060802003/job/86262023533) | fail | **pass** (37m16s) |
| [Rust (ubuntu-24.04-arm)](https://github.com/xberg-io/xberg/actions/runs/29060802003/job/86262023564) | fail | **pass** (37m17s) |

`docs / Build` and both `validate` jobs also pass. Passing `Build FFI` unblocked the 14-language E2E matrix, which is running now against the regenerated bindings.

`Live HF preset tests` fails, and fails on `main` too (`late_interaction_preset_live`, colbert query augmentation). No file on this branch goes near embeddings.

Full run: https://github.com/xberg-io/xberg/actions/runs/29060801956

tobocop2:
## CI is green — and it found two real bugs on the way

CI cannot run on this PR directly (`pull_request: branches: [main]` means a PR into a feature branch only triggers `validate`), so the same tree was pushed to a throwaway branch and run against `main`. The check runs bind to the commit, so they show here too. Head: `897c6e261f`.

### The four jobs that fail on #1236

All fail there with one error — `error[E0609]: no field ocr_coverage_threshold on type &xberg::HierarchyConfig` — and all pass once the bindings are regenerated:

| job | #1236 | with this regen |
|---|---|---|
| Build FFI (aarch64-unknown-linux-gnu) | fail | **pass** (24m21s) |
| Build FFI (x86_64-unknown-linux-gnu) | fail | **pass** (37m15s) |
| Rust (macos-latest) | fail | **pass** (46m16s) |
| Rust (ubuntu-24.04-arm) | fail | **pass** (31m30s) |

### The 14-language e2e matrix

Passing `Build FFI` unblocks it. **13 of 14 pass.**

`csharp` · `elixir` · `go` · `java` · `kotlin-android` · `node` · `php` · `python` · `ruby` · `rust` · `swift` · `wasm` · `zig` — all green.

`dart` fails with `error: no such command: \`expand\`` — `cargo-expand` is not installed on the runner. It fails identically on `main` ([main](https://github.com/xberg-io/xberg/actions/runs/29036375043/job/86190556345) · [here](https://github.com/xberg-io/xberg/actions/runs/29068994861/job/86290600103)). Not from this branch.

`Live HF preset tests` (`late_interaction_preset_live`, colbert query augmentation) is also red on `main`. No file here goes near embeddings.

### Two bugs the e2e matrix caught that a Rust-only sweep could not

Both are in the parent branch, both are fixed, and both were invisible to the 6,269-page corpus sweep because it only ever exercised the Rust API.

**`ocr_strategy` rejected an explicit `null`** (`b784cfce63`). Bindings serialize a zero-valued mirror struct with every field present, so Go sent `"ocr_strategy": null`. `#[serde(default)]` only covers a *missing* key, and an internally-tagged enum rejects `null`, so every Go batch call died with `invalid type: null, expected internally tagged enum OcrStrategy`. `deserialize_null_default` already existed for exactly this — its doc-comment names the failure mode. A test now reproduces the Go error before the fix.

**The `Default` impl hid the variant name** (`de442b2d0d`). It read `ocr_strategy: OcrStrategy::default()`. alef reads that literal to emit each field's default in the target language, and `::default()` gives it nothing to name, so the generated Kotlin declared `val ocrStrategy: OcrStrategy` with no default and every `ExtractionConfig()` in the Kotlin e2e suite stopped compiling. Spelling it `OcrStrategy::Auto` — the way `OcrConfig` already does with `vlm_fallback: VlmFallbackPolicy::Disabled` — restores the generated default.

Worth noting alef was correct throughout. It generated exactly what the Rust described.

PR #1259, opened 2026-07-18
feat(xberg-wasm): OCR results include each line's coordinates, and JS apps can plug in their own OCR and NER

### LINKED ISSUE #1266: feat: let the browser supply its own entity-detection engine to the wasm engine

A browser app should be able to hand the wasm engine its own entity-detection function (for example one backed by a JS model), the same way it can already hand over its own OCR. This adds that, plus the one core trait change it requires.

#1259 lands `XbergEngine` with an injectable OCR backend. NER should work the same way: the host passes `ner: { ner(text, categories): Promise<Entity[]> }` and `XbergEngine.ner()` routes through it.

### Where it stands

The implementation already exists in #1253 (`bridge/ner.rs`, `JsNerBridge`). It was deliberately left out of #1259 because it cannot compile against the current trait.

### The blocker

`NerBackend` is declared `#[async_trait]` with `Send + Sync` supertraits. A JS bridge's futures hold `JsValue`s driven by `wasm_bindgen_futures::JsFuture`, which are not `Send`. The fork's answer was a per-target split:

```rust
#[cfg_attr(not(target_arch = "wasm32"), async_trait)]
#[cfg_attr(target_arch = "wasm32", async_trait(?Send))]
pub trait NerBackend: Send + Sync { ... }
```

### What review needs

Splitting a public trait's contract by target deserves its own scrutiny. Three things before merge: document the split on the trait, add a wasm32 compile check to CI so the `?Send` arm cannot rot, and audit every `Box<dyn NerBackend>` consumer for `Send` assumptions.

### Attribution

The bridge implementation and the trait split were written by @jamon8888 in #1253. This issue tracks landing them upstream as their own reviewable unit.

tobocop2:
Closes #1266. Design and implementation by @jamon8888 in #1253; his commits carry his authorship, review fixes are separate commits.

## What was broken

When a browser app ran OCR through the wasm package, it got back one undifferentiated blob of text. There was no way to know where each recognized line sits in the image, so the app could not highlight a line, scroll to it, or overlay it on the page. There was also no way for the app to plug in its own OCR or entity-detection engine: the wasm build either bundled one or you had nothing.

## What this adds

`XbergEngine`, constructed with a config and an injection object, exposing `extract()`, `ocr()`, and `ner()`. An injected OCR backend returns `{ text, lines: [{ text, confidence, bbox? }] }`; a backend without geometry degrades to an empty `lines` array. An injected NER backend works the same way, with a public `JsNerBridge` adapter for core consumers. All bridge calls run under a configurable timeout with leak-safe cleanup.

`NerBackend`'s `async_trait` futures are `?Send` on wasm32 only (JS futures are not `Send`; the target is single-threaded). Native contracts are unchanged.

The hand-written `bridge`/`engine` modules are declared through `custom_rust_modules` in alef.toml; the regenerated `lib.rs` is included, no hand edits.

## Verification

`cargo check`/`clippy --all-targets` clean on wasm32; host build clean under workspace feature unification. `engine_ocr.test.ts` and `engine_ner.test.ts` cover geometry round-trip, degrade paths, category forwarding, and error paths against the built package. Host-target `cargo check -p xberg-wasm` in isolation fails on main already (tokio-gated paths); unchanged here.

**BLOCKED follow-up (#1269):** in-crate `#[wasm_bindgen_test]` suites wait on xberg-io/alef#182 shipping in a release.


tobocop2:
@jamon8888 tagging you here since this is your work. One scoped feature (the OCR bridge with per-line geometry and the engine handle). This is the shape we look for when reviewing code for next time.

tobocop2:
Follow-up: once xberg-io/alef#182 lands and the pin is bumped, the in-crate wasm-bindgen test suites can come back via `[crates.wasm.extra_dev_dependencies]`. The e2e vitest coverage in this PR carries the same cases until then.

Goldziher:
Wow, nice

Goldziher (approved):

PR #1261, opened 2026-07-18
feat(redaction): make token redaction reversible for passphrase holders, with per-person erasure

### LINKED ISSUE #1260: feat: reversible redaction: restore originals with a passphrase, erase one person's data on request

Today, redacting a document destroys the original values forever. This adds a way to keep them safely: the substitutions are captured into a map, encrypted under a passphrase, and stored by the caller. An authorized reviewer can restore a document, and one person's data can be found and deleted on request without touching anyone else's.

Today `redaction` with `RedactionStrategy::TokenReplace` swaps each PII hit for a token like `[EMAIL_1]` and throws the original away. That is correct for pure redaction, but several real workflows need controlled reversal: a reviewer with the passphrase re-inflating a redacted document, or a data-subject request ("forget everything about alice@example.com") that must remove one subject's entries without touching the rest.

Concretely, a new `redaction-rehydrate` feature gates a `text::redaction::rehydration` module:

- `encrypt_map(map, passphrase) -> Vec<u8>` / `decrypt_map(bytes, passphrase) -> RehydrationMap`: AES-256-GCM with an scrypt-derived key, versioned `XPII\x01` wire format, fresh salt and nonce per encryption.
- `find_subject(map, query)`: match a token exactly or an original value by case-insensitive substring.
- `forget_subject(map, query)`: remove every matching entry and return what was removed, so the caller re-encrypts and persists the remainder.

The map never touches disk inside xberg; persistence and passphrase handling stay with the caller.

Follow-up, not in the first PR: wiring capture into the document-redaction engine itself (`redact_capturing_rehydration_map`) so an `ExtractedDocument` pass can emit the map directly, and the wasm/binding surface on top.

### Attribution

Designed and implemented by @jamon8888 in #1253. This issue tracks landing it upstream as its own reviewable unit.

tobocop2:
Closes #1260. Design and implementation by @jamon8888 in #1253; his commit carries his authorship, fixes are separate commits.

## What was broken

Redaction replaces PII with tokens like `[EMAIL_1]` and destroys the originals. That is right for pure redaction, but it makes two legitimate needs impossible: an authorized reviewer holding a passphrase cannot restore a document, and a person asking to be forgotten cannot have just their entries deleted from what was captured.

## What this adds

A `redaction-rehydrate` feature gating `text::redaction::rehydration`:

- `encrypt_map`/`decrypt_map`: AES-256-GCM, scrypt-derived key, versioned `XPII\x01` wire format, fresh salt and nonce per encryption
- `find_subject`: exact token match or Unicode case-insensitive substring on originals
- `forget_subject`: removes matching entries and returns them so the caller re-encrypts the remainder
- `redact_capturing_rehydration_map`: same document pass as `redact`, returning the token map (captured from the `TokenCounter`'s dedup cache, so every call site is covered with no per-site changes)

Key material and decrypted plaintext are `Zeroizing`. The map never touches disk inside xberg. Types carry `alef(skip)`; binding surface lands on a later canonical regeneration.

## Verification

27 tests green, including a TypeScript cross-implementation fixture, wrong-passphrase rejection, fresh salt/nonce per encryption, Unicode matching, and a redact-then-rehydrate round trip. Clippy clean; `alef generate` produces no output changes.


tobocop2:
@jamon8888 tagging you here since this is your work. One scoped feature (encrypted rehydration maps with find/forget-subject). This is the shape we look for when reviewing code for next time.

Goldziher (approved):

PR #1263, opened 2026-07-18
feat(ner): detect entities without ONNX Runtime via a pure-Rust GLiNER2 implementation

### LINKED ISSUE #1262: feat: entity detection that runs without ONNX Runtime

Entity detection (finding people, organizations, emails, phone numbers in text) currently requires ONNX Runtime, a native library. Browsers cannot load it, and some deployments do not want to ship it. This adds a second implementation of the same model family in pure Rust (via the Candle ML framework), so entity detection works anywhere Rust compiles.

Today NER routes through `ner-onnx` (ORT) or an injected backend. That leaves the `no-ort-target` family (the whole reason that feature set exists) without in-binary NER at all.

Concretely, a new `crates/xberg-gliner-candle` crate implements the GLiNER2 architecture (DeBERTa-v2 encoder, span representation, count-LSTM, schema-gather and scoring heads) with runtime PEFT LoRA adapter merge-at-load and a streaming safetensors loader that converts F32→F16 from raw bytes to avoid materializing intermediate tensors on memory-tight targets. A `ner-candle` feature wires it into `text::ner::candle::CandleBackend` behind the existing `NerBackend` trait, so the redaction pipeline and NER post-processor consume it like any other backend.

### What lands when

- **First PR:** the crate plus the native `ner-candle` path.
- **Follow-ups:** the wasm32 wiring (a future `ner-candle-wasm` feature) needs the `NerBackend` `?Send`-on-wasm32 discussion resolved first; and two upstream dependency bugs discovered during browser testing are now filed with verified reproducers: huggingface/candle#3750 (debertav2 hardcodes F32 scalars, breaking F16 forward) and safetensors/safetensors#817 (metadata validation spuriously overflows on 32-bit targets for tensors past 2^32 bits). Until those land upstream, the wasm path needs patched builds; the native path does not. A third vendored patch on the fork (candle-nn LayerNorm skip-fused-for-F16) turned out not to be load-bearing: with only the debertav2 fix applied, F16 forward passes on stock candle-nn 0.11.

### Attribution

Designed and implemented by @jamon8888 in #1253, including real-browser verification of model load and entity detection. This issue tracks landing it upstream in reviewable stages.

tobocop2:
Closes #1262. Design and implementation by @jamon8888 in #1253, including real-browser verification; his commit carries his authorship, fixes are separate commits.

## What was broken

Detecting entities (people, organizations, emails, phone numbers) required ONNX Runtime, a native library that cannot run in browsers or other restricted targets. Those targets had no entity detection at all, which also means no ML-assisted redaction there.

## What this adds

A `candle` feature on `xberg-gliner` that provides GLiNER2 inference on Candle (DeBERTa-v2 encoder, span scoring heads, PEFT LoRA merge-at-load, streaming F32-to-F16 safetensors loader), living at `xberg_gliner::candle` with `cuda`/`metal` pass-throughs. Per review, this is one crate rather than a sibling: the V2 prompt-encoding surface the split would have forced `pub` stays crate-private, and the shared v2 pipeline only compiles when a backend that uses it does. `ort` becomes optional behind `ort-backend`; default builds keep `ort-bundled`, unchanged. A `ner-candle` feature exposes `text::ner::candle::CandleBackend` behind the existing `NerBackend` trait with a process-wide `get_or_init` cache.

Native path only. `CandleBackend` carries `alef(skip)` until the NER dispatch follow-up.

## Verification

All gliner tests green in the merged crate (candle unit tests included), with regression tests for truncation bounds and non-char-boundary offsets. Clippy clean on every feature cell. The unit test sweep tests `xberg-gliner` with an explicit feature list instead of `--all-features` (cuda/metal cannot build on runners): default ONNX features plus `candle` and `ort-dynamic` everywhere, with only Linux aarch64 dropping `candle` (gemm-f16 needs fullfp16), so ONNX coverage no longer disappears with the skip. Lockfile change is purely additive.

**BLOCKED follow-up (#1268):** browser wiring waits on huggingface/candle#3751 and safetensors/safetensors#818 merging and releasing.


tobocop2:
@jamon8888 tagging you here since this is your work. One scoped feature (the GLiNER2 Candle backend plus the v2 pipeline it needs). This is the shape we look for when reviewing code for next time.

Cargo.toml
```diff
@@ -5,6 +5,7 @@ members = [
     "crates/xberg-cli",
     "crates/xberg-ffi",
     "crates/xberg-gliner",
+    "crates/xberg-gliner-candle",
```
Goldziher: Can't this be in the same crate as the other impl?

Goldziher (approved):

Cargo.toml
tobocop2: Makes sense for the smaller public API

I did some  digging and it needs a CI restructure: candle's cuda/metal features cannot build on the runners, so the all-features sweep that today just excludes the candle crate would have to exclude xberg-gliner too. 

The fix is to test the merged crate with an explicit feature list instead, everything except cuda/metal (which no CPU runner can exercise anyway), keeping full ONNX and CPU-candle coverage, with the gemm-f16 skip narrowed to just the candle tests on Linux aarch64.

PR #1285, opened 2026-07-22
fix: link Microsoft ONNX Runtime on glibc Linux to honor the manylinux floor

### LINKED ISSUE #1284: chore: Consolidate ONNX Runtime across release channels: the pyke prebuilt static-links glibc-2.38 symbols and breaks glibc 2.28–2.37
## Problem

ONNX Runtime is pulled into the release artifacts via `ort/download-binaries`, which downloads and **statically links** the pyke prebuilt runtime. That prebuilt is compiled against glibc ≥ 2.38, so it drags unversioned `__isoc23_*` / `__libc_single_threaded` symbols into the binary. The manylinux tag / package metadata only reflects *versioned* glibc deps, so the artifacts install everywhere and then crash at import/load on glibc 2.28–2.37:

```
undefined symbol: __isoc23_strtoll        # e.g. glibc 2.35
undefined symbol: __libc_single_threaded  # e.g. glibc 2.31
```

Every channel handles ONNX Runtime differently — some static-link pyke, some reach Microsoft's runtime by a bespoke path — so the floor silently regresses per channel. Confirmed by `objdump` on rc30: the Python wheel, Go FFI, C FFI, Java, and Elixir NIF are **broken** (static pyke); C# and Node reach Microsoft's runtime by their own ad-hoc mechanisms.

## Locked design: one consolidated strategy

**Dev (unchanged, no regression):** `ort/download-binaries` stays the default, so a plain `cargo build` works with no system ORT. The prebuilt it fetches needs glibc ≥ 2.38 to run — documented, dev-only.

**Release (every channel):**
1. **Link** Microsoft's ONNX Runtime (glibc-clean: 2.27 floor, zero `__isoc23`) via the `system` strategy, using the existing `setup-onnx-runtime` action. No `download-binaries` in the shipped build; no per-channel special-casing.
2. **Bundle** it beside the artifact with the mature tool for each artifact type — `auditwheel` (wheel), `delocate`/`dylibbundler` (macOS), and a single shared ldd-closure + `patchelf --set-rpath '$ORIGIN'` vendorer (`vendor-native-closure.sh`) for Node, the FFI libraries, and the Elixir NIF. Node drops its bespoke vendoring script for this shared vendorer, keeping napi-rs's standard manylinux build container.
3. **Verify** every artifact with `verify-glibc-floor.sh`, which fails the release if a binary references too-new symbols or omits the bundled runtime — the check whose absence let this ship. (`auditwheel` does not catch the *unversioned* symbols.)

**Scope:** wheel, Go/C FFI, Java, Elixir, Node — across linux-gnu, linux-musl, macOS, and Windows. Result: pyke's prebuilt runtime is used by no shipped artifact, every channel ships the same clean Microsoft runtime by the same mechanism, and the floor is enforced in CI.

## Verified mechanism (pod, glibc 2.31 / 2.35)

- `download-binaries` **off** + `ORT_STRATEGY=system` + `ORT_LIB_LOCATION` → builds, links Microsoft ORT (`NEEDED onnxruntime`), `__isoc23=0`, `RPATH=$ORIGIN`, loads with no `LD_LIBRARY_PATH`.
- Same via cargo-zigbuild at the glibc-2.28 floor → highest versioned `GLIBC_2.28`.
- `verify-glibc-floor.sh` flags every broken rc30 artifact (wheel + Go FFI: 3 unversioned `__isoc23` + `__libc_single_threaded`) and is not fooled by merely bundling the runtime.

| ONNX Runtime | Highest versioned GLIBC | `__isoc23` |
|---|---|---|
| pyke prebuilt (current, static) | 2.28 versioned, **2.38 unversioned** | 3 |
| Microsoft 1.24.2 | 2.27 | 0 |

tobocop2:
Fixes #1284.

## Problem

The Python wheel and the Go/C FFI and Java artifacts statically link the pyke prebuilt ONNX Runtime, which is compiled against glibc ≥ 2.38. That pulls **unversioned** `__isoc23_*` and `__libc_single_threaded` symbols into the binary.

`auditwheel` only checks *versioned* glibc requirements, so it never sees them — the wheel gets a `manylinux_2_28` tag it can't honor and dies at import on glibc 2.28–2.37:

```
undefined symbol: __isoc23_strtoll
```

## Fix

Link **Microsoft's** ONNX Runtime (glibc-clean, floors at 2.27) in those builds — the exact runtime the Docker images and the C#/Elixir/Node artifacts **already use**. No source changes; the affected builds were simply the ones that never got that treatment.

- **Wheel:** download Microsoft ORT in-container and point `ort-sys` at it (`ORT_STRATEGY=system`). `auditwheel` then bundles it, like any other native dep.
- **Go/C FFI, Java:** enable the existing `setup-onnx-runtime` step on linux-gnu (it was Windows-only), matching C#/Elixir.

C#, Elixir, and Node already link Microsoft ORT on linux-gnu and are unchanged. musl, macOS, and Windows are unaffected.

## Why not just build in an older-glibc container?

The wheel is already built in a `manylinux_2_28` (glibc 2.28) container. The bad symbols don't come from the build compiler — they're baked into a **downloaded prebuilt** ONNX Runtime compiled elsewhere against glibc 2.38. Linking a too-new prebuilt into an old container doesn't strip its symbols. The fix is to link a glibc-clean runtime instead.

## Evidence

On Ubuntu 20.04, x86_64, **glibc 2.31**:

- Before — the shipped rc30 wheel installs, then import fails: `undefined symbol: __libc_single_threaded`.
- Microsoft ORT 1.24.2 loads on the same host: `OrtGetApiBase` present, highest versioned GLIBC **2.27**, **zero** `__isoc23`.

| ONNX Runtime | Highest versioned GLIBC | `__isoc23` |
|---|---|---|
| pyke prebuilt (current, static) | 2.28 versioned, **2.38 unversioned** | 3 |
| Microsoft 1.24.2 | 2.27 | 0 |

The `publish.yaml` changes run only in release CI, so they are validated there.

### DIFF
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -36,6 +36,10 @@ on:
   repository_dispatch:
     types: [publish-release]
 
+env:
+  # glibc floor for the Linux artifacts: zigbuild target suffix + post-build assertion.
+  GLIBC_FLOOR: "2.28"
+
 permissions:
   contents: read
   id-token: write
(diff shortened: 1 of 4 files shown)


tobocop2:
Superseded: this started as a wheel-only fix and the design has since been locked to a **consolidated** ONNX Runtime strategy across all release channels (see #1284). Do not merge this in its current form — it will be replaced by the consolidated implementation. Keeping it draft as the working branch.

Issue #1311, opened 2026-07-24
feat: word-perfect support


Goldziher:
## Summary

Add WordPerfect document extraction (`.wpd`, `.wp5`, `.wp6`, and the wider WordPerfect binary family from WP 4.2 through the X-series) by binding to [**libwpd**](https://libwpd.sourceforge.net/). One auto-detecting extractor covers the whole family. This is a self-contained, native-backed extractor and a good contribution for someone comfortable with C++ ↔ Rust FFI.

## Licensing requirement (please read first — non-negotiable)

libwpd and its document-model dependency [librevenge](https://wiki.documentfoundation.org/DLP/Libraries/librevenge) are dual-licensed **MPL-2.0 OR LGPL-2.1+**. xberg is MIT and ships **statically linked** native artifacts across many registries (crates.io, PyPI, npm, Maven, NuGet, …).

- **Build against the MPL-2.0 arm of both libraries — not the LGPL arm.**
- MPL-2.0 is file-level copyleft and explicitly permits static linking into a larger, differently-licensed work (this is how LibreOffice ships libwpd). Obligations are limited to keeping the libwpd/librevenge source and any modifications *to their files* available under MPL, plus preserving license notices in distributed artifacts.
- The LGPL arm is rejected because LGPL-2.1 static linking imposes relink/object-file obligations we do not want on redistributed binaries.
- No copyleft reaches xberg's MIT core or the language bindings when the MPL arm is used.
- Note: modern libwpd (0.10+) does **not** require boost.

### 1. Native crate `xberg-libwpd` (model on `crates/xberg-libheif/`)
- New crate wrapping the C++ libraries. Follow the `xberg-libheif` pattern (thin wrapper over a native lib, `links = ...`, system-lib provisioning). The heavier `xberg-tesseract` crate (vendored source + `build.rs` + feature-gated WASM path) is the reference for building C/C++ from source.
- There is **no C API and no existing Rust `-sys` crate** for libwpd/librevenge — the whole native layer is greenfield. Do not assume a crate exists on crates.io; there isn't one.

### 2. The librevenge callback shim (the core of the work)
[...]

PR #1312, opened 2026-07-24
feat(libwpd): add xberg-libwpd crate for WordPerfect text extraction

MannXo:
## Related

Part of #1311. First of two PRs: this adds the standalone native crate; a follow-up wires it into the extraction pipeline (the `wordperfect` feature, MIME entry, extractor, and registration).

## Description

Adds `xberg-libwpd`, a native crate that extracts text from WordPerfect documents (the WP 4.2 through X-series binary family) via [libwpd](https://libwpd.sourceforge.net/) and its document-model dependency librevenge.

libwpd has no `extract()` entry point; it drives librevenge's SAX-like `RVNGTextInterface`. The crate provides:

- a hand-written C++ shim (`src/shim.cpp`) that implements the interface, accumulates a plain-text rendering (paragraphs, list items, tables), and exposes a flat C API returning owned UTF-8. Every entry point catches C++ exceptions so nothing unwinds across the FFI boundary.
- a safe Rust surface (`extract_text`, `is_supported`) over that API.
- a `build.rs` that downloads the librevenge 0.0.6 and libwpd 0.10.3 release tarballs (checksum verified, cached under `OUT_DIR`) and compiles them from source together with the shim into one static library, following the `xberg-tesseract` build-from-source model.

Both libraries are built and linked under their MPL-2.0 arm, per the licensing requirement in #1311. `THIRD_PARTY_LICENSES.md` documents the static-linking posture.

Build-time note: contrary to the "no boost" line in the issue, both librevenge and libwpd need header-only `boost::spirit` at build time (flagged on the issue already). The Linux and macOS `install-system-deps` scripts add boost and zlib headers accordingly.

Platform support is desktop only (Linux glibc/musl and macOS). On wasm/android/windows the crate compiles to stubs returning `WpdError::UnsupportedPlatform`, so those builds pull in no C++ toolchain.

## Testing

- `cargo test -p xberg-libwpd`: 4 tests pass. Three error-path tests (empty, random bytes, WordPerfect-magic-but-truncated) run everywhere and confirm malformed input is rejected without crashing across the FFI boundary. A decode test extracts text from a real `.wpd` sample and asserts non-empty output; it skips gracefully when the `test_documents` submodule is absent.
- `cargo clippy -p xberg-libwpd --all-targets -- -D warnings`, `cargo fmt --check`, and `clang-format` on the shim all pass.
- Verified on macOS (arm64), building both libraries entirely from source.

## Notes

- The decode test looks for a WordPerfect sample in the `test_documents` submodule (`wordperfect/sample.wpd`). Happy to open a companion PR adding one, or you may prefer a sample you already have rights to. The test skips cleanly until then.
- Sources are fetched at build time. If you would rather have a CI source-cache step (as tesseract has), I can add one.

## Checklist

- [ ] CI passing


Goldziher (commented):
## Technical review — `xberg-libwpd` WordPerfect crate

The FFI hygiene here is genuinely good (details at the bottom). The concerns are that the crate currently ships untested and unwired, with a licensing doc that describes it as already integrated. Recommend landing this explicitly as "crate only, step 1 of N" with the fixes below, rather than as a shipped feature.

### 🔴 Should block / must track before wiring into the pipeline

1. **Embedded-NUL silent truncation.** The shim (`src/shim.cpp`) copies all `n` bytes but returns only a NUL-terminated `char*`, and `src/lib.rs` reads it back with `CStr::from_ptr(out).to_str()`, which stops at the first embedded NUL. Any extracted text containing a `\0` is silently truncated with no error. `TextCollector::insertText` appends `RVNGString::cstr()` content without NUL filtering, so nothing enforces the no-NUL assumption. **Fix:** return an explicit length (`*mut usize`) across the ABI and build the `String` from `slice::from_raw_parts` + `String::from_utf8` instead of `CStr`.

2. **`catch (...)` does not cover the real risk class.** It correctly stops a *thrown* C++ exception from unwinding into Rust, but libwpd/librevenge (pinned to the ~2015-era `libwpd-0.10.3` / `librevenge-0.0.6`) fail on malformed input as **undefined behavior — segfault, heap corruption, OOB read/write** — which `catch (...)` cannot intercept. Not exploitable today (the crate is unwired), but before it enters the extraction pipeline it needs a `cargo-fuzz`/AFL pass against `extract_text`, and/or isolation via `spawn_blocking` in a size/time-bounded sandbox. The fuzz-style tests only prove three specific inputs don't crash; they say nothing about structured-but-malformed WP4.2/WP6 records, which is where parser memory bugs live.

3. **The only real-decode test never runs.** `tests/decode.rs::extracts_text_from_sample` reads `test_documents/wordperfect/sample.wpd` and early-returns if it's absent — and that directory does not exist in the pinned `test_documents` submodule, so the test is skipped on every machine including CI. Every other test exercises only error paths on garbage input. **Fix:** add a small, license-clean `.wpd` fixture to the submodule and assert real extracted content.

4. **Crate is unwired, but the licensing doc says it's shipped.** There is no `wordperfect` Cargo feature anywhere in the workspace and no `.wpd` MIME mapping / extractor wiring — yet `THIRD_PARTY_LICENSES.md` states *"optional `wordperfect` Cargo feature (part of `full`/`formats`)"*. That's inaccurate as of this diff and would mislead a licensing audit into believing libwpd (LGPL) is already linked into shipped builds. **Fix:** either correct the wording to "not yet integrated," or land the MIME/extractor wiring in this PR so the claim is true.

### 🟡 Lower

5. No policy-driven size cap beyond the 4 GiB `u32::MAX` guard, and the API is a plain blocking call with no timeout/cancellation. Whoever integrates it must run it via `spawn_blocking` with a `SecurityLimits`-style cap, not directly from an async context.
6. `build.rs` fetches libwpd/librevenge from two SourceForge URLs (not independent mirrors) — SHA-256-verified so tamper-safe, but availability-fragile. Consistent with the existing `xberg-tesseract` precedent.
7. `Send`/`Sync` are implied for the public functions but libwpd's freedom from internal global/static mutable state isn't confirmed. A concurrent-call smoke test (N threads, different buffers) would be cheap insurance.

### ✅ Done correctly (verified)

- Allocator pairing: `std::malloc` in the shim ↔ `xberg_wpd_free_string` → `std::free`. No cross-allocator mismatch.
- Ownership ordering in `lib.rs`: the buffer is copied into an owned `String` **before** `xberg_wpd_free_string`, and freed on both the `Ok` and `Utf8Error` branches — no use-after-free, no leak on the invalid-UTF-8 path.
- The Rust-side `data.len() > u32::MAX` reject preempts the shim's `static_cast<unsigned int>(len)` narrowing.
- Null handling is defensive on both sides.
- `install-linux.sh` / `install-macos.sh` additions match what `build.rs` actually needs.
- The MPL-2.0-vs-LGPL static-linking rationale in `THIRD_PARTY_LICENSES.md` is well-reasoned.

— Claude (technical review)

MannXo:
@Goldziher thanks for the review. Will push a commit reflecting.
Note that leaving the wiring to a subsequent PR was intentional.

crates/xberg-libwpd/src/error.rs
```diff
@@ -0,0 +1,47 @@
...
+    #[error("libwpd returned invalid UTF-8")]
+    InvalidUtf8,
+
+    /// WordPerfect extraction is not available on this platform.
+    #[error("WordPerfect extraction is not supported on this platform")]
+    UnsupportedPlatform,
+}
+
+impl WpdError {
+    /// Map a shim result code (see `shim.cpp`) to an error. Code 0 is success
```
Goldziher, line 37: add `~keep` to comments you want to be kept. Also install `poly` on your machine and run poly lint.

Goldziher: here: https://github.com/Goldziher/poly

crates/xberg-libwpd/src/lib.rs
```diff
@@ -0,0 +1,86 @@
+//! WordPerfect text extraction for Xberg.
+//!
+//! Thin, safe wrapper over [libwpd](https://libwpd.sourceforge.net/) and its
+//! document-model dependency librevenge, both built from source against their
+//! MPL-2.0 arm (see `build.rs`). libwpd covers the whole WordPerfect binary
+//! family (WP 4.2 through the X-series).
+//!
+//! libwpd has no `extract()` entry point; it drives a librevenge callback
+//! interface. A hand-written C++ shim (`src/shim.cpp`) implements that
+//! interface, accumulates a plain-text rendering, and exposes a flat C API this
```
Goldziher: there is no way to get structured data through this?

We would optimally want to preserve layout information (headers, formatting etc.) and be able to translate to markdown.

crates/xberg-libwpd/src/lib.rs
```diff
@@ -0,0 +1,86 @@
...
+//! crate wraps. WordPerfect support is desktop-only (Linux and macOS); on other
+//! platforms the functions return [`WpdError::UnsupportedPlatform`].
+
+mod error;
+
+pub use error::WpdError;
+
+#[cfg(any(target_os = "linux", target_os = "macos"))]
```
Goldziher: missing windows - we need windows as well. You can test it in CI here as needed.

crates/xberg-libwpd/src/lib.rs
```diff
@@ -0,0 +1,86 @@
...
+        if data.is_empty() || data.len() > u32::MAX as usize {
+            return false;
+        }
+        // SAFETY: `data` is a valid slice of `len` bytes; the shim only reads it
+        // and catches any C++ exception internally.
+        unsafe { xberg_wpd_is_supported(data.as_ptr(), data.len() as c_ulong) != 0 }
+    }
+
+    /// Extract the text of a WordPerfect document held entirely in memory.
+    pub fn extract_text(data: &[u8]) -> Result<String, WpdError> {
```
Goldziher, line 54: ooptimally we would be returning a DTO from this function rather than a string. If impossible we will do plain text, but note - out extractors implement markdown where possible.

crates/xberg-libwpd/src/shim.cpp
```diff
@@ -0,0 +1,182 @@
...
+        text += '\t';
+    }
+    void closeTableRow() override {
+        text += '\n';
+    }
+    void closeTable() override {
+        text += '\n';
+    }
+
+    // Remaining pure virtuals are structural and produce no text of their own.
```
Goldziher: are you certain of this? 

We need some WP files to test as well. 

Test documents should go into the `test_documents` git submodule (its a separate repository, you can open a PR: https://github.com/xberg-io/test_documents to add them. 

WP5 and 6 I would wager, but whatever we will need to support. I would actually test what can be extracted from these and implement these in full - where possible.

crates/xberg-libwpd/src/shim.cpp
```diff
@@ -0,0 +1,182 @@
...
+    void drawRectangle(const RVNGPropertyList &) override {}
+    void drawEllipse(const RVNGPropertyList &) override {}
+    void drawPolygon(const RVNGPropertyList &) override {}
+    void drawPolyline(const RVNGPropertyList &) override {}
+    void drawPath(const RVNGPropertyList &) override {}
+    void drawConnector(const RVNGPropertyList &) override {}
+};
+} // namespace
+
+extern "C" {
```
Goldziher, line 553: formatting?

crates/xberg-libwpd/src/shim.cpp
```diff
@@ -0,0 +1,182 @@
...
+        const size_t n = collector.text.size();
+        char *buf = static_cast<char *>(std::malloc(n + 1));
+        if (!buf)
+            return XBERG_WPD_OUT_OF_MEMORY;
+        std::memcpy(buf, collector.text.data(), n);
+        buf[n] = '\0';
+        *out_text = buf;
+        return XBERG_WPD_OK;
+    } catch (...) {
+        return XBERG_WPD_PANIC;
```
Goldziher, line 652: can we add some tracing / logging? Will be difficult to debug

crates/xberg-libwpd/build.rs
```diff
@@ -0,0 +1,187 @@
...
+//! (see `src/lib.rs`), so wasm/android/windows builds never pull in a C++
+//! toolchain.
+//!
+//! Both libraries are built against their MPL-2.0 arm. They are downloaded from
+//! their upstream release tarballs at build time (checksum-verified) and cached
+//! under `OUT_DIR`, mirroring how `xberg-tesseract` provisions its native
+//! dependencies. librevenge and libwpd both require boost headers at build time
+//! (header-only `boost::spirit`), which must be present on the system.
+
+#[cfg(any(target_os = "linux", target_os = "macos"))]
```
Goldziher: same - windows needed.

Goldziher:
## Follow-up — `build.rs` should use a persistent, overridable cache dir

Comparing against `crates/xberg-tesseract/build.rs` (the established precedent for downloading + compiling a native C/C++ dep at build time), the env-var handling here is close but missing one piece.

**Already correct:**
- `BOOST_INCLUDE_DIR` override with `cargo:rerun-if-env-changed=BOOST_INCLUDE_DIR` (`build.rs:49`, `:180`).
- `rerun-if-changed` on `build.rs` and `src/shim.cpp`.

**Gap — cache location:** the tarballs are downloaded/extracted/compiled under `OUT_DIR/native` (`build.rs:139`). `xberg-tesseract` instead uses a persistent cache dir overridable via `TESSERACT_RS_CACHE_DIR` (falling back to a `$HOME`-based cache), precisely so the heavy native build survives `cargo clean`. Because `cargo clean` wipes `OUT_DIR`, every clean build here **re-downloads both SourceForge tarballs and recompiles libwpd + librevenge from source** — slow, and it fails outright offline or during a SourceForge outage (compounding the mirror-availability concern raised in the main review).

**Suggested fix — mirror the tesseract pattern:**
- Resolve the cache root from an override env var (e.g. `XBERG_LIBWPD_CACHE_DIR`), falling back to a persistent per-user/per-target cache dir rather than `OUT_DIR`.
- Add `cargo:rerun-if-env-changed=XBERG_LIBWPD_CACHE_DIR`.
- Keep the existing SHA-256 verification and the `.ok` marker so a populated cache is reused rather than re-fetched.

**Optional (not blocking):** there's no way to link a system-installed libwpd (no pkg-config / `*_NO_VENDOR` opt-out) — it always vendors from source even where `libwpd-dev` is available. `xberg-tesseract` also always vendors, so this is consistent with precedent; noting it only as a possible convenience for packagers.

— Claude (technical review)

crates/xberg-libwpd/Cargo.toml
```diff
@@ -0,0 +1,35 @@
...
+
+# libwpd and librevenge are built from vendored-at-build-time C++ source, so the
+# native toolchain and downloader are only needed on the desktop targets that
+# actually compile them. WordPerfect support is gated off wasm/android/windows.
+[target.'cfg(any(target_os = "linux", target_os = "macos"))'.build-dependencies]
+cc = "1.3"
+reqwest = { workspace = true, default-features = false, features = ["blocking", "rustls"] }
+flate2 = "1.0"
+tar = "0.4"
+sha2 = "0.10"
```
Goldziher: why not workspace

Goldziher (commented):
Nice work. This needs more love - and substantially more tests. Its also needs to involve updating publish.yaml, and documentation.

crates/xberg-libwpd/src/error.rs
MannXo: Ran poly lint. FYI even the crates already in `poly.toml`'s comment/doc-comment exemption list (xberg-ffi etc.) still surface these same warnings , the exemption doesn't seem to actually suppress the `uncomment` rule for anyone right now, and poly isn't wired into CI. Added `xberg-libwpd` to that exemption list for consistency anyway. lmk if I'm missing something about how that config is meant to work.

crates/xberg-libwpd/src/lib.rs
MannXo: Added `extract_markdown()`

crates/xberg-libwpd/src/lib.rs
MannXo: fair point but I'd suggest to hold this for the follow-up PR, once the crate is wired into `xberg`/`xberg-cli`. The Windows CI job doesn't build this crate at all today (it's a standalone workspace member), so support landed now would ship with nothing verifying it. It also needs real work beyond a cfg flip  (no system boost or zlib on Windows, unlike Linux/macOS). Happy to take it on once there's an actual Windows target and CI job to prove it against.

crates/xberg-libwpd/src/lib.rs
MannXo: Went with the markdown convention you mentioned rather than a full DTO. As mentioned in a few comments above, added `extract_markdown()`for this purpose.

crates/xberg-libwpd/src/shim.cpp
MannXo: I wasn't certain since I didnt have a real file to test it, but made it somehow more testable now. footnotes/endnotes/comments/text boxes/headers/footers are now collected separately and spliced back behind `[kind: ...]` markers instead of straight into body text, proven by a self-test that drives the collector's callbacks directly. Passing

crates/xberg-libwpd/src/shim.cpp
MannXo: Ran clang-format

crates/xberg-libwpd/src/shim.cpp
MannXo: reflected

crates/xberg-libwpd/build.rs
MannXo: same follow-up plan as the `lib.rs:21` reply above.

crates/xberg-libwpd/Cargo.toml
MannXo: `cc`/`flate2`/`tar`/`sha2` now come from `[workspace.dependencies]`

crates/xberg-libwpd/src/lib.rs
Goldziher: its not one or the other. You misunderstood. Please see the different places where we construct an internal document. We render markdown only in the end.

crates/xberg-libwpd/src/lib.rs
Goldziher: what do you mean? we use windows runners in CI as well, once you add a windows target this becomes testable. 

I am afraid this cannot be left to a followup. This cant be merged without this.

MannXo: My bad. Windows (MSVC) is now a supported target

crates/xberg-libwpd/src/lib.rs
MannXo: fair enough. pushed a fix for this

crates/xberg-libwpd/src/error.rs
Goldziher: It is wired in and used. See `ci-lint:59` - its an org level shared gh action. I'll cehck the other part of what you wrote. It shouldnt be exempted in any case.

Goldziher:
## Review — blocking issues before merge

Consolidated from a review of the FFI/C++ surface, the `build.rs` supply-chain, and the workspace/CI wiring. Grouped by severity; file:line against the current head.

### Must-fix (blocking)

**1. musl release builds break — CI does not catch it.** `Cargo.toml` adds `crates/xberg-libwpd` to `[workspace] members`, but none of the four musl images (`docker/Dockerfile.musl-build`, `-ffi`, `-node`, `-rustler`) copy or `sed`-strip the new crate. They COPY the root `Cargo.toml`/`Cargo.lock` (which now list the member) but not the crate dir, and Cargo loads the whole workspace before building any `--package` → `failed to read crates/xberg-libwpd/Cargo.toml`. This kills `cli-binaries-musl`, `java-natives-musl`, `node-bindings-musl`, and the rustler musl job. `ci-rust` is green only because it uses a full checkout. Fix: add `crates/xberg-libwpd/` to each musl Dockerfile's COPY list (it stubs out on musl, no C++ toolchain pulled), or add it to the sed strip list.

**2. `r3650242215` (poly `uncomment` exemption) is still open.** `poly.toml` still carries `"crates/xberg-libwpd/**" = ["comment", "doc-comment"]`. This crate is hand-written, so it should not be blanket-exempted from the uncomment lint. Note the exemption block is generated from `alef.toml`'s `[workspace.poly.per-file-ignores]`, so a bare `poly.toml` edit is overwritten on regen — the change belongs in `alef.toml`. The crate's genuinely load-bearing comments (SAFETY/FFI-invariant/why-not-what) should be preserved with `~keep`; the rest removed.

### High

**3. Embedded-NUL truncation in `insertText` is not actually fixed.** `src/shim.cpp:72` constructs `std::string` from `RVNGString::cstr()` (a bare `const char*`), which truncates at the first `\0` via `strlen` — before the new length-aware FFI transport ever runs. So a WordPerfect text run containing an embedded NUL is silently truncated at record time and returned as `Ok` with data loss. The doc comment at `src/lib.rs:110` claiming an embedded NUL "can't silently truncate the result" is accurate for the FFI boundary but false for this path. Fix: `std::string(s.cstr(), s.len())`, and add a regression test that drives `insertText` with a NUL-containing run through the public `extract_text` path (not `xberg_wpd_self_test_separation`, which never touches NUL bytes).

**4. Download-at-build model breaks docs.rs and crates.io consumers, yet the crate is in the publish list.** `build.rs` fetches the librevenge/libwpd tarballs from SourceForge at build time. docs.rs builds in a network-isolated sandbox (no `[package.metadata.docs.rs]` opt-out present) → build-script panic → docs fail for the published crate. Ordinary `cargo build`/`cargo install` consumers also need network + system boost/zlib. `publish.yaml` adds the crate to `publish-crates`. Decide explicitly: vendor the tarballs in-repo (idiomatic `-sys` model, also satisfies MPL-2.0 §3.2 source-availability without depending on SourceForge uptime), or don't publish it.

### Medium

**5. `xberg_wpd_self_test_separation` (`src/shim.cpp:489`) is `extern "C"` with no `try`/`catch`,** violating the file's own stated invariant (shim.cpp:13). It calls `RVNGString`/`std::string`/`render()`, all of which can throw `std::bad_alloc`; an exception unwinding across the `extern "C"` frame into Rust is UB. Test-only today, but wrap the body in `try { ... } catch (...) { return 0; }` for consistency.

**6. The only real-decode test never runs.** `tests/decode.rs:44` early-returns when `test_documents/wordperfect/sample.wpd` is absent — and no such fixture exists in the submodule. Every other test only exercises the early-rejection path (random bytes hit `WPD_CONFIDENCE_NONE` immediately) and never reaches the `DocumentBuilder` callback surface. Combined with #3, nothing would catch the NUL regression. Add a license-clean `.wpd` fixture and a test that reaches the callback path.

**7. `catch (const std::exception&)` (`src/shim.cpp:470`) may be dead code for libwpd's actual exception types,** and `std::strlen(e.what())` is unguarded. Verify librevenge/libwpd's exception hierarchy (bare custom types of that vintage often don't derive from `std::exception`), and add `e.what() ? std::strlen(e.what()) : 0` regardless.

**8. Cache reuse skips checksum re-verification.** `build.rs` (`provision`, ~line 495) verifies the sha256 only on first download; on reuse it trusts the extracted tree if a version-only marker exists. Anything writing to the cache dir (or a restored `target/` cache) can substitute source that is then compiled unverified. Store the tarball digest in the marker and re-verify, or retain the verified tarball and re-extract.

**9. `license = "MIT"` (`Cargo.toml:8`) hides the statically-linked MPL-2.0 from cargo-deny/SBOM tooling.** Because the MPL-2.0 sources are fetched at build time rather than as cargo deps, license scanners see only MIT. `THIRD_PARTY_LICENSES.md` is itself accurate, but the tooling won't surface the obligation — another argument for vendoring.

### Low / notes

- `src/shim.cpp:468` `out_len` narrows `size_t`→`unsigned long` (32-bit on Windows LLP64); safe (buffer sized separately) but silently drops trailing content on a >4 GiB output — not practically reachable.
- `src/shim.cpp:419,455` narrows `len`→`unsigned int` at the C ABI; mitigated by the Rust wrapper's `u32::MAX` check, but the shim itself has no guard for direct callers.
- Unbounded `std::vector<Node>` growth is a DoS vector (deferred to the wiring PR); none of the current tests exercise a pathological-but-valid document, so graceful degradation is unverified.
- `concurrent_calls_do_not_crash` only drives the early-reject path, so the thread-safety claim is weaker than the test comment implies.

### Correct as-is (so these aren't re-raised)

Allocator pairing (`std::malloc`→`xberg_wpd_free_string`→`std::free`, no `CString` mismatch); copy-before-free on every path (no UAF); checked `String::from_utf8` (no unchecked UTF-8); null-pointer discipline on both sides; every `unsafe` block has an upheld SAFETY comment; platform gating reads `CARGO_CFG_TARGET_OS` so wasm/android/windows correctly no-op with no C++ toolchain; checksums are HTTPS-pinned constants, verified before extraction, hard-fail on mismatch.

Goldziher:
@MannXo - rebase on main. Then you will see under the `test_documents` submodule (make sure to clone it) - there is a new wordperfect folder with test documents. We need test that show fully decoding and handing, using these files.

MannXo:
Thanks @Goldziher . the fixtures from `test_documents` were very useful.

- Tests decode real documents now and `extract_text`/`extract_markdown` match the ground truth exactly for `wp42`, `wp50`, `wp51`, `wp6`, `wp_mac1`, `wp_mac3`. It also caught five rendering bugs I'd have missed such as: italic used `_` instead of `*`, trailing spaces sat inside emphasis so `**TITLE: **` wasn't bold, literal `*`/`#`/`$` went out unescaped, and underline and superscript were dropped entirely.

- One correction on the # 3 comment. `s.len()` would make it worse, it's a UTF-8 character count (`librvng_utf8_strlen`), not bytes. That truncates any multibyte doc, `corel_wp6` shows it right away. so I ussed `s.size()`.

- No fixture has an embedded NUL, and I'd rather not hand-roll a `.wpd` pretending to be one. or maybe I didnt properly undersand this one. lmk the correct direction on this.

- the following ones are addressed: # 1, # 2, # 5, # 7, # 8, both narrowing notes. On # 7 I checked the hierarchy, `ParseException`/`FileException`/`GenericException` are bare classes and don't derive from `std::exception`, so that arm only catches allocation failures. `what()` is null-guarded anyway.

- Windows had two real problems. `RVNGBinaryData.cpp` needs `boost/archive/iterators/*`, so vcpkg installs `boost-serialization` too. `RVNGDirectoryStream.cpp` aHnd `RVNGStreamImplementation.cpp` call `S_ISREG`/`S_ISDIR`, which MSVC lacks, so a force-included header supplies them and upstream sources stay unmodified. Will wait for the next CI run.

- On # 4/# 9, I want your direction before building it. Vendoring is my preference and it closes # 9 too, but it's ~3 MB of upstream source in the tree and hard to walk back. The alternative is dropping the crate from `publish-crates` plus a `docs.rs` opt-out. Either works as a follow-up, say which.

also i realized the E2E and Python jobs are red on main too. `xberg-tesseract` wants `cc ^1.4.0` while `e2e/rust/Cargo.lock` pins `1.3.0`, reproduces on a clean `main`. Python is a stale `uv.lock`.

Goldziher:
> Thanks @Goldziher . the fixtures from `test_documents` were very useful.
> 
> * Tests decode real documents now and `extract_text`/`extract_markdown` match the ground truth exactly for `wp42`, `wp50`, `wp51`, `wp6`, `wp_mac1`, `wp_mac3`. It also caught five rendering bugs I'd have missed such as: italic used `_` instead of `*`, trailing spaces sat inside emphasis so `**TITLE: **` wasn't bold, literal `*`/`#`/`$` went out unescaped, and underline and superscript were dropped entirely.
> * One correction on the # 3 comment. `s.len()` would make it worse, it's a UTF-8 character count (`librvng_utf8_strlen`), not bytes. That truncates any multibyte doc, `corel_wp6` shows it right away. so I ussed `s.size()`.
> * No fixture has an embedded NUL, and I'd rather not hand-roll a `.wpd` pretending to be one. or maybe I didnt properly undersand this one. lmk the correct direction on this.
> * the following ones are addressed: # 1, # 2, # 5, # 7, # 8, both narrowing notes. On # 7 I checked the hierarchy, `ParseException`/`FileException`/`GenericException` are bare classes and don't derive from `std::exception`, so that arm only catches allocation failures. `what()` is null-guarded anyway.
> * Windows had two real problems. `RVNGBinaryData.cpp` needs `boost/archive/iterators/*`, so vcpkg installs `boost-serialization` too. `RVNGDirectoryStream.cpp` aHnd `RVNGStreamImplementation.cpp` call `S_ISREG`/`S_ISDIR`, which MSVC lacks, so a force-included header supplies them and upstream sources stay unmodified. Will wait for the next CI run.
> * On # 4/# 9, I want your direction before building it. Vendoring is my preference and it closes # 9 too, but it's ~3 MB of upstream source in the tree and hard to walk back. The alternative is dropping the crate from `publish-crates` plus a `docs.rs` opt-out. Either works as a follow-up, say which.
> 
> also i realized the E2E and Python jobs are red on main too. `xberg-tesseract` wants `cc ^1.4.0` while `e2e/rust/Cargo.lock` pins `1.3.0`, reproduces on a clean `main`. Python is a stale `uv.lock`.

Cool, yes please vendor these.

Goldziher (approved):

PR #1313, opened 2026-07-24
feat(api): add job cancellation endpoint (DELETE /jobs/{job_id})

### LINKED ISSUE #1310: feat: Add job cancellation endpoint (DELETE /jobs/{job_id}) to the API server

Add the ability to cancel a queued or running async extraction job via the HTTP API, e.g.: `DELETE /jobs/{job_id}`

Suggested semantics:

- `pending` job → removed from the queue, state becomes `cancelled`
- `running` job → extraction task is aborted, state becomes `cancelled`
- `completed` / `failed` job → `409 Conflict` (or no-op removal of the stored result)
- unknown/expired job → `404`, as with `GET /jobs/{job_id}`

From a look at the current implementation, this would require two changes beyond the new route:

1. `JobStore` (`crates/xberg/src/api/jobs.rs`) currently only exposes `create` / `get` / `set_running` / `complete` / `fail` — it would need a `cancel` transition and a `cancelled` state.
2. `extract_async_handler` (`crates/xberg/src/api/handlers.rs`) spawns the extraction with `tokio::spawn(...)` and drops the `JoinHandle`, so there is currently no reference that could abort a running job. The handle (or a `CancellationToken` checked between pipeline stages) would need to be retained alongside the job entry.

A cooperative `CancellationToken` checked at stage boundaries would probably be safer than `JoinHandle::abort()`, since extraction stages may hold OCR/model resources.

Currently, once a job is submitted to `/extract-async` there is no way to stop it. This has practical consequences:

- **Abandoned jobs block capacity.** The server returns `429 Too Many Requests` when the concurrent job limit is reached, and jobs a client no longer cares about still count against that limit until they finish. Long-running work (large PDFs with `force_ocr`, transcription, batch inputs) can occupy slots for a long time with no recourse.
- **No recourse for mistakes.** Submitting the wrong file or a wrong config (e.g. accidentally forcing OCR on a 500-page document) means paying the full CPU/GPU cost anyway.
- **Common API convention.** Most async job APIs pair "create job" and "poll job" with "cancel job"; clients (and the MCP tools built on top) generally expect it.

The 5-minute TTL eviction only cleans up *finished* jobs, so it doesn't address any of the above. A cancellation endpoint would make the async API usable for interactive frontends (user hits "stop"), agent workflows with timeouts, and multi-tenant deployments where one runaway job shouldn't starve others.

MannXo:
## Related

closes: #1310

## Description

Adds `DELETE /jobs/{job_id}` to cancel a pending or running async extraction job.

- Pending job: removed from the queue, state becomes `cancelled`.
- Running job: the extraction's existing `CancellationToken` is fired, so the pipeline stops cooperatively at its next checkpoint rather than being force-aborted. This frees the concurrency slot immediately instead of waiting for the job to finish or time out.
- Completed, failed, or already-cancelled jobs return 409 Conflict.
- Unknown or expired jobs return 404, matching `GET /jobs/{job_id}`.

No new dependencies: this reuses the `CancellationToken` / `ExtractionConfig.cancel_token` plumbing already used by the FFI, Python, and Node bindings.

## Checklist

- [x] CI passing

### DIFF
--- a/crates/xberg/src/api/handlers.rs
+++ b/crates/xberg/src/api/handlers.rs
@@ -1077,6 +1077,7 @@ pub(crate) async fn extract_async_handler(
     let mut effective_config = request.config.unwrap_or_else(|| (*state.default_config).clone());
     apply_multipart_config_fields(&mut effective_config, request.output_format, request.pdf_passwords);
     enforce_api_uri_policy(&request.inputs)?;
+    effective_config.cancel_token = state.job_store.cancellation_token(&job_id);
     let inputs = request.inputs;
 
     let job_store = Arc::clone(&state.job_store);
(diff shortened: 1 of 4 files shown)


Goldziher (commented):
## Technical review — job cancellation endpoint

Overall this is a well-shaped change: a dedicated `Cancelled` state, a clean `NotFound`/`Conflict`/`Cancelled` outcome enum mapped to 404/409/200, late-result no-op guards in `complete`/`fail`, the token correctly threaded into `effective_config.cancel_token`, and thorough tests (9 store-level + 3 HTTP-level, all with exact-value assertions). One concurrency bug should be fixed before merge.

### 🔴 Data race: `active` counter double-decrement → underflow

`JobStore::cancel` (`crates/xberg/src/api/jobs.rs`) does a non-atomic `moka` `get` → `insert` and then `active.fetch_sub(1)`. The `DELETE /jobs/{id}` handler runs concurrently with the job's own spawned task calling `complete()`/`fail()`.

If `complete()`'s `get()` reads `Running` **before** `cancel()`'s `insert(Cancelled)` lands, the `Cancelled` guard in `complete`/`fail` is missed and **both paths call `fetch_sub(1)`**. On `AtomicUsize` that second decrement underflow-wraps to `~2^64`, and because `MAX_ACTIVE_JOBS = 100` gates new submissions, the server then **rejects all new async jobs until restart**. The final persisted state is also nondeterministic (last `insert` wins).

This is newly reachable: before this PR, `active` was only ever decremented once, from the single terminal transition inside the job's own task. `cancel()` introduces the first *concurrent* decrement path.

**Suggested fix:** make the state transition + decrement atomic via moka's `entry(job_id).and_compute_with(...)` (available in 0.12), returning the decision, and decrement `active` only on the transition that actually wins. The same treatment applied to `complete`/`fail` closes the window fully.

### 🟡 Minor

- **DELETE idempotency:** a repeat `DELETE` on an already-cancelled job returns **409**, not 200/204. Defensible (it distinguishes "you cancelled it" from "already terminal"), but worth a conscious call versus strict REST idempotency semantics.
- **Token cache** entries are reclaimed only by the 5-min TTL, never explicitly on terminal transition. Bounded by capacity + TTL, so not a leak — fine as-is, just noting.
- **CI is not running on this branch** ("no checks reported"). This endpoint touches concurrent shared state and hasn't been compiled or tested by the pipeline — worth triggering before merge.

Nice work on the test coverage — the late-completion/late-failure no-op tests are exactly the right cases to pin down.

— Claude (technical review)

crates/xberg/src/api/handlers.rs
```diff
@@ -1154,6 +1155,60 @@ pub(crate) async fn job_status_handler(
...
 
+/// Cancel a pending or running async extraction job.
+///
+/// DELETE /jobs/{job_id}
+///
+/// A pending job is removed from the queue; a running job's extraction is
+/// signalled to stop cooperatively at its next checkpoint. Both transition to
+/// `cancelled` and return the updated `JobStatus`. A job that already reached
+/// `completed`, `failed`, or `cancelled` cannot be cancelled again and returns
+/// 409. Jobs expire after 5 minutes and return 404 once evicted, as with
```
Goldziher, line 1166: Please add OTEL and tracing. 

` #[cfg_attr(feature = "otel", tracing::instrument(name = "api.cancel_job", fields(job_id = %job_id,
  outcome = tracing::field::Empty)))]` -- this is the common pattern. 

Otherwise this PR is clean. Good job.

Goldziher:
Please address both comments. This will ship once done and CI is green.

crates/xberg/src/api/handlers.rs
MannXo: reflected in my most recent push

Goldziher (approved):

PR #935, opened 2026-05-11
fix: decode RTF hex escapes with declared ANSI codepage

dergachoff:
## Problem

RTF extraction decodes Cyrillic ANSI RTF files as mojibake when the document declares `\ansicpg1251` and stores text as `\'hh` byte escapes.

Fixes #934.

## Root cause

The RTF parser decoded every hex byte escape through a fixed Windows-1252 mapping. It did not track the document's `\ansicpgNNNN` declaration, so CP1251 bytes such as `cf f0 e8 e2 e5 f2` became `Ïðèâåò` instead of `Привет`.

## Fix

Track the active RTF ANSI codepage per parser group and decode adjacent `\'hh` byte runs with that codepage, preserving Windows-1252 as the default. Added a CP1251 regression test for `\ansicpg1251` + `\fcharset204` Cyrillic byte escapes.

## Validation

- `cargo test -p kreuzberg --lib --features office test_rtf_ansicpg1251_hex_escape_extraction`
- `cargo test -p kreuzberg --lib --features office rtf`

PR #1324, opened 2026-07-27
fix(rtf): decode hex escapes with declared ANSI codepage

### LINKED ISSUE #1323: bug: RTF CP1251 hex escapes decode as mojibake in 1.0 (v4.9 fix never reached main)
## Problem

RTF files declaring `\ansicpg1251` with text stored as `\'hh` byte escapes extract as Windows-1252 mojibake: xberg 1.0.0 returns `Ïðèâåò` where the text is `Привет`.

## Evidence

#934/#935 fixed this, but the PR merged into `chore/v4.9-lts` only (now the LTS repo) and never reached the 1.0 line. `crates/xberg/src/extractors/rtf/parser.rs` still decodes every `\'hh` escape via `decode_windows_1252`, ignoring `\ansicpgNNNN`.

Repro: `{\rtf1\ansi\ansicpg1251\deff0{\fonttbl{\f0\fnil\fcharset204 Arial;}}\f0 \'cf\'f0\'e8\'e2\'e5\'f2}` → `Ïðèâåò`, expected `Привет`.

## Expected fix

Track `\ansicpgNNNN` in the parser and decode `\'hh` escapes with that codepage, defaulting to Windows-1252. Port of #935 incoming.

dergachoff:
> **Status: blocked on xberg-io/test_documents#1.** The fixture-backed tests fail until that merges and the submodule pointer is bumped here.

## Problem

RTF with non-1252 `\ansicpgNNNN` and `\'hh` hex escapes extracts as mojibake: CP1251 bytes `cf f0 e8 e2 e5 f2` become `Ïðèâåò` instead of `Привет`. Fixes #1323.

## Root cause

The parser decodes every hex byte escape through the fixed `decode_windows_1252` mapping and ignores `\ansicpgNNNN`. The original fix (#935) merged into `chore/v4.9-lts` only and never reached the 1.0 line.

## Fix

Port of #935: track the active ANSI codepage per parser group, decode adjacent `\'hh` byte runs with it via `encoding_rs` (existing dependency), keep Windows-1252 default and `\uN`/`\ucN` handling unchanged. Adds a parser unit test plus a fixture-backed extractor regression.

## Validation

- `cargo test -p xberg --lib --features office rtf` — 28 passed (new test fails pre-fix with `Ïðèâåò`)
- `cargo test -p xberg --features office --test rtf_extractor_tests` — 22 passed
- `cargo clippy --locked -p xberg --features full --all-targets -- -D warnings` — clean
- `poly hooks run pre-commit` — green

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- **#1323**: RTF hex byte escapes now honor `\ansicpgNNNN` via the shared Windows-codepage table, so
+  CP1251 Cyrillic and other non-1252 ANSI byte runs decode as readable text instead of Windows-1252
+  mojibake; adjacent escapes decode as one multi-byte run, surviving line wraps, and formatting spans
+  stay aligned with the decoded text.
+
 ## [1.0.0] - 2026-07-27
 
 xberg 1.0.0 is the first stable release of the document-intelligence engine previously developed as
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -11645,6 +11645,7 @@ dependencies = [
  "cfb 0.14.0",
  "chardetng",
  "chrono",
+ "codepage",
  "comrak",
  "crawlberg",
  "criterion 0.8.2",
--- a/crates/xberg/Cargo.toml
+++ b/crates/xberg/Cargo.toml
@@ -827,6 +827,7 @@ calamine = { version = "0.36.1", features = ["dates"], optional = true }
 cfb = { workspace = true, optional = true }
 chardetng = { version = "1.0.0", optional = true }
 chrono = { workspace = true, optional = true }
+codepage = "0.1.2"
 comrak = { workspace = true }
 crawlberg = { workspace = true, optional = true }
 dbase = { workspace = true, optional = true }
--- a/crates/xberg/src/extraction/email.rs
+++ b/crates/xberg/src/extraction/email.rs
@@ -24,12 +24,11 @@
 //! # Ok(())
 //! # }
 //! ```
-use bytes::Bytes;
-use encoding_rs::Encoding;
-
 use crate::error::{Result, XbergError};
 use crate::text::utf8_validation;
+use crate::text::windows_codepage::encoding_for_windows_codepage;
 use crate::types::{EmailAttachment, EmailExtractionResult};
+use bytes::Bytes;
 use mail_parser::MimeHeaders;
 use regex::Regex;
 use std::collections::HashMap;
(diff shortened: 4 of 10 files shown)


tobocop2 (changes requested):
Thanks for porting this over from #935, and for splitting the fixture into xberg-io/test_documents#1. The fix works. With the fixture staged locally, the whole RTF test suite passes on this branch.

### Use the codepage crate for the mapping

The new `encoding_for_windows_codepage` is a copy of a function that already exists in the email extractor, and both hand-roll a table that the [`codepage`](https://crates.io/crates/codepage) crate already provides. That crate is written by the encoding_rs author, it is already in our lockfile through calamine, and it covers 51 code pages against roughly 20 here. Please also point the email extractor's copy at the same shared helper while you're here, since it is the identical function and keeping both invites drift.

Suggested shape:

```rust
u16::try_from(cp).ok()
    .and_then(codepage::to_encoding_no_replacement)
    .unwrap_or(encoding_rs::WINDOWS_1252)
```

Use the no_replacement variant. The plain `to_encoding` maps 50225, 50227, and 52936 to the replacement encoding, and every byte would decode to U+FFFD.

Code pages the current table silently sends to the 1252 fallback, all covered exactly by the crate:

| Code page | Encoding | Affected text |
| --- | --- | --- |
| 874 | windows-874 | Thai |
| 866 | ibm866 | DOS Cyrillic |
| 20866 / 21866 | koi8-r / koi8-u | Cyrillic |
| 10000 | macintosh | Mac Roman |
| 54936 | gb18030 | Simplified Chinese |
| 28593 and up | remaining iso-8859 parts | Arabic, Hebrew, Baltic, others |

If you want to keep the Mac CJK approximations (10001, 10002, 10003, 10008), a small documented pre-match in front of the crate call is fine; the crate returns None for those. Also worth knowing: three arms of the current table don't do what they look like, because encoding_rs folds labels per the WHATWG registry:

- `us-ascii` resolves to windows-1252
- `iso-8859-1` resolves to windows-1252
- `iso-8859-9` resolves to windows-1254

### Formatting offsets drift for multi-byte code pages

The formatting pass still counts every hex escape as one character of output (`extract_rtf_formatting`, the `text_offset += 1` arm), but the extractor now decodes a run of escapes into fewer characters than that under any multi-byte codepage (932, 936, 949, 950, 65001). So bold, italic, and link spans will land in the wrong place for CJK documents. The CP1251 tests can't catch this because single-byte pages still line up one to one. The two passes need to agree on the count for runs.

### Newlines split a byte run

`consume_adjacent_hex_escape` stops at anything that isn't literally the next `\'`, so a line wrap between the two halves of one double-byte character breaks it into two U+FFFD. RTF writers wrap lines freely and readers are supposed to ignore CR and LF, so the lookahead should skip `\r` and `\n`. LibreOffice's hex buffer survives newlines for the same reason.

### Test coverage for the multi-byte path

CP1251 is single-byte, so every test in this PR would pass even if the run coalescing were broken. Could you add one `\ansicpg932` case where two adjacent escapes form a single Shift-JIS character, plus a small table-driven test over a few codepages and their expected output? That would also cover the unknown-codepage fallback.

### Smaller things

- A changelog entry under Unreleased. #935 had one.
- The 1252 fast path decodes the undefined bytes 0x81, 0x8D, 0x8F, 0x90, and 0x9D as question marks, while the encoding_rs fallback turns them into C1 controls, so a declared 1252 and an unknown codepage disagree on the same input. Dropping the fast path and the hand-rolled table removes the divergence.

### Deferred: per-font charsets

Per-font charset handling (`\fcharset`, `\cpg`) still isn't there, so a file that declares `\ansicpg1252` but uses `\fcharset204` fonts keeps garbling. That is fine here since it is a different mechanism, font table parsing plus per-run font state, not more of this codepage plumbing. Tracked in #1326.

dergachoff:
Thanks for the thorough review — all of it should be addressed in 0060fb4. The mapping now goes through `codepage::to_encoding_no_replacement` (with the Mac CJK pre-match) in a helper shared with the MSG extractor, and both hand-rolled tables plus the 1252 fast path are gone. The formatting pass now tracks `\ansicpg` and counts decoded run lengths, the run lookahead skips CR/LF, and there are new tests for the Shift-JIS pair, a wrapped run, a small codepage table with the unknown fallback, plus a changelog entry.

On the fixture newline: the current head of xberg-io/test_documents#1 (9a3e665) already ends with `}\n` — I force-pushed a message-only amend right before your review, so you probably saw the earlier blob.

One heads-up: while checking why CI was green here despite the missing fixture, I found the Rust test script masks cargo-test leg failures — filed as #1327 with details.

tobocop2 (approved):
Thanks for the fixes, this all looks good now.

One nit, not blocking: `codepage` could be `optional = true` behind the `office` and `email` features like the neighboring deps, since the module it feeds is feature-gated.

@dergachoff otherwise this just waits on xberg-io/test_documents#1 and the submodule bump. When you push that, please also rebase on latest main, the branch has fallen slightly behind.

PR #1352, opened 2026-07-30
fix(layout): use CPU for RT-DETR auto acceleration on macOS

### LINKED ISSUE #1344: bug: layout inference fails on CoreML (macOS) and silently falls back to no-layout output
On macOS arm64 with layout enabled, `acceleration.provider="auto"` picks CoreML, the layout model fails at inference (`onnxruntime ExecuteKernel ... CoreMLExecutionProvider` on stderr), and xberg quietly returns plain non-layout markdown. Output is byte-identical to running without layout, `processing_warnings` is empty, no error anywhere.

Forcing `provider="cpu"` fixes it — layout runs and output changes.

Two things wrong here:

1. `auto` picks an EP that can't run the model.
2. When inference fails, the caller gets degraded output with zero signal. At minimum this should land in `processing_warnings`.

Repro (any native-text PDF, macOS arm64, 1.0.3 wheel):

```python
with_layout = await extract({"layout": {}, "use_layout_for_markdown": True, "output_format": "markdown"})
without    = await extract({"output_format": "markdown"})
# with_layout.content == without.content, with_layout.processing_warnings == []
# stderr shows the CoreML ExecuteKernel failure
```

Hit this on a 32-doc eval corpus: all 32 identical to control under `auto`, 30/32 differ under `cpu`.

dergachoff:
## Problem

On macOS arm64, `acceleration.provider="auto"` still selects CoreML for RT-DETR. Session setup succeeds, inference fails with ORT `ExecuteKernel`, and extraction continues without layout. Current main now warns on the markdown path, but the OCR layout path still logs and discards the same failure.

Fixes #1344.

## Root cause

Auto selection checks whether CoreML is available, not whether this RT-DETR export can execute on it.

## Fix

Resolve RT-DETR auto acceleration to CPU on macOS in both file- and byte-backed constructors. Explicit providers and the global `XBERG_ORT_EP` override remain unchanged.

Reuse the shared layout warning for OCR and deduplicate it through the existing diagnostics helper.

## Validation

- Auto extraction completed layout detection with nested and global acceleration config.
- Explicit CoreML reproduced the failure and returned one layout warning from markdown and OCR paths.
- `cargo clippy --locked -p xberg --features full --all-targets -- -D warnings`
- `RUSTFLAGS="-D warnings" cargo check --locked -p xberg-cli --no-default-features --features pdf-ocr`
- `cargo clippy --locked -p xberg-cli --no-default-features --features pdf-ocr -- -D warnings`
- `cargo clippy --locked -p xberg-cli --no-default-features --features all -- -D warnings`
- `cargo test --locked -p xberg-cli --no-default-features --features pdf-ocr feature_profile_tests` (7 passed)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,7 +17,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   language.
 - **#1344**: A hard layout-inference failure (for example a CoreML `ExecuteKernel` runtime error)
   now surfaces a `ProcessingWarning` instead of silently returning byte-identical no-layout output
-  with empty `processing_warnings`; the failure is also logged at `warn` level.
+  with empty `processing_warnings`; the failure is also logged at `warn` level. macOS auto
+  acceleration uses CPU for RT-DETR, whose current export fails under CoreML.
 
 ### Changed
 
--- a/crates/xberg/src/extractors/pdf/layout_runner.rs
+++ b/crates/xberg/src/extractors/pdf/layout_runner.rs
@@ -475,6 +475,15 @@ type LayoutForMarkdownOptional = (
     Option<crate::types::ProcessingWarning>,
 );
 
+pub(super) fn layout_failure_warning(error: &crate::XbergError) -> crate::types::ProcessingWarning {
+    crate::types::ProcessingWarning {
+        source: std::borrow::Cow::Borrowed("layout"),
+        message: std::borrow::Cow::Owned(format!(
+            "layout detection failed ({error}); document extracted without layout hints"
+        )),
+    }
+}
+
 #[cfg(all(feature = "pdf", feature = "layout-detection"))]
 pub(super) async fn maybe_run_layout_for_markdown(
     content: &[u8],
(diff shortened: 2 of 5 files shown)


tobocop2 (changes requested):
Thanks for the fix, it holds up well. Tested on an M1.

What I checked:

- Auto now lands on CPU and layout completes. Explicit CoreML still goes to CoreML and gets the warning.
- `XBERG_ORT_EP` still wins, since the env override is applied later in `apply_execution_providers`.
- All three `RtDetrModel` construction sites go through the new mapping.
- The engine pool matches on config, so auto and explicit CoreML engines stay separate.
- Unit tests pass, and the `cfg(target_os = "macos")` style matches `ort_discovery.rs`.

One build break to fix before merge: `layout_warning` in `run_ocr_with_layout` (`crates/xberg/src/extractors/pdf/mod.rs:335`) is `mut`, but the only mutation sits inside the `cfg(all(feature = "pdf", feature = "layout-detection"))` block. With OCR on and layout detection off that's an `unused_mut`, and CI checks that profile with `-D warnings`:

```
RUSTFLAGS="-D warnings" cargo check --locked -p xberg-cli --no-default-features --features pdf-ocr

error: variable does not need to be mutable
   --> crates/xberg/src/extractors/pdf/mod.rs:335:9
```

The cfg-split that `ocr_layout_gate_decisions` uses a few lines up fixes it.

Tip for next time: for changes near cfg-gated layout/OCR code, `pdf-ocr` is the profile that bites, and the command above is all it takes to check. CI's "Check xberg-cli feature profiles" step runs the full sweep.

dergachoff:
Fixed the cfg split, and the `pdf-ocr` profile now builds cleanly with warnings denied. I also reran its Clippy/profile tests and the layout-enabled `all` profile; everything passes. Thanks for catching that! 🙏

tobocop2 (approved):

tobocop2:
One housekeeping item before merge: both commits are unsigned and this repo requires signed commits. Re-signing is one sweep:

```
git rebase -S origin/main
git push --force-with-lease
```

Issue #1346, opened 2026-07-30
bug: paddle-ocr silently ignores all but the first entry of the language list


dergachoff:
`ocr.language` takes a list and tesseract honors it (`['eng','rus']` reads both scripts). The paddle backend accepts the same list but only uses the first entry — the rest are dropped with no warning.

Repro: mixed Cyrillic/Latin scanned PDF, `language: ['en', 'ru']` → Latin-only output, all Cyrillic gone. Same file with `language: ['ru']` reads both scripts fine (the cyrillic rec model covers Latin chars). 100 chars vs 260.

Root cause: `crates/xberg/src/paddle_ocr/backend.rs:452-454` takes `effective_languages()[0]` and maps only that. Metadata then records the full list joined with `+` (line 550), so the output claims multi-language OCR happened when it didn't. Unknown codes also fall through to `"english"` in the family mapper, again silently.

I get that PP-OCR rec models are per script family, so true multi-script single-pass isn't trivial. But the current behavior is silent data loss. Suggestions:

1. Short term: warn (processing_warnings) when the paddle path gets more than one language, and on the unknown-code fallback.
2. Longer term, the common case (Latin + one non-Latin script) may not need multi-pass at all: the eslav rec model already reads Latin fine in my testing, so `['en','ru']` could collapse to the cyrillic family. Proper multi-family support would need detection once + recognition per requested family.

PR #1359, opened 2026-08-02
feat: xberg doctor — probe configured backends and report what will actually execute

### LINKED ISSUE #1347: feat: xberg doctor — probe configured backends and report what will actually execute
xberg backends degrade quietly: if something can't run, extraction continues with different output. Reasonable at runtime, but it means "configured" and "actually works on this host" diverge invisibly until someone reads the output closely. Recent examples: layout silently degrading when CoreML can't run the model (#1344), paddle silently truncating multi-language lists (#1346), missing tessdata as an easy-to-miss warning, VLM provider errors masked as serialization errors. Those bugs will get fixed; the class will recur with every new backend, model, or execution provider.

Proposal: a `doctor` probe — per configured backend, report pass / fail / skip with a one-line reason. Lives in the core, exposed per binding (`xberg.doctor()` in Python) and as an `xberg doctor` CLI subcommand over the same call, so wheel users get identical checks.

Shape: a `probe()` hook on the backend trait — each backend self-reports health, new backends and plugins get diagnostics by implementing it. Plus a config-lint pass for silently-truncated settings (e.g. language list on paddle).

What the hook would check today:

- layout: one inference on a tiny input per configured execution provider
- tesseract: binary + tessdata for each configured language
- paddle: model cache status per script family (reports "will download on first use", doesn't download)
- VLM: API key present, endpoint reachable (no billable test call)

This complements runtime warnings, doesn't replace them: bugs like #1344 should still land in `processing_warnings`. Doctor is for setup and triage — catch environment problems before the first document, answer "is it my document or my environment?" in one command.

Non-goals: no downloads, no quality scoring, no benchmarking, no auto-fix. JSON output for bug reports.

dergachoff:
Fixes #1347.

Adds `xberg doctor` — a preflight check for the "configured but won't actually run" class: CoreML layout silently degrading (#1344), missing tessdata, paddle language lists truncating (#1346). Each check is pass / fail / skip with a one-line reason. The CLI exits nonzero on any failure. Nothing gets downloaded, nothing billable is called.

## How it works

Core API is `doctor(&ExtractionConfig) -> DoctorReport`. Backends self-report through a new `OcrBackend::probe` hook — defaults to skip and is `alef(skip)`, so custom backends and the 15 binding surfaces need no changes.

- tesseract: resolves tessdata per configured language through the same search chain the runtime uses (extracted into a shared `tessdata_search_dirs`). Unknown lang code fails; known-but-missing skips with "will download on first use".
- paddle: checks the pinned SHA-256 of every artifact the configured version/family/tier would need. No downloads (`ModelManager::check_models_cached`).
- vlm: builds the liter-llm client (provider + env key resolution), then an unauthenticated GET to the endpoint. Any HTTP status counts as reachable; no key sent, no key in the report.
- layout: one real RT-DETR inference on a synthetic page — this is what catches the #1344 class. Mirrors the runner's auto→CPU retry, so hosts that degrade to CPU report pass-with-warning instead of failing.
- config lint: `vlm_fallback` without `vlm_config`; paddle language lists the selected model can't read (#1346, statically).
- `force_ocr` with no ocr config probes the default backend, same as the runtime builds.

`--clean` removes stray regular files from the xberg cache root, and only when the directory actually is an xberg cache (contains a known module dir) — a repointed `XBERG_CACHE_DIR` can't lose unrelated files. The shared HF cache is report-only.

`doctor()` is alef-discoverable: `alef generate` (0.49.0) emits it for all 15 targets, incl. Python `xberg.doctor()`. Generated files left out on purpose, per the regen-chore convention.

## Validation

- clippy `-D warnings`: `full`, `layout-tract`, `ocr,auto-rotate-tract`; xberg-cli default + `all`/`pdf-ocr`/`core-cli`/`paddle-ocr`/`candle-ocr`/`liter-llm` profiles
- check: no-default, `ocr-wasm`, wasm32 `wasm-target`
- unit tests: tesseract probe, paddle cache check, `--clean` guards (incl. symlink and non-xberg-dir refusal), lint, serde; full xberg-cli suite
- live: tesseract pass, vlm key/endpoint checks, real RT-DETR inference, `--clean` both ways, `force_ocr`-only config

Heads up, pre-existing: `xberg-cli --features pdf-heuristic` clippy fails on main (dead code in `extractors/pdf/ocr.rs` and `pdf/oxide/images.rs`, untouched here).


dergachoff:
@siddhanttiwari19 since you volunteered for pr — review welcome when and if you have time

crates/xberg/src/doctor/cache.rs
```diff
@@ -0,0 +1,235 @@
...
+    for entry in entries.flatten() {
+        if entry.file_type().is_ok_and(|t| t.is_file()) {
+            stray.push(entry.path());
+        }
+    }
+    stray
+}
+
+/// Top-level cache entries that are not known module dirs. Report-only.
+fn unknown_dirs(base: &Path) -> Vec<String> {
```
siddhanttiwari19, line 182: unknown_dirs doesn't have the same "is this actually an xberg cache" guard that stray_files has above it. If XBERG_CACHE_DIR is misconfigured to point at something like $HOME, stray_files correctly bails out (per the comment above it), but unknown_dirs will walk that directory and report every top-level folder name in the doctor output.

That seems worth fixing given xberg doctor --format json is documented as the thing to paste into bug reports — this would leak a directory listing of whatever got misconfigured as the cache root.

Suggest adding the same KNOWN_MODULE_DIRS.iter().any(|dir| base.join(dir).is_dir()) early return here, plus a test mirroring stray_files_empty_when_base_is_not_an_xberg_cache

crates/xberg/Cargo.toml
```diff
@@ -91,7 +91,7 @@ xml = ["dep:quick-xml", "dep:roxmltree"]
 archives = ["dep:zip", "dep:tar", "dep:sevenz-rust2", "dep:flate2"]
 mdx = []
 
-liter-llm = ["dep:liter-llm", "dep:minijinja", "ocr-pipeline"]
```
siddhanttiwari19, line 94: Quick check before approving does ureq here have a TLS backend enabled (rustls/native-tls feature)? The VLM probe does an unauthenticated GET that's likely https:// most of the time; if TLS isn't on, that'd fail as a scheme error rather than a real "endpoint unreachable" signal. If ureq's already pulled in with TLS elsewhere in the workspace this is probably a non-issue, just flagging.

siddhanttiwari19 (commented):
Nice design overall — the probe-mirrors-runtime approach (tessdata search chain, paddle SHA check, layout CPU-retry mirror) is well thought out, and the --clean cache-root guard on stray_files is careful about not deleting outside an actual xberg cache.

One real gap, left as an inline comment on cache.rs: unknown_dirs doesn't have the same "is this actually an xberg cache" guard that stray_files has right above it. If XBERG_CACHE_DIR is ever misconfigured to point somewhere like $HOME, this would list real folder names in the doctor report — which matters since --format json is documented as the thing to paste into bug reports.

Two smaller non-blocking notes also left inline: the stray-files/unknown-dirs/HF-stale checks render as green "pass" even when reporting something actionable (might read better as Skip), and a quick sanity check on whether ureq's TLS backend is enabled for the VLM probe's https:// GET.

Happy to re-review once the cache-root guard is fixed — everything else here looks solid.

Goldziher (approved):
nice work

dergachoff:
@Goldziher @siddhanttiwari19 pushed a follow-up.

Went a step further than the `unknown_dirs` guard: name-based checks can't really prove a dir is ours (`extraction/`, `layout/`, or `tessdata/` can all exist in normal user dirs), so doctor now just never inspects or cleans an `XBERG_CACHE_DIR` override — it reports a skip saying why. The default cache path is inspected as before. A marker file written at cache creation could bring override support back — can do as a follow-up.

The green-pass point was fair, but these checks do run, so instead of `skip` I added a `warn` status: yellow, doesn't affect the exit code.

Also fixed while re-testing: the HF stale-revision check mixed up pinned revisions between `layout-models` and `paddleocr-onnx-models` — now matched per repo. TLS is fine, ureq has rustls wherever the VLM probe compiles. Rebased on main for the #1366 CI fixes. Re-review welcome.

Goldziher:
> @Goldziher @siddhanttiwari19 pushed a follow-up.
> 
> Went a step further than the `unknown_dirs` guard: name-based checks can't really prove a dir is ours (`extraction/`, `layout/`, or `tessdata/` can all exist in normal user dirs), so doctor now just never inspects or cleans an `XBERG_CACHE_DIR` override — it reports a skip saying why. The default cache path is inspected as before. A marker file written at cache creation could bring override support back — can do as a follow-up.
> 
> The green-pass point was fair, but these checks do run, so instead of `skip` I added a `warn` status: yellow, doesn't affect the exit code.
> 
> Also fixed while re-testing: the HF stale-revision check mixed up pinned revisions between `layout-models` and `paddleocr-onnx-models` — now matched per repo. TLS is fine, ureq has rustls wherever the VLM probe compiles. Rebased on main for the #1366 CI fixes. Re-review welcome.

awesome, nice.

Issue #1220, opened 2026-07-06
bug: three PDF table-extraction defects from a car-manual audit (footer-as-table, rotated charts, cell-text fragmentation)


tobocop2:
Auditing PDF extraction quality end-to-end on one of my most common verification documents — my car owner's manual (2011 Ford Crown Victoria, 334 pages; gist: https://gist.github.com/tobocop2/0d30905e2375e09b153ce50296316f91, `cv-manual.pdf`) — surfaced three distinct table-extraction defects beyond #1213. None of them is specific to this manual; each is a generic failure mode that this document happens to exercise all at once. Filing them together so they can be reviewed (and fixed) as one unit.

## 1. Heuristic table tier emits the running page footer as a table on ~300 of 333 pages

Any document with a repeating margin footer hits this. The manual's three-line running footer:

```
2011 Crown Victoria (cro)
Owners Guide, 3rd Printing
USA (fus)
```

is a vertically isolated slab of short aligned text, so the heuristic tier's region clustering accepts it as a table candidate on every page where no real table wins. Default-config extraction yields **377 tables, 300 of which are this footer**:

```
| 2011 | Crown | Victoria | (cro) |
| --- | --- | --- | --- |
| Owners | Guide, | 3rd | Printing |
| USA | (fus) |  |  |
```

`result.tables` is ~80% page furniture. Repro:

```python
res = await xberg.extract(xberg.ExtractInput(kind="uri", uri="cv-manual.pdf"))
doc = res.results[0]
footer = [t for t in doc.tables if "Owners" in t.markdown and "Printing" in t.markdown]
print(f"{len(footer)} footer tables of {len(doc.tables)}")   # ~300 of 377; expected 0
```

The pipeline already has header/footer awareness in the structured path; the heuristic table tier doesn't consult it. A cross-page repetition signature (same normalized candidate text at the same position on 3+ pages) would also catch it without any layout knowledge.

## 2. Rotated (90°) full-page tables extract as word salad

Any landscape table typeset on a portrait page hits this. PDF pages 312–313 are the scheduled-maintenance charts: fully ruled grids rotated 90° (service items read bottom-to-top, mileage intervals as columns, bullet-dot cells). The text spans carry the rotation, but reading order runs in portrait coordinates, so the output interleaves fragments of every row:

```
Whichever * if and tread abnormal (if brake for for lubricate level suspension,
measure drums, shields dealer fluid (recommended) concentration and joints, heat
[...]

PR #1363, opened 2026-08-02
fix(pdf): stop dropping CFF glyphs that use the deprecated dotsection operator

### LINKED ISSUE #1362: bug: rendered pages silently drop every i, j, period and colon when a PDF embeds CFF fonts that use dotsection
## What happens

Any page rasterized through pdf_oxide loses every glyph whose CFF charstring contains the deprecated `dotsection` operator (`12 0`). The glyph is not painted but the cursor still advances, so output is gap-filled instead of visibly broken: "Owners Guide, 3rd Printing" renders as "Owners Gu de, 3rd Pr nt ng". Text-layer extraction is unaffected. The damage is confined to the raster path, which means OCR: every backend (tesseract, paddle, custom plugins) receives images with the letters already missing and transcribes the gaps faithfully. "capacities" becomes "capac t es", so an OCR'd page is unfindable in a downstream index while still looking populated.

Distiller-produced Type 1C subsets carry `dotsection` routinely: Adobe's Type 1 to Type 2 conversion preserves it on dot-bearing glyphs.

## Reproduce

`cv-manual.pdf` from the #1220 gist (https://gist.github.com/tobocop2/0d30905e2375e09b153ce50296316f91), every font is Type 1C:

```python
cfg = ExtractionConfig(ocr=OcrConfig(backend="tesseract"), force_ocr=True)
# "Service refill capacities are determined by filling"
# comes back as
# "Serv ce ref ll capac t es are determ ned by f ll ng"
```

`pdftoppm -f 5 -l 5 -r 150 -png cv-manual.pdf poppler` paints every glyph at the same DPI, so it is not the input.

## Evidence

Eight pages rendered through the production backend path at 150 dpi, OCR'd with tesseract, compared against the PDF text layer. Occurrences per 1000 characters:

| char | text layer | render + OCR | ratio |
|------|-----------:|-------------:|------:|
| i | 67.9 | 0.2 | ~0 |
| . | 29.4 | 0.4 | ~0 |
| : | 2.6 | 0.1 | ~0 |
| j ; ! ? * | present | 0 | 0 |
| l | 1.10 |
| n | 1.13 |
| c | 1.04 |
| , | 1.18 |

The set of glyphs carrying `dotsection` and the set of missing glyphs are identical (Century-Light 11 of 95 glyphs, Helvetica-Bold 8 of 81, Century-Bold 9 of 82). Ruled out: DPI, font substitution (shapes and metrics match poppler exactly), TrueType (a Monaco-embedded PDF renders `i j . : !` fine through the same path). Identical on the 1.0.4 and 1.0.8 wheels.

## Root cause

ttf-parser 0.25.1 handles only the flex family (escape opcodes 34-37) in its two-byte operator match (`src/tables/cff/cff1.rs`) and returns `CFFError::UnsupportedOperator` for everything else, aborting the entire charstring: the glyph yields no outline at all. pdf_oxide wraps the bare `/FontFile3` in a synthetic OTTO shell (`wrap_cff_in_opentype`) and asks ttf-parser for outlines; `render_cid_direct` treats a `None` outline as nothing to paint, with no log line, and advances the cursor anyway.

`dotsection` takes no arguments and is purely a hinting hint, so an outline consumer should clear the argument stack and continue. read-fonts does exactly this in `src/ps/cs.rs` ("Brackets an outline section for dots in letters such as i, j and !. Purely metadata that a hinter can use."). The fix is one line in ttf-parser's two-byte match plus a regression test; control glyphs keep byte-identical outlines, so byte-identical render gates on unaffected corpora stay green.

## Possible xberg-side handling

Fixing this upstream is the priority: the one-line ttf-parser fix, and separately a pdf_oxide `warn` when an outline comes back `None` (the silent swallow is what let this reach a corpus unnoticed). If upstream releases do not pick these up soon, a `[patch.crates-io]` entry pointing at a patched 0.25.1 is a reasonable interim that ships a fixed wheel immediately.

tobocop2:
> **Superseded by #1384**, which vendors ttf-parser into the workspace as `xberg-ttf-parser` with the `dotsection` fix already applied. That removes the wait on an upstream release entirely, so the `[patch.crates-io]` git pin and the deny.toml allow-git entry in this PR are no longer the route to shipping the fix. The regression test here has been carried over to #1384. Previously blocked on harfbuzz/ttf-parser#228 (issue harfbuzz/ttf-parser#227), which remains unreviewed upstream.

## Problem

Pages rasterized through pdf_oxide silently lose every `i`, `j`, `.`, `:`, `;`, `!` and `?` when the page uses a Type 1C (CFF) font whose charstrings carry the deprecated `dotsection` operator, which distiller-produced subsets do routinely. The cursor still advances, so OCR backends receive gap-filled images and transcribe "capacities" as "capac t es". Text-layer extraction is unaffected. Fixes #1362.

## Root cause

ttf-parser 0.25.1 rejects `dotsection` (escape `12 0`) with `UnsupportedOperator`, aborting the entire charstring, so the glyph yields no outline at all.

## Fix

`[patch.crates-io]` pins a patched 0.25.1 (the harfbuzz/ttf-parser#228 change on a 0.25.1 base), plus the matching deny.toml allow-git entry. The pin covers repo-built artifacts (wheels, packages, prebuilt CLI, Docker); `cargo install` from crates.io picks the fix up via the ttf-parser release.

## Regression test

Single-page fixture embedding a synthetic fontTools-generated Type 1C font whose dot-bearing glyphs carry `dotsection` the way converted Type 1 subsets do (no third-party font data), rendered through the production path, asserting ink per glyph cell. Red on stock ttf-parser (`i` renders 0 dark pixels), green with the fix; control glyphs (`l n ,`) guard the fixture itself.

## Validation

Gates on a clean ubuntu-24.04 x86_64 box mirroring the CI leg (install-linux.sh, tesseract 5, rust 1.95):

| check | result |
|---|---|
| clippy: workspace, full-ORT surface, both tract legs (`-D warnings`) | green |
| `cargo test -p xberg --features full --all-targets` | 161 test binaries green; the 17 failing tests (paddle cache-dir, typst suite, pdf quality gates) fail identically at base `8d6647fe` with stock ttf-parser, so pre-existing (#1365) |
| `cargo check` no-default-features and ocr-wasm `--lib` | green |
| dotsection regression test | fails on stock ttf-parser (0 ink for `i`), passes with the pin |

Eight pages of the #1220 manual through the production render path, tesseract vs the PDF text layer, occurrences per 1000 chars:

| char | layer | OCR | ratio | before |
|------|------:|----:|------:|-------:|
| i | 76.0 | 75.9 | 1.00 | ~0 |
| . | 9.4 | 9.7 | 1.03 | ~0 |
| j : ; ? | 2.6 | 2.6 | 1.00 | 0 |
| l n c , (controls) | 146.8 | 146.4 | 1.00 | 1.00 |

## Not covered

pdf_oxide still swallows `None` outlines with no log (upstream issue to follow); ttf-parser rejects sibling operators (`div`, reserved opcodes) the same way, tracked for upstreaming separately.


tobocop2:
Superseded by #1384

PR #1388, opened 2026-08-05
feat(doctags): add Docling DocTags rendering and ingestion

### LINKED ISSUE #1383: feat: Add support for Docling DocTags
### What

Is there a version of Xberg that allows access to DocTags via the Python API?

If not, is this planned for a future version?
Thanks in advance!

### Why

It's useful for switching between formats...

MannXo:
## Related

Closes #1383

## Description

Adds DocTags support in both directions: xberg can emit it as an output format and read it as an input format, with tables carried as OTSL and geometry as `<loc_*>` tokens.

```python
import xberg
# emit
config = xberg.ExtractionConfig(output_format=xberg.OutputFormat("doctags"))
# ingest
result = await xberg.extract(xberg.ExtractInput.from_bytes(data, "text/vnd.docling.doctags"))
```

### Emit

`rendering/doctags.rs` serialises the internal model. Elements map onto the DocTags vocabulary (`Heading{level}` to `section_header_level_N`, `ContentLayer::Header`/`Footer`/`Footnote` to `page_header`/`page_footer`/`footnote`, `Caption` relationships nested inside `<otsl>`/`<picture>`/`<code>`, list items wrapped in `<ordered_list>`/`<unordered_list>`).

Location tokens are emitted when an element has a bounding box and its page recorded dimensions. `BoundingBox` is PDF space with a bottom-left origin and DocTags counts from the top-left, so the vertical axis is flipped. Checked against the vendored ground truth for `2305.03393v1-pg9`: x tokens match exactly (110, 393) and y lands within about two grid units on blocks that line up. The remaining difference is our block segmentation, not the geometry.

This is also why PPTX is excluded. `extraction::pptx::position_to_bbox` fills `y0` with the *top* edge, since DrawingML's origin is top-left, so its boxes do not follow the `BoundingBox` contract that `y0` is the bottom. Rather than special-casing it, emission is gated on the page having recorded dimensions, which PPTX never does. Worth fixing separately, and it is visible today through `ExtractedImage.bounding_box`.

### Ingest

`extraction/doctags.rs` parses a stream back into an `InternalDocument`, with an extractor registered for `text/vnd.docling.doctags` and a `mime.rs` entry for `.doctags`. Two properties of the format shaped it:

- **No escaping.** Real output carries literal `<` in prose: a caption in `2203.01017v2` discusses `' < td > '`. Scanning to the next `>` swallows the `</caption>` that follows and the document stops being well formed from there on. So only recognised tag names count as tags and everything else is content, which is how docling's own tokenizer behaves.
- **Location tokens are page relative** and the page size is not in the stream. Pages are rebuilt as 500-unit squares, which makes re-emitting a parsed document reproduce the original tokens exactly.

OTSL merge tokens are expanded into the flat grid: `lcel` repeats the cell to its left, `ucel` the cell above, `xcel` either.

### One change outside the DocTags files

`needs_structured` in `extractors/pdf/extraction.rs` matched only `Markdown | Djot | Html`, so `OutputFormat::Custom` took the flat path and its elements had no headings and no bounding boxes at all. That made location tokens inert on real PDFs. Adding `Custom(_)` fixes it, and it is the same class of gap in `extractors/epub/mod.rs:126,229` and `extractors/docx.rs:570`, which I have left alone as out of scope.

On your earlier question about a first-class `OutputFormat::DocTags` variant: I could not do it. `cargo install alef-cli` gives 0.17.34, which cannot parse this repo's `alef.toml` (`vendor_mode = "registry"` is not a variant it knows), so I have no way to regenerate the binding crates or verify the result. Adding the variant without regenerating would leave five generated crates inconsistent with the core enum. The registry route needs no regeneration and reaches every binding, since an unrecognised format string already resolves to `Custom` and `derive_extraction_result` routes it through the `RendererRegistry`. Happy to switch if you regenerate, the renderer itself does not change.

## Checklist

- [ ] CI passing

### Testing

`cargo test -p xberg --lib`: **1232 passed, 0 failed** (1173 before). 60 of those are DocTags tests.

Round trip, which is the property that matters here:

- `render(parse(x)) == x` exactly, for a stream covering headings, lists, OTSL with a caption, formulas, page breaks and footers.
- `parse` then `render` twice is idempotent.
- Every vendored Docling file is parsed, and re-emitting preserves the element count. This is what caught a cell-less `<otsl>` in `2203.01017v2` that the parser was turning into an empty table.

Structural validation runs in both directions, over our output and over the vendored corpus, so if our reading of the format drifts from docling's the corpus side fails rather than our tests agreeing with themselves. It checks the tag vocabulary, `<doctag>` nesting, location tokens on the 0 to 500 grid with left <= right and top <= bottom, and rectangular OTSL rows. Vocabulary and OTSL are skipped on the corpus because `2305.03393v1` and `redp5110_sampled` still carry legacy `<table>`/`<tr>`/`<td>` markup.

Validating against real output corrected two of my assumptions: that the format could be tokenized by scanning to the next `>`, and that `checkbox_*` were standalone rather than wrapping their label.

`cargo clippy --locked -p xberg --all-targets -- -D warnings` is clean, as is `rustfmt` on the touched files.

Two things I could not run locally. CI's `--features full` leg builds tesseract from source and pulls candle/ONNX. And `cargo test -p xberg --features pdf --lib` reports 18 failures both with and without this branch, because the `test_documents` fixtures are unfetched LFS pointers here.

### Known limitation

`Table::cells` is a flat `Vec<Vec<String>>` with no span data, so emitted OTSL can never produce `lcel`/`ucel`/`xcel`/`rhed`, and merged cells come through duplicated. Ingest parses the full grammar and expands it, so only the emit direction is lossy. Fixing it means adding spans to the table model, which affects more than DocTags.


MannXo:
One thing worth flagging before I take location tokens further.

I have the `<loc_*>` mapping working and checked against the vendored Docling ground truth for `2305.03393v1-pg9`. The x tokens match exactly (110, 393) and y lands within about two grid units on blocks that line up, so the normalisation and the vertical flip are right.

The problem is that it is inert in practice. Element `page` and `bbox` only get populated when the PDF extractor takes its structured path, and that path is chosen by matching on the output format:

```rust
// extractors/pdf/extraction.rs
let needs_structured = hierarchy_enabled
    || matches!(config.output_format,
        OutputFormat::Markdown | OutputFormat::Djot | OutputFormat::Html)
    || ocr_inline_images;
```

`Custom("doctags")` is not in that list, so DocTags gets the flat path with no geometry at all. I only saw tokens by forcing `pdf_options.hierarchy.enabled`. It also explains something I mentioned in the description above: headings coming out as `<text>` instead of `section_header_level_*` is this gate, not a missing feature flag, so I was wrong about the cause there.

It is not only PDF. `extractors/epub/mod.rs:126` and `:229` gate `wants_markup` the same way, and `extractors/docx.rs:570` checks for `Markdown` specifically. In each case `Custom` takes the degraded branch.

So the registry-only approach in this PR has a real cost rather than just a smaller diff: DocTags ends up with poorer input than Markdown gets, which is backwards for a format that is mostly about geometry and structure.

Two ways I can see:

1. Add a first-class `OutputFormat::DocTags` variant. Correct by construction and it picks up every existing gate for free, but it means regenerating the alef binding crates, so a much larger diff.
2. Broaden those gates to include `Custom(_)`. Close to a one line change each, and arguably more correct in general since any non-plain renderer wants structure. The downside is that it quietly turns structured extraction on for every third party custom renderer, including ones that would rather not pay for it.

I lean towards 1 for correctness, but 2 is yours to weigh since the cost lands on other people's renderers. Happy to do either, and location tokens are ready to go once it is settled.

Separately, and only if you think it is worth its own issue: `Table::cells` is a flat `Vec<Vec<String>>` with no span data, so emitted OTSL can never produce `lcel`/`ucel`/`xcel`/`rhed`. Merged cells come through duplicated, which you can see in the `docx_tables.docx` output in the description. That limit is not specific to DocTags, it would also affect rowspan-faithful HTML.

Goldziher (commented):
This looks good @MannXo . 

1. you decided to split rendering and ingestion into two different PRs? I assume so. 
2. we need a round-trip test. That is, we need to test how what is outputted by xberg compares with what is outputted by docling, and how docling-core parses what we output.

MannXo:
Thanks. Answers below.

**1.** Yes, though three PRs rather than two if that is alright? this one for the renderer, a second for the `<loc_*>` tokens, and a third for ingest. The loc token work is already written and checked locally, it is just waiting on the question in point 3.

**2.** Agreed on the round trip, with one caveat worth settling before I build it. A byte equality test against docling's output will not pass, and I do not think it should be the target. Our layout analysis differs from theirs: on `2305.03393v1-pg9` we merge the page header into the following paragraph and classify some section headers as plain text. The geometry itself lines up well, x tokens match exactly and y lands within about two grid units, so the divergence is segmentation rather than serialisation.

So I would suggest three layers:

- Parseability. Feed our output to `docling-core` and assert it loads into a `DoclingDocument` without error. This is the one that actually answers "can docling consume what we emit".
- Structural invariants, in Rust. Every tag is in the DocTags vocabulary, `<otsl>` rows are rectangular, loc tokens are within 0 to 500 and correctly ordered, `<doctag>` balanced. Cheap and runs everywhere with no Python.
- Comparison against the vendored corpus as a reported metric rather than a hard assert. Element counts, tag distribution, bbox overlap. That way regressions surface without pinning us to docling's segmentation choices.

One thing I could use direction on: where should the `docling-core` test live? `e2e/python/tests` is entirely alef generated, so anything hand written there is wiped on the next regen, and I would rather not put a new Python dependency in the wrong place. Happy to add it wherever you prefer.

**3.** Still open from my previous comment, and it is what blocks the loc tokens. Do you want `OutputFormat::DocTags` as a real variant, or should I broaden the `Custom(_)` gates instead? As it stands `Custom` misses `needs_structured`, so DocTags gets no geometry at all and the tokens are inert on real PDFs. Either approach is fine by me, I would just rather not guess and regenerate the bindings for nothing.

Unrelated to this PR, but you may want to know. CI is currently failing at checkout for everything, including main. The `test_documents` submodule pointer on main is `05a9214b`, which no longer exists in that repo, so `git submodule update --init` aborts before anything builds. The current tip there is `6140cdb8`. Happy to send a one line bump as a separate PR if that helps.

MannXo:
I made a start on the structural layer while waiting, and running it against the vendored corpus turned up two things about the format that are worth recording, since both matter for ingest.

First, DocTags genuinely cannot be parsed by scanning for the next `>`. There is no escaping, and real output carries literal `<` in prose: the caption in `2203.01017v2` discusses `' < td > '` and `' < '`, so a naive scan swallows the `</caption>` that follows it and the document stops being well formed from there on. A parser has to recognise known tag names only and treat everything else as text. That is also the argument for not escaping on the way out, which is what docling does.

Second, `checkbox_selected` and `checkbox_unselected` wrap their label rather than standing alone, for example `<checkbox_unselected><loc_193><loc_141><loc_208><loc_149>بلی</checkbox_unselected>` in `right_to_left_03`. I had them down as standalone tokens to begin with.

The checks cover the tag vocabulary, the `<doctag>` wrapper and nesting, location tokens on the 0 to 500 grid with left <= right and top <= bottom, and rectangular OTSL rows. They run in both directions, over our output and over the vendored corpus, so if our reading of the format drifts from docling's then the corpus side fails rather than ours quietly agreeing with itself. Vocabulary and OTSL are skipped on the corpus because `2305.03393v1` and `redp5110_sampled` still carry legacy `<table>`/`<tr>`/`<td>` markup.

That is layer 2 of the three I sketched. Layer 1 needs your steer on where a `docling-core` dependency should live, and layer 3 is straightforward once layer 1 has a home.

Happy to push these now if you want them on this branch, or to hold them if the variant question means this PR gets reworked. Just say which.

Goldziher:
@MannXo - let roll into this PR also the ingestion side (i.e. reading it as input, extractor layer etc.). and the tests. It doesnt need to be byte identical, but it needs to be inter-operable with docling, and correct.

MannXo:
Ingest is in, along with the round-trip tests. Description updated, and the PR now closes the issue rather than being the first of three.

**Round trip.** The property I went for is `render(parse(x)) == x` exactly, and it holds for a stream covering headings, lists, OTSL with a caption, formulas, page breaks and footers. Parse then render is also idempotent. That works because a parsed document rebuilds its pages as 500-unit squares: the real page size is not recoverable from the stream, so using the grid itself as the page makes the tokens survive a round trip unchanged.

**Against docling.** Every vendored file is parsed in tests, and re-emitting preserves the element count. That check earned its place immediately: it caught a cell-less `<otsl>` in `2203.01017v2`, location tokens and nothing else, which the parser was turning into an empty table that then vanished on re-emit. It now drops it, since there is no table there.

Structural validation runs over our output and over the corpus, so if our reading of the format drifts from docling's then the corpus side fails rather than our own tests agreeing with themselves. Two of my assumptions did not survive that: the format cannot be tokenized by scanning to the next `>`, and `checkbox_*` wrap their label rather than standing alone.

**On the variant question.** I tried to do it properly and could not. `cargo install alef-cli` gives 0.17.34, which cannot parse this repo's `alef.toml`, since `vendor_mode = "registry"` is not a variant it knows, and `cargo install alef` does not give a newer binary either. So I have no way to regenerate the binding crates or check the result, and adding the variant without regenerating would leave five generated crates inconsistent with the core enum. I went with the registry route and broadened the `needs_structured` gate to include `Custom(_)` instead. If you would rather have the variant and can regenerate, the renderer itself does not change.

**One change outside the DocTags files**, worth a look since it affects more than this feature. `needs_structured` in `extractors/pdf/extraction.rs` matched only `Markdown | Djot | Html`, so `Custom` renderers took the flat path and their elements had no headings and no bounding boxes, which made location tokens inert on real PDFs. Adding `Custom(_)` fixes that, but it does mean any third party custom renderer now gets structured extraction and pays for it. Same gap exists in `extractors/epub/mod.rs:126,229` and `extractors/docx.rs:570`, left alone as out of scope. Say the word if you want it done differently.

`cargo test -p xberg --lib` is 1232 passing, 60 of them DocTags. Clippy and rustfmt clean.

Two things I could not run here: the `--features full` leg, which builds tesseract from source, and the `--features pdf` tests, which report 18 failures both with and without this branch because the `test_documents` fixtures are unfetched LFS pointers on my side.

Unrelated but still worth a look, CI is red at checkout on every branch including main, because the `test_documents` submodule pointer is `05a9214b` and that commit no longer exists in the repo. Current tip is `6140cdb8`.

Issue #300, opened 2026-01-15
bug: Embeddings feature is not available in full Docker image


rsass-qulto:
Using the `latest` (currently equivalent to `4.0.6`) “full” Docker image in server mode, when I call the `/embed` endpoint with a JSON body containing a `texts` array field (so the request should be assembled correctly), I receive an HTTP 500 response with the following body:

```
{
  "error_type": "MissingDependencyError",
  "message": "Missing dependency: Embeddings feature is not enabled. Rebuild with --features embeddings",
  "status_code": 500
}
```

Since the Dockerfile already includes ONNX Runtime installation steps, I would expect embeddings to work out of the box. Therefore, I believe this is a bug.

Goldziher:
Bug, thanks.

PR #1409, opened 2026-08-08
fix(rendering): handle comment element kinds in the DocTags renderer

MannXo:
## Related

Blocks every other PR. `main` does not compile at `425b81f5f`.

## Description

`chore: full regen` added `ElementKind::CommentDefinition` and
`ElementKind::CommentRef` without updating the DocTags renderer, so the crate
fails to build on any feature set:

```
error[E0004]: non-exhaustive patterns: `ElementKind::CommentDefinition` and
`ElementKind::CommentRef` not covered
  --> crates/xberg/src/rendering/doctags.rs:77:15
```

Reproduced on `main` with plain `cargo check -p xberg`, and with
`--no-default-features` plus `xml,svg`. Every renderer other than DocTags
already handles both variants.

References are dropped, matching how the same match already drops
`FootnoteRef`. Definitions are keyed by anchor and emitted as a footnote rather
than dropped, because DocTags has no comment tag and the Djot renderer went out
of its way to surface the same element rather than lose the body (#300). Say the
word if you would rather they were dropped outright and I will cut it down.

Two tests added alongside the existing footnote pair.

```
cargo test -p xberg --lib rendering::doctags
test result: ok. 48 passed; 0 failed
```

## Checklist

- [x] CI passing

### DIFF
--- a/crates/xberg/src/rendering/doctags.rs
+++ b/crates/xberg/src/rendering/doctags.rs
@@ -76,7 +76,7 @@ pub(crate) fn render_doctags(doc: &InternalDocument) -> String {
 
         match elem.kind {
             ElementKind::QuoteStart | ElementKind::QuoteEnd | ElementKind::GroupStart | ElementKind::GroupEnd => {}
-            ElementKind::FootnoteRef => {}
+            ElementKind::FootnoteRef | ElementKind::CommentRef => {}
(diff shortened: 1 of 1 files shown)


MannXo:
closing this. another fix #1408 already merged

Issue #579, opened 2026-03-25
feat: Export Chart Relationships to Graphviz DOT Format


mits87:
### Summary

Add support for extracting chart/diagram structures (nodes, edges, labels, and styling) into the [Graphviz DOT format](https://graphviz.org/doc/info/lang.html)￼ - [graphiz_rust](https://docs.rs/graphviz-rust/latest/graphviz_rust/). 

This would allow users to programmatically reconstruct diagrams, visualize relationships, and integrate with graph-based tooling.

### Problem

Currently, when processing documents with diagrams (e.g. org charts, flowcharts, architecture diagrams), the extracted output lacks a structured representation of relationships between elements.

This makes it difficult to:
* Reconstruct diagrams programmatically
* Analyze relationships between entities
* Visualize or export diagrams into other systems
* Perform graph-based reasoning or transformations

### Proposed Solution

Extend the extraction pipeline to detect and output diagram structures in DOT format.

The system should:

1. Detect Nodes
* Identify all diagram elements (boxes, shapes, entities) and node colors (if applicable).
* Extract texts (eg. using OCR)

2. Detect Edges (Connections)
* Identify relationships between nodes (arrows, lines)
* Extract:
  * Source node
  * Target node
  * Direction (if applicable)
  * Edge labels (if present)

3. Capture Styling (Optional but valuable)
* Node color / fill
* Border styles
* Edge styles (dashed, solid, etc.)

4. Preserve Text Context
* Text inside nodes
* Text between nodes (e.g. relationship labels)

### Expected Behavior

When a document contains a diagram:
* The system outputs a DOT representation alongside existing extraction results

The output should be:
* Deterministic
* Clean (no duplicates)
* Referencing consistent node IDs

### Use Cases
* Org chart extraction
* System architecture diagrams
* Process flows
* Dependency graphs
* Legal/entity relationship mapping

### Why This Matters

This unlocks:
* Graph-based processing (Neo4j, NetworkX, etc.)
* Visualization pipelines (Graphviz, D3, Mermaid conversion)
* Better explainability (traceable relationships vs flat text)
* Stronger downstream automation (agents operating on graphs)

### Example diagram output

```
digraph G {
  rankdir=TB;

  node1 [label="Node A" shape=box fillcolor="#a6cee3"];
  node2 [label="Node B" shape=diamond fillcolor="#b2df8a"];
  node3 [label="Node C" shape=box fillcolor="#fb9a99"];
[...]

PR #1410, opened 2026-08-08
feat(diagram): recover Graphviz DOT from vector SVG diagrams

MannXo:
## Related

Part of #579. Stacked on #1409, without which `main` does not compile. The first
commit here is that fix; it disappears from this diff once #1409 merges.

## Description

Recovers a diagram's nodes and edges from vector SVG and emits them as Graphviz
DOT. Deterministic, no model, no new dependency.

Against `test_documents/xml/org_chart.svg`, which the repo already ships:

```
digraph "Organization Chart" {
  n0 [label="Jane Smith\nChief Executive Officer" shape=box style=filled fillcolor="#2c3e50" color="#1a252f"];
  n1 [label="Bob Chen\nChief Technology Officer" shape=box style=filled fillcolor="#2980b9" color="#1f6da0"];
  n2 [label="Maria Garcia\nChief Financial Officer" shape=box style=filled fillcolor="#27ae60" color="#1e8449"];
  n3 [label="Alex Johnson\nChief Operating Officer" shape=box style=filled fillcolor="#8e44ad" color="#6c3483"];
  n4 [label="Engineering" shape=box style=filled fillcolor="#3498db" color="#2980b9"];
  n5 [label="Product" shape=box style=filled fillcolor="#3498db" color="#2980b9"];
  n6 [label="Finance" shape=box style=filled fillcolor="#2ecc71" color="#27ae60"];
  n7 [label="Sales" shape=box style=filled fillcolor="#9b59b6" color="#8e44ad"];
  n8 [label="Operations" shape=box style=filled fillcolor="#9b59b6" color="#8e44ad"];

  n0 -> n1 [color="#7f8c8d"];
  n0 -> n2 [color="#7f8c8d"];
  n0 -> n3 [color="#7f8c8d"];
}
```

Every node, label, fill, stroke and edge is what the source draws. `flowchart.svg`
comes out the same way with 4 nodes and 3 edges. `data_dashboard.svg` (a bar
chart) and `simple_svg.svg` produce nothing, which is the intended answer.

### How it works

`usvg` already resolves `use`, styles, units and the transform chain, so shapes
arrive in canvas coordinates. A closed path is a node, an open stroke is a
connector, and a connector's endpoints name the shapes they land on.

Closedness is the discriminator rather than fill. SVG's initial `fill` is black,
so `usvg` gives a bare `<line>` a fill exactly as it gives a `<rect>` one, and
classifying on fill turns every connector into a node.

Node shape is measured, not assumed, since by the time `usvg` is done there is no
`<rect>` or `<circle>` left to read. A path filling its bounding box is a box,
π/4 of it an ellipse, half of it with four corners a diamond.

Text needs a second pass. `usvg` is built here without its `text` feature, which
wants a font database, and text elements are dropped during conversion. So labels
come from a small walk over the source XML that reproduces the one thing that
walk lost, the transform chain down to each `<text>` anchor. There is a test for
a diagram inside a translated and scaled `<g>`, and another for a `viewBox` that
does not match `width`/`height`.

### Answers to my own two open questions

You said everything but hybrid extraction is in scope, so I picked defaults for
the two you did not need to answer, and both are easy to change.

**Where DOT lives.** A renderer registered as `dot`, reached through
`OutputFormat::Custom("dot")`, the same route `doctags` takes. A field on
`ExtractedDocument` would read better against "alongside existing extraction
results" in the issue, but it would leave every generated binding crate stale,
and `cargo install alef-cli` gives 0.17.34, which cannot parse this repo's
`alef.toml` (`vendor_mode = "registry"` is unsupported), so I cannot regenerate
them to check. Renderer now, result field whenever you want it, from the same
`InternalDocument.diagrams` this already populates.

**How far.** SVG only. `pdf_oxide`'s `extract_paths_in_rect` and DrawingML
`<p:cxnSp>` connectors both feed the same `assemble` step, which is why the
matching rules sit in `extraction/diagram/mod.rs` rather than in the SVG file.

### What this does not do

- **Edge direction comes from path point order**, not from arrowheads. `usvg`
  resolves `marker-end` into real geometry, so detecting an arrowhead near a
  connector endpoint is possible, but neither shipped fixture defines a marker
  that resolves and I would rather not ship a rule no test exercises.
- **Raster diagrams are untouched.** Your Stage 2 model plan stands for those.
- **A drawing with no connectors is not reported at all.** Bar charts, logos and
  illustrations all produce closed outlines, and an edgeless node list would be
  noise presented as structure.

### Changes

- `types/diagram.rs`, `InternalDocument.diagrams`. New internal types, all
  `alef(skip)`, so no binding regeneration is needed.
- `extraction/diagram/mod.rs`. Format-independent matching: reading order,
  duplicate collapse, background-panel and decoration filters, label-to-shape
  containment, connector snapping.
- `extraction/diagram/svg.rs`. The SVG geometry walk and the text pass.
- `rendering/dot.rs` plus registration as the `dot` renderer.
- `extractors/xml.rs`. Ten lines calling recovery for `image/svg+xml`.

### Testing

```
cargo test -p xberg --no-default-features --features xml,svg --lib -- diagram rendering::dot
test result: ok. 32 passed; 0 failed
```

32 new tests, including the three shipped fixtures and an end-to-end check that
`output_format="dot"` reaches the renderer through both `FromStr` and serde.

Both load-bearing decisions were mutation tested. Replacing the closedness
discriminator with a fill check fails 10 of 15 SVG tests; dropping the transform
chain from the text pass fails the translated-group test.

Full suite, before and after, on the same machine:

```
cargo test --locked -p xberg --features full --all-targets
  before (build fix only):  6423 passed, 83 failed
  after  (this branch):     6455 passed, 83 failed
```

Exactly +32 passing and the same 83 failures, whose names diff identically. They
are unfetched Git LFS fixtures in a shallow clone (`.doc`, `.docx`, `.ppt`,
`.pst`, `.msg`, `.heic`, `.jp2` binaries are still pointer files here), so none
of them are reachable from this change. Note the "before" baseline is the build
fix rather than `main`, since `main` does not compile.

Also run locally and clean:

```
cargo fmt --all -- --check
cargo clippy --locked -p xberg --features full --all-targets -- -D warnings
cargo clippy --workspace --exclude xberg-php --all-targets -- -D warnings
cargo test --locked -p xberg --features full --doc     # 118 passed, 0 failed
cargo check --locked -p xberg --no-default-features
cargo check --locked -p xberg --no-default-features --features ocr-wasm --lib
cargo check --locked -p xberg --no-default-features --features tokio-runtime,pdf,layout-detection --lib
cargo check --locked -p xberg --no-default-features --features xml --lib
cargo check --locked -p xberg --no-default-features --features svg --lib
```

The last two matter because recovery needs both features and the renderer needs
neither, so each has to build without the other.

One gate I could not run: `xberg-php` is excluded from the workspace clippy above
because `ext-php-rs` needs `php-config`, which is not installed on this machine.
Nothing in this change touches it.

## Checklist

- [x] CI passing


crates/xberg/src/extraction/diagram/svg.rs
```diff
@@ -0,0 +1,792 @@
+//! Diagram recovery from SVG.
+//!
+//! Shapes come from `usvg`, which resolves `use`, styles, units and the whole
+//! transform chain, so every outline and connector arrives in canvas
+//! coordinates already. Text does not: `usvg` is built here without its `text`
+//! feature, which needs a font database and drops text elements during
```
Goldziher: We have fontdb already in the dependency chain..

Why would it drop elements.

crates/xberg/src/rendering/doctags.rs
```diff
@@ -134,6 +134,16 @@ pub(crate) fn render_doctags(doc: &InternalDocument) -> String {
             ElementKind::FootnoteDefinition => {
                 push_element(&mut out, "footnote", loc, &normalize_inline_text(&elem.text), None);
             }
+            // DocTags has no comment tag. Rather than drop the body, key it and
+            // emit it as a footnote, which is what the Djot renderer does with
+            // the same element for the same reason (#300).
+            ElementKind::CommentDefinition => {
```
Goldziher: This is mixed scope

crates/xberg/src/types/diagram.rs
```diff
@@ -0,0 +1,101 @@
+//! Diagram graph recovered from a vector source.
+//!
+//! Vector diagrams (SVG today, vector PDF and DrawingML later) already contain
```
Goldziher, line 3: Should be handled, but wired to the PDF pipeline

Goldziher (commented):
So methodology - 

1. Please add test documents of the appropriate types - with diversity, to the `test_documents` submodule. 

2. Tests need to be integration with real files - first .


3. We need ground truth - to measure against and calibrate against. A few strong fixtures that prove the overall correctness. 

Then - code should be modularized, and we need to close the implementation gaps.

That's the bare minimum. 

Bonus points - 

We identify if any other oss tool does the same, we add diagraming to our benchmark harness and we measure and tune our quality and perf to Maxime quality/perf (identify hotspots, memory profuling, etc.).

MannXo:
Thanks for the review. Everything you named is in now, so here is the whole set rather than a running commentary.

### Fixtures, ground truth, integration tests

Fixtures are generated by rendering known graphs through Graphviz, which gives ground truth for free: the source `.dot` **is** the correct answer, so the corpus measures recovery instead of freezing whatever the code currently returns. They are in xberg-io/test_documents#1 along with the sources, so every fixture is regenerable.

Scoring is recall and precision against those answers, keyed by node label so it does not depend on how any recogniser numbers its output. That follows `tests/pdf_table_ground_truth.rs`, which uses thresholds the same way. Current state, all seven fixtures:

```
graphviz_flow           nodes 4/4  edges 4/4   arrowheads, edge labels, three shapes, negative root translate
graphviz_states         nodes 4/4  edges 4/4   doublecircle, antiparallel edge pair
graphviz_network        nodes 5/5  edges 4/4   undirected edges, no arrowhead anywhere
graphviz_bidirectional  nodes 3/3  edges 3/3   dir=both and dir=back
nested_transforms       nodes 4/4  edges 3/3   nested transform groups, viewBox != viewport
org_chart               nodes 9/9  edges 3/3   two-line labels, isolated leaf nodes
flowchart               nodes 4/4  edges 3/3   annotations outside every shape
```

33/33 nodes, 24/24 edges. Two fixtures have deliberately empty ground truth, `data_dashboard` and `simple_svg`: a bar chart and a two-shape drawing are not diagrams, and recovering a graph from either is a false positive, so that is asserted too.

Generating real fixtures was worth it on its own. It immediately exposed four defects hand-written SVG could never reach, including one where no straight connector could ever carry a label, because a straight line flattens to exactly two points and the midpoint was being indexed rather than measured along the line.

### Modularisation

`svg.rs` was 849 lines doing two unrelated jobs. It is now `svg/mod.rs` (recovery), `svg/geometry.rs` (the usvg walk) and `svg/text.rs` (the transform chain down to each text anchor), and the point math that is not SVG-specific moved to `polyline.rs` where the PDF front end shares it.

### PDF pipeline

Wired. `extraction/diagram/pdf.rs` produces the same outline, connector and label intermediates and hands them to the same `assemble`, so no matching logic is duplicated. Gating went from `all(svg, xml)` to `any(all(svg, xml), pdf)`; each front end keeps its own requirement.

Three things differ from SVG and each is handled:

- A PDF path object is a set of subpaths under one paint operator, so one `PathContent` routinely holds a node, an arrowhead and a connector. Each subpath is measured separately; as one run of points it produces a shape nobody drew.
- Painting operators close paths implicitly, so a filled arrowhead never emits `h`. Fill therefore implies closed, the exact opposite of SVG where every path inherits a black fill and fill means nothing. Whether a path was filled is read from the absence of a stroke colour, since a path painted in the default colour reports no colour at all.
- User space is bottom-up, so coordinates are flipped against the media box on the way in.

Cost control: a page stops after the path pass unless it yields at least two outlines and one connector, so a document of prose pays for one content stream parse per page and never a second text extraction.

The test builds its diagram with `lopdf` rather than reading a fixture, so it needs nothing from the corpus and the input is legible as operators. A `dot -Tpdf` fixture would be better still, but `*.pdf` is bucket-managed in test_documents. Happy to hand over the files if you want to publish them; the existing DOT ground truth applies unchanged to a PDF render of the same graph, which is the point of keying it by label.

### One finding in pdf_oxide worth your attention

`extract_paths` in pdf_oxide 0.3.77 has arms for `Stroke`, `Fill`, `FillEvenOdd` and `CloseFillStroke`, but none for `FillStroke` (`B`), `FillStrokeEvenOdd` (`B*`) or `CloseFillStrokeEvenOdd` (`b*`). A `B`-painted path is never finalized, so its geometry leaks into the next painted path and takes that path's colour, and if nothing follows it is dropped. `PathExtractor::fill_and_stroke` already exists and is simply never called.

Graphviz paints every filled node and every arrowhead with `B`, so on `dot -Tpdf` output this shows up as filled nodes reporting the *next* shape's colour, and each file losing exactly its last arrowhead. `graphviz_network` is the control: undirected, no arrowheads anywhere, and it scores 4/4 edges. PDF recovery is currently 16/16 nodes and about 11/15 edges, and the gap is entirely this.

It is not specific to diagrams. A `re B` table cell border merges into the next path for the layout gate's rule counting and for table detection too. I have not touched it here since it is a different repo, but I can open an issue with the repro if useful.

### Still open from your list

The OSS-tool comparison, benchmark-harness integration and profiling are not started. Say the word if you want them in this PR or as a follow-up.

### Verification

`cargo clippy --locked -p xberg --features full --all-targets -- -D warnings` clean, `cargo fmt --check` clean, and `pdf`, `xml,svg`, `pdf,xml,svg` and no-default-features all compile. Clippy is run with `--exclude xberg-php` locally, since `ext-php-rs` needs `php-config` installed.

crates/xberg/tests/diagram_dot_pdf.rs
```diff
@@ -0,0 +1,142 @@
+//! End-to-end diagram recovery from vector PDF (#579).
```
Goldziher, line 1: This file is nice, but its not enough. I will find some baseline PDFs and add them to the test_documents repository, then you can test properly.

Goldziher:
Corpus is up: xberg-io/test_documents#2. Point the submodule at it and this PR's ground-truth test
scores 7 fixtures instead of skipping all of them. Below is the full inventory you asked about —
what the corpus should hold, which extractors that implies, what the benchmark harness needs, and
what documentation goes stale.

## First, two findings from wiring the corpus up

**Your ground-truth test has been measuring nothing.** It reads `diagrams/*.svg` and
`ground_truth/dot/*.dot`; neither path existed in the submodule, so every case hit
`else { continue }`, `scored` stayed 0, and both tests printed "test_documents not populated" and
passed. That is why the numbers looked stable — nothing was being compared. Fixed by #2; you need to
repoint the fixture paths at `diagrams/svg/`.

**The Graphviz fixtures stated their own answer.** `dot -Tsvg` writes one `<title>` per node holding
the node id and one per edge holding `src->dst` — the complete edge list, verbatim, in the file. And
`SVG_TEXT_ELEMENTS` in `extraction/xml.rs:40` already includes `title`, so it reaches your
extractor. A recogniser could have scored 4/4 nodes and 4/4 edges on all four fixtures without
reading a single coordinate. Every Graphviz fixture now ships twice, as emitted and with element
titles stripped.

Good news: **your recovery scores identically on both**, 4/4, 4/4 and 3/3 on the stripped variants.
The geometry inference is real, and now there is a fixture that proves it rather than a fixture that
would have hidden it.

Third, smaller: `graphviz_network`'s ground truth restated an undirected source (`graph`, `--`) as a
digraph. It is now `graph`/`--`, which makes it score **0/0 edges** — `parse_dot` only understands
`->`, so undirected edges parse as none and score a vacuous 100%. Undirected graphs are currently
unmeasurable end to end. `diagrams/manifest.json` carries `directed: false` for the scorer to key
on.

---

## 1. Corpus

`diagrams/manifest.json` is the index. Every fixture is tagged `class: "A"` or `"B"`, and the class
decides how it should be scored:

- **Class A — the file states its graph.** Connector endpoints, model ids, `source`/`target`
  attributes, or diagram source text. Recovery is lossless, so a correct implementation scores
  **exactly 1.0**. A miss is a bug, not a threshold to tune.
- **Class B — geometry only.** SVG, vector PDF. The graph is inferred; precision/recall against a
  threshold is the right instrument.

Conflating the two is how you end up tuning a threshold to paper over a parsing bug.

### What is in the corpus now

| fixture | class | exercises |
|---|---|---|
| `diagrams/svg/graphviz_flow.svg` | B | box/diamond/ellipse, arrowheads, edge labels, dashed edge, negative root translate |
| `diagrams/svg/graphviz_states.svg` | B | `doublecircle`, antiparallel edge pair |
| `diagrams/svg/graphviz_network.svg` | B | undirected edges, no arrowheads, `neato` |
| `diagrams/svg/graphviz_bidirectional.svg` | B | `dir=both`, `dir=back` |
| `diagrams/svg/*_geometry.svg` | B | the same four with the answer removed from metadata |
| `diagrams/svg/nested_transforms.svg` | B | nested transform groups, viewBox≠viewport |
| `xml/org_chart.svg`, `xml/flowchart.svg` | B | multi-line labels, isolated nodes, `marker-end` |
| `xml/data_dashboard.svg`, `xml/simple_svg.svg` | B | negatives |
| `images/5_level_paging_..._2017.svg` | B | negative — inferno flame graph, 77 nested rects, CSS + script, zero connectors. Hardest negative in the corpus and it was already shipping. |

### What is missing, and why each matters

**Producer dialects.** Everything above is Graphviz or hand-authored. Real files are not, and each
producer has an idiom that breaks recovery tuned on Graphviz: Mermaid uses `<foreignObject>` HTML
labels; draw.io routes **orthogonal elbow connectors** with waypoints and embeds its own
`mxGraphModel`; PlantUML emits a `<g>` per element with UML stereotypes; Excalidraw's hand-drawn
roughness turns one logical stroke into many wobbly subpaths; LibreOffice uses `draw:custom-shape`
enhanced geometry.

**Graph features with no coverage at all**, roughly in order of how badly they break things:

- **cluster / subgraph containers** — a rectangle that is not a node. Guaranteed false positives.
- **swimlanes and BPMN pools** — the same problem, bigger.
- **orthogonal elbow routing** — the default in every non-Graphviz tool.
- **self-loops**, and **crossing edges that must not be joined**.
- **icon/image nodes** (AWS/Azure architecture style) — a node with no outline at all.
- **edge labels on an opaque background box** — looks exactly like a node.
- **record/HTML-like labels with ports** — one outline, many logical sub-fields.
- rotated text, CJK/RTL labels, stroke-only nodes, gradient/opacity fills, dark canvas.
- arrowhead styles beyond `normal`/`none`: `vee`, `open`, `diamond`, `tee`, crow's-foot.
- a **100+ node graph**, for the perf and memory work.

**Negative controls.** Three today. The one that worries me most is missing: **a table drawn with
ruling lines**. Same signature as a diagram — closed rectangles plus straight strokes — and it
collides directly with table detection. Also absent: pie/line charts, Gantt, maps, floor plans, icon
sets, forms, and a **mixed page** (prose + table + one diagram) where recovery has to be selective.

**Document context.** #579 requires every node and edge be traceable to "page reference and bounding
box". Nothing tests that: no multi-page PDF with the diagram on page N, no document with two or more
diagrams, no diagram inside a layout-detected figure region. The manifest has `page` and `bbox`
slots waiting.

**Vector PDF.** This is the one I said I would supply. `diagram_dot_pdf.rs` builds its input with
`lopdf` in-test, so the PDF path has never seen a real file. Coming in the next drop — `.pdf` is
bucket-managed here so it needs a publish step, which is on me, not you.

### Scope

Everything recoverable **programmatically, without a model**. Raster diagram images and scanned
diagrams are out — that is the VLM/hybrid path and it is not this PR's problem.

---

## 2. Extractors

The registry is `core/mime.rs:174-681` (`FORMATS`), registered at `extractors/mod.rs:372-473`.

### Already reached, geometry thrown away

These are the cheap wins. In each case the parse tree already walks the right element and discards
the attributes that carry the graph:

| hook | today |
|---|---|
| `extraction/pptx/parser.rs:99` | matches `"sp" \| "cxnSp"` — connectors **are visited**, only text kept. The endpoint binding is never read. |
| `extraction/pptx/parser.rs:276-290`, `:344-372` | SmartArt `dgm:relIds` resolved, but `parse_diagram_text` collects only `a:t`. `dgm:ptLst`/`dgm:cxnLst` — literally a node list and an edge list — are discarded. |
| `extractors/odp.rs:227-228` | `draw:connector` matched **by name**; `draw:start-shape`/`draw:end-shape` never read. ODF binds connectors to shapes explicitly, so this is exact recovery for ~50 lines. |
| `extraction/docx/drawing.rs:481-518` | VML `v:textbox` text only. `v:line`, `o:connectortype`, `wpc:` canvas not handled. |
| `extraction/excel.rs` | `xl/drawings/*.xml` is **never opened**. |
| `extractors/rtf/parser.rs:1098-1107` | `\shp`/`\shpinst` skipped as ignorable destinations; only `\shptxt` rescued. |
| `extraction/xml.rs:40,111,131` | SVG geometry **deliberately discarded** — in `svg_mode` attributes are skipped and only text elements emitted. This is what your PR works around. |

Effort→payoff order: **PPTX `cxnSp` + `dgm:cxnLst`** first (highest fidelity, tree already walked),
then **ODP/ODT `draw:connector`**, then a **structural SVG mode** (Mermaid and Graphviz SVG carry
`class="node"`/`class="edgePath"` and are trivially machine-readable), then DOCX `wpc:`, then RTF
`\sp`. iWork IWA and legacy `.doc`/`.ppt` Escher are expensive and can wait.

### No extractor at all

`.vsdx .vsd .drawio .excalidraw .graphml .gml .gexf .emf .wmf .eps .ps` — no MIME entry, no
registration. Each is a small, well-specified parser, and each is Class A. What happens today:

- **`.vsdx`** — it is a ZIP, `detect_office_format_from_zip` (`core/mime.rs:1006-1050`) has no Visio
  marker, so the **archive extractor walks it** and `visio/pages/page1.xml` surfaces as archive
  text. Accidental, but a real foothold. Visio's `<Connect FromSheet= ToSheet=>` is the purest graph
  of any format we handle.
- **`.drawio`/`.graphml`/`.gexf`** — sniffed to `application/xml` (`core/mime.rs:969-971`), and the
  generic XML extractor **keeps attributes** (`extraction/xml.rs:232-254`), so
  `<edge source="a" target="b"/>` survives as the literal text `edge (source: a, target: b)`. The
  graph is right there, one parser away.
- **`.excalidraw`** — sniffed to `application/json`, flattened by `StructuredExtractor`.

### Fenced diagram source is already free

`push_code(text, Some(lang), …)` (`types/internal_builder.rs:210-219`) preserves the language tag
through Markdown, MDX, Djot, AsciiDoc, RST, Org, Jupyter markdown cells, DocBook and HTML, and it
round-trips in the Markdown, Djot and JSON renderers — `rendering/json.rs:286-293` gives
`JsonNode::Code { text, language }`. A ` ```mermaid ` block needs **zero parsing work**; it needs a
consumer that reads `language ∈ {mermaid, dot, graphviz, plantuml, d2}` and hands the literal on.

Two caveats: `rendering/plain.rs:83-89` drops the fence and the tag, and `extractors/asciidoc.rs:538`
only records a language for `[source,x]`, not a bare `[mermaid]` block macro.

### Two loose ends

**`.dot` collides with Word.** `core/mime.rs:282` maps `extensions: &["doc", "dot"]` →
`application/msword`, so a Graphviz `.dot` is routed to the legacy CFB extractor. Ground-truth
`.dot` files are never extracted so the corpus is safe, but a `.dot` *input* fixture needs
magic-byte disambiguation first.

**`Page.has_vector_graphics`** (`types/page.rs:112-119`) is a serialized, documented API field that
is **hardcoded `false` at every construction site**. It is exactly the "this page holds vector art"
signal a diagram gate wants, already plumbed end to end, and nothing sets it. A path-count heuristic
in the PDF backend would light it up.

---

## 3. Benchmark harness

`tools/benchmark-harness/`. Concretely, to score diagrams:

**Output format.** `src/types.rs:12-40` — the harness's own `OutputFormat` is `{Markdown,
Plaintext}`; add `Dot` plus `Display`/`FromStr` arms. Then `src/adapters/xberg.rs:123-126`
(format→CLI flag), `:208-211` (format→slug), `:249`. And `src/quality.rs:105-107` needs an explicit
`Dot` branch — today `Dot` would fall through into the *markdown* structural path, which is wrong.

**Ground truth.** `GroundTruth` in `src/fixture.rs:65-92` already has vestigial `fields_json` and
`formulas_json` slots with no consumer; add `diagram_json` the same way, plus a resolver near `:292`,
`ground_truth_diagram` on `CorpusDocument` (`src/corpus.rs:13-30`) and
`require_diagram_ground_truth` on `CorpusFilter` (`:34-50`, matched at `:81-103`).

**Scoring.** Put it in a **new `src/diagram_quality.rs`**, attached from `quality.rs` the way
`structural_sidecar` is at `quality.rs:58-59`. Not folded into `StructuralSidecar` — a DOT graph has
no reading order and no headings, so it would poison every `present_*` gate. Node F1 on labels (ids
are arbitrary, so match on label the way `diagnostic_matches` matches on `repr_text`), edge F1 on
the induced node pairing.

**Do not fold it into SF1 yet.** Adding a dimension changes the normalization denominator at
`structural_sidecar.rs:707`, which changes **every existing `min_sf1`** in `guardrails.json` — 200+
contracts to regenerate. Report-only first, the way D6 was introduced. Your own CHANGELOG states the
policy: *"Add score contracts to `guardrails.json` only after fresh fixed outputs have been
measured."* If you do fold it later, `dimensions()` and `dimensions_pr()`
(`structural_sidecar.rs:226`, `:240`) have the arity **in the signature**, and the weights to
rebalance are the six constants at `:48-56`.

**Cohort.** Add a `BenchmarkGroup` at `src/groups.rs:57`. It will not work as-is:
`resolve_group_docs:181-209` reroots into `fixtures_dir.join("pdf")` and hardcodes
`file_types: Some(vec!["pdf"])`, so SVG fixtures are invisible and you get a hard `Error::Config`.
Add a `file_types` field to `BenchmarkGroup` (`:9-23`) and thread it through; the test at `:311-320`
asserts today's PDF-only behaviour and must be updated, while `:329-353` freezes `smoke`=10,
`promotion`=26, `holdout`=1 and must stay green.

Fixtures already exist and are half-ready: `fixtures/svg_flowchart.json` and
`fixtures/svg_org_chart.json` point at the two `xml/` diagrams. They sit at the fixtures **root**, so
`resolve_group_docs` cannot see them today.

**Frame it as intra-xberg, not competitive.** Only 2 of 7 competitor adapters declare `svg` at all
(`src/adapters/external.rs:53`, `:66-71`), and **none emits a graph format** — every wrapper script
produces text or markdown, and `unstructured` is pinned plaintext-only (`:169`). A cross-framework
diagram ranking would be vacuous. This belongs in the `compare`/`pipeline-benchmark` path where
`--group` lives, not the `run`/`consolidate` CI matrix. I would skip the `benchmarks.yaml` axis
entirely for v1.

---

## 4. Documentation

**`README.md` is generated** — it is rendered by `alef` from `templates/readme/root.md`
(`alef.toml:3866-3881`). Edit the template, not the output. Same for every
`docs-site/src/content/docs/reference/*.md`, which carry an `alef:hash` header.

Hand-edit:

1. `docs-site/src/content/docs/guides/output-formats.mdx:107-116` — the canonical `OutputFormat`
   table; `:31` — the prose list.
2. `docs-site/src/content/docs/reference/formats.mdx:53` — the SVG row currently reads "Treated as
   XML document", which this PR makes untrue. Also `:166-179`, the Content Formats table (already
   stale — missing `json` and `doctags`).
3. `docs-site/src/content/docs/concepts/plugin-system.md:217-224` — "four built-in renderers" (already
   wrong; `doctags` is a fifth, `dot` a sixth).
4. `docs-site/src/content/docs/features.mdx:189` — "five output formats" (already wrong).
5. `docs-site/src/content/docs/cli/usage.mdx:463`.
6. **New** `guides/diagram-output.mdx`, modelled on `guides/doctags-output.mdx`, plus a sidebar entry
   in `astro.config.mjs` after line 126 — without it the page is unreachable.
7. `templates/readme/root.md:11,39,430` and `templates/readme/partials/features.md.jinja:102` — the
   "6 output formats" counts. That partial fans out to 11 package READMEs, so one template edit
   fixes all of them.
8. `CHANGELOG.md:10` — hand-written Keep a Changelog; add an `### Added` paragraph under
   `## [Unreleased]`.
9. `plugin/skills/picking-a-format/SKILL.md:3,21`.
10. `sdks/spec/api/openapi.yaml:4825` — the `output_format` description enumerates values in prose.

Then `task alef:generate` and `task plugin:generate`.

**CI will catch you** on `task alef:verify` (`ci-lint.yaml:131-156`) if Rust doc comments change
without regenerating, and on `task docs:snippets:all` (`:217-231`) — `snippets:gaps` fails if you add
a `dot` snippet for one language and not all 15.

### The feature is currently unreachable from two entry points

- `crates/xberg-cli/src/main.rs:673-698` — `ContentOutputFormatArg` is a `clap::ValueEnum`, so
  `--content-format dot` is **rejected at argument parsing**. The harness shells the CLI, so this
  blocks benchmarking too.
- `crates/xberg/src/api/handlers.rs:443-456` — `parse_output_format` is a hard whitelist of
  `plain|markdown|djot|html`. The REST multipart path cannot reach `dot` (and cannot reach
  `json`/`structured`/`doctags` either — pre-existing, worth a separate issue).

### On `Custom("dot")` vs a real variant

`DocTags` is the precedent: it is **both** an `OutputFormat::DocTags` variant and a registry renderer
named `"doctags"` reachable via `Custom("doctags")`. Going `Custom`-only means nothing appears in the
generated reference docs at all, because `Custom(String)` carries no enumeration. Every binding
already models `Custom(String)`, so no regeneration is *required* — but nothing self-documents
either. I would follow the DocTags shape.

---

## 5. Where the remaining fixtures come from

Sourced and licence-verified against each upstream's actual LICENSE file, not inferred from the
tool's licence. Only permissive classes are usable here — MIT / Apache-2.0 / BSD / CC0 / CC-BY /
CDLA-Permissive / MPL-2.0 / US-PD. ShareAlike, NonCommercial and GPL content are out.

**`bpmn-miwg/bpmn-miwg-test-suite` (CC-BY-3.0) is the find of the audit.** The same 12 canonical
process graphs, rendered by ~40 different tools into 840 `.bpmn` + 27 PDF + 20 SVG + 13 VSD +
3 VSDX + 990 PNG. So for one graph you get the ground truth *and* a vector PDF *and* a Visio
`<Connect>` edge list *and* an SVG — the same answer in five containers, which is exactly what a
container-agnostic recogniser needs to be tested against. Verified:
`itp-commerce.../B.1.0.vsdx` has 58 `<Connect>`; `Reference/B.1.0.pdf` page 1 has 353 vector
curves, 22 rects, 16 lines and live text.

| need | source | licence |
|---|---|---|
| PPTX bound connectors | `LibreOffice/core` `oox/qa/unit/data/{connectorConnection,elbowConnectors,curvedConnectors}.pptx`, `tdf165180_standardConnectors{ECMA,ISO29500}.potx` | MPL-2.0 |
| PPTX SmartArt | `LibreOffice/core` `sd/qa/unit/data/pptx/smartart-*.pptx` — 40 files: org-chart, cycle, chevron, pyramid, matrix, recursion, maxdepth | MPL-2.0 |
| SmartArt in 3 containers | `apache/tika` `test{PPT,WORD,EXCEL}_diagramData.{pptx,docx,xlsx}` — one data model, three wrappers (54 / relIds / 12 `dgm:cxn`) | Apache-2.0 |
| DOCX drawing canvas | `LibreOffice/core` `oox/qa/unit/data/WPC_{Bent,Curved}Connector*.docx` — `wpc:wpc` + `wps:cNvCnPr` + bound endpoints | MPL-2.0 |
| ODF connectors | `LibreOffice/core` `sd/qa/unit/data/odg/rhbz1870501.odg` (22 connectors), `odp/tdf147919.odp` (16) — `draw:start-shape`/`draw:end-shape` + glue points | MPL-2.0 |
| Visio | `apache/poi` `test-data/diagram/60489.vsdx` — 30 `<Connect FromSheet= ToSheet=>` | Apache-2.0 |
| draw.io | `jgraph/drawio-diagrams` — 615 files across 20 domains, plus `diagrams/flowchart.drawio.svg` which carries the whole `mxGraphModel` in the SVG `content=` attribute | Apache-2.0 |
| Mermaid | `mermaid-js/mermaid` — 159 `.mmd` + `demos/*.html` | MIT |
| PlantUML source | `plantuml-stdlib/C4-PlantUML` `samples/*.puml` — 17 real C4 diagrams as `Rel()` edges | MIT |
| GraphML / GEXF / D2 | `apache/tinkerpop` `data/*.xml`; `graphology` `src/gexf/test/resources/*.gexf`; `terrastruct/d2` (64 `.d2`) | Apache-2.0 / MIT / MPL-2.0 |
| C4 workspaces | `structurizr/java` `structurizr-*-workspace.json` — explicit `relationships[]` plus per-view x/y | Apache-2.0 |
| vector PDF at scale | arXiv CC-BY-4.0 via OAI-PMH — the `<license>` field is machine-readable, and ~45% of a day's `cs` submissions are CC-BY-4.0 | CC-BY-4.0 |

Rejected, for the audit trail: **Graphviz's own `.gv` gallery is EPL-1.0**, so we generate our own
rather than vendoring theirs. Gephi is GPL-3.0 and its sample datasets carry no licence statement at
all. Network Repository is CC-BY-SA. `bpmn-io/bpmn-js` has a watermark clause that makes it not
clean MIT (use `bpmn-moddle`, which is). arXiv's *default* licence is a non-exclusive grant to arXiv
only — not redistributable — so the CC-BY filter is mandatory, not optional.

Two corrections worth having:

- **The OOXML attribute is not `stCxnId`/`endCxnId`.** It is `<a:stCxn id="…" idx="…"/>` and
  `<a:endCxn …/>` inside `p:cxnSp/p:nvCxnSpPr/p:cNvCxnSpPr`. Grepping for `stCxnId` returns zero
  hits everywhere and will make you wrongly conclude a good file has no bound connectors. (Our
  `pptx/pitch_deck_presentation.pptx` genuinely has none under either spelling — its connectors are
  unglued, which makes it a good geometry-inference fixture rather than a Class A one.)
- **US-government PDFs are the wrong bet for vector diagrams.** NIST SP 800-207 was checked page by
  page: every one of Figures 1-12 is a raster paste-in from Word, produced by Acrobat PDFMaker.
  Licensing is perfect, extraction value is nil. CC-BY arXiv (LaTeX/TikZ) and the MIWG exports are
  where real vector geometry lives.

MannXo:
Picked up the extended corpus. It found six real bugs, all now fixed, and the results are on the whole 26-fixture set.

| fixture | nodes | edges |
|---|---|---|
| graphviz_flow / states / network / bidirectional | 4/4, 4/4, 5/5, 3/3 | 4/4, 4/4, 4/4, 3/3 |
| graphviz_clusters | 5/5 | 4/4 |
| graphviz_selfloop | 4/4 | 5/5 |
| graphviz_ortho / record / cjk | 5/5, 3/3, 5/5 | 5/5, 2/2, 4/4 |
| graphviz_large | 128/128 | 137/141 |
| mermaid_flow | 6/6 | 6/6 |
| plantuml_activity / swimlane | 5/5, 5/5 | 4/4, 4/4 |
| libreoffice_connectors | 4/4 | 3/3 |
| nested_transforms / mixed_page / org_chart / flowchart | 4/4, 3/3, 9/9, 4/4 | 3/3, 2/2, 3/3, 3/3 |
| icon_nodes | 0/4 | 0/3 |
| negative_ruled_table / negative_pie_chart / negative_form / data_dashboard / simple_svg / 5_level_paging | no graph | no graph |

24 of 26 at 1.0 on both, and **every `_geometry` variant scores exactly what its original scores**. That was the question the split was built to ask, and the answer is that recovery is reading coordinates, not `<title>`.

Scoring is now driven by `diagrams/manifest.json` instead of a list in the test, so a fixture you add upstream gets measured without a change here. Class A fixtures are held to exactly 1.0.

### What the new fixtures caught

**Containers.** Cluster and swimlane boxes were being reported as nodes, and they also stole the edges, since a connector between two boxes inside a cluster reaches the cluster border before its target. A shape enclosing two or more others is a container; one enclosed shape is a double border, which was already handled, so the threshold is two.

**Self-loops** were dropped outright. A connector returning to the shape it left is a real edge when it arcs clear of the node, while one lying entirely inside a shape is interior detail and joins nothing.

**Mermaid** recovered nothing at all. It draws no `<text>`: labels are XHTML inside `<foreignObject>`, which carries a box rather than an anchor, so the anchor is its centre.

**The ruled table** in `mixed_page` was a node. It is one closed rectangle full of text, which is a node in every respect except that its text is laid out in rows *and* columns. Each alone is something a real node does, a wrapped caption being many rows in one column and a record being one row in many columns, so only both together is a table.

**The pie chart** was recovering as a graph, legend boxes wired to wedges by leader lines. The wedges all meet at the centre, so their boxes overlap, and a layout engine keeps real nodes apart.

**PlantUML terminal dots** were reported as unlabelled nodes. An anonymous shape much smaller than the labelled ones is decoration.

Both geometric tests only ever condemn a shape nobody labelled, so a named node stays a node whatever its neighbours do. `graphviz_selfloop` proved that necessary: its fixed-position nodes overlap, and an overlap rule that ignored labels deleted half the graph.

### The two that are not there yet

Pinned at what they score today, so the gap is visible and a regression still fails the test.

`graphviz_large` loses 4 of 141 edges, all long-range shortcuts passing close to intervening nodes.

`icon_nodes` needs two things: merging the several subpaths of one glyph into a single node, and reading a caption that sits below a shape rather than inside it. The second is the risky half, because a caption below a shape and an edge label above the next one are the same geometry, and getting it wrong costs edge labels everywhere. I would rather do it deliberately than bolt it on. Say if you want it in this PR.

### Unrelated, while I was here

Fetching the corpus with `scripts/fetch_corpus.py` took the `xberg` lib suite from 18 failures to 1, and that last one is `extractors::tests::test_register_default_extractors` asserting 12 extractors where 13 are registered under `pdf,xml,svg`. It fails the same way with my branch stashed, so it is not from this work.

The pdf_oxide `B` operator gap from my previous comment still stands and still costs the PDF path its arrowheads.

Goldziher:
Can you open an issue in pdf_oxide, and link it to this PR + add a tracking comment on the GitHub issue?

Great work @MannXo   

If you join our discord server we can discuss org membership (if you intend to keep contributing regularly).

MannXo:
Filed as yfedoseev/pdf_oxide#1028 (issues are disabled on your fork, and that is the repository `Cargo.toml` points at).

Narrowing it down for the issue turned up a bit more than I had in the comment above. It is not only `B`: `B*` and `b*` are affected too, and `b` escapes only because `close_fill_and_stroke` happens to be wired up. Painting one rectangle and counting what comes back:

| operator | paths returned |
|---|---|
| `B` | **0** |
| `B*` | **0** |
| `b` | 1 |
| `b*` | **0** |
| `f` | 1 |
| `S` | 1 |

Two distinct symptoms, depending on what follows. With another painted path after it, the unfinalized geometry is emitted as part of that path and takes its colours, which is where `graphviz_flow`'s Start box picks up the diamond's green. With nothing after it, the shape is dropped outright, which is why each Graphviz PDF loses precisely its last arrowhead.

The issue has a self-contained repro (a PDF built with lopdf, no fixture needed), the operator table, and the three missing match arms. I offered to send the PR there.

Worth flagging that this reaches beyond diagrams: `re B` is the ordinary way to paint a table cell border, so `pdf/layout_gate.rs` ruled-line counting and `pdf/oxide/table.rs` detection both see those borders merged into whatever is painted next.

Tracking comment added on #579.

Thanks, and yes to Discord.

PR #1413, opened 2026-08-11
fix(pptx): preserve slide page numbers for untitled slides

### LINKED ISSUE #1412: bug: PPTX: the slide counter only counts *titled* slides, so tables and list items get the wrong page
## Summary

`PptxExtractor::build_internal_document` (`src/extractors/pptx.rs`, xberg 1.0.14)
rebuilds the document by splitting the already-rendered markdown on `"\n\n"` and
walking the blocks. Its slide counter advances **only** on a block that starts
with `"# "`:

```rust
let mut slide_num: u32 = 0;
for block in &blocks {
    if let Some(title_text) = trimmed.strip_prefix("# ") {
        slide_num += 1;                                          // titles only
        ...
    }
    if trimmed.starts_with('|') {
        builder.push_table_from_cells(&cells, Some(slide_num), None);   // ~line 101
    }
    ...
    builder.push_list_item(item_text, ordered, vec![], Some(slide_num), None);  // ~line 138
}
```

So `slide_num` is *the number of titled slides seen so far*, not the slide
number. A slide that emits no title never advances it, and every table and list
item after that point is stamped with a page number short by the number of
untitled slides that preceded it.

The error compounds down the deck and is unbounded: it grows with every
untitled slide. It is also silent — the page number is plausible and in range.

## Proposed fix

The slide number is already known upstream of this function — `SlideIterator`
assigns `slide.slide_number` from the archive order and `ContentBuilder::end_slide`
stores it on each `PageContent`. `build_internal_document` should not re-derive it
from rendered markdown at all; it should carry the per-slide structure through, or
at minimum advance its counter on the page markers `ContentBuilder::start_slide`
already knows how to insert (`cfg.marker_format`, `{page_num}`) rather than on
the presence of a title.

Deriving document structure by re-parsing the extractor's own markdown output is
the root of it: the markdown is lossy about slide boundaries, because a slide
without a title leaves no boundary marker in it.

## Reproduction

Both decks are 8 slides with one 3×3 table on slide 8. The only difference is
how many slides carry a title placeholder.

| file | titled slides | table really on | `tables[0].page_number` | `pages[]` carrying the table |
|---|---|---|---|---|
| `fail.pptx` (12 KB) | 3 of 8 | slide 8 | **3** | **3 and 8** |
| `pass.pptx` (12 KB) | 8 of 8 | slide 8 | 8 | 8 |

Note for anyone building their own reproducer: a slide with *no* title
placeholder can still emit a `# ` block, because `Slide::to_markdown`
(`src/extraction/pptx/mod.rs`) falls back to "the first text element under 100
characters is the title". The untitled slides in the generator therefore carry a
single 100+ character paragraph.

[xberg-upstream-pptx-table-page-fail.pptx](https://github.com/user-attachments/files/30905062/xberg-upstream-pptx-table-page-fail.pptx)
[xberg-upstream-pptx-table-page-pass.pptx](https://github.com/user-attachments/files/30905061/xberg-upstream-pptx-table-page-pass.pptx)

```text

```

FU-max-boop:
## Related

Fixes #1412.

## Description

- Carry each archive-derived slide number alongside its rendered content in a crate-internal `PptxInternalExtraction` value instead of embedding a forgeable marker in user-controlled Markdown.
- Parse each slide independently when rebuilding the internal document, so tables, lists, headings, and paragraphs receive the archive slide number.
- Preserve the legacy all-untitled `ElementKind::Slide` sentinel while avoiding new slide nodes for titled decks.
- Keep combined content and page-boundary offsets free of synthetic slide comments.

Tables and list items after untitled slides now retain their actual slide number. Marker-like text inside a deck remains ordinary content and cannot restamp later elements.

## Validation

- `cargo test -p xberg --features office --lib 'pptx::tests::'` — 40 passed
- `XBERG_SKIP_TEST_DOCUMENTS=1 cargo test -p xberg --features office --lib` — 2,317 passed
- `cargo clippy -p xberg --features office --lib -- -D warnings`
- `cargo fmt --all -- --check`
- `git diff --check`
- Real-PPTX regression fixture: an untitled slide with a long paragraph and table receives page 1 for both elements and retains the legacy slide sentinel

The corpus skip is needed only because this shallow checkout does not initialize the `test_documents` submodule.

## Checklist

- [ ] CI passing

### DIFF
--- a/crates/xberg/src/extraction/pptx/mod.rs
+++ b/crates/xberg/src/extraction/pptx/mod.rs
@@ -78,6 +78,17 @@ pub struct PptxExtractionOptions {
     pub inject_placeholders: bool,
 }
 
+/// Crate-internal PPTX extraction output.
+///
+/// `slide_contents` keeps the archive-derived slide number alongside each
+/// rendered slide. The public result remains unchanged, while the internal
+/// extractor can rebuild page-aware elements without embedding forgeable
+/// boundary markers in user-controlled text.
+pub(crate) struct PptxInternalExtraction {
+    pub(crate) result: PptxExtractionResult,
+    pub(crate) slide_contents: Vec<(u32, String)>,
+}
+
 impl Default for PptxExtractionOptions {
     fn default() -> Self {
         Self {
@@ -118,12 +129,13 @@ fn join_runs_with_spacing(runs: &[Run], extract: impl Fn(&Run) -> String) -> Str
 ///
 /// # Returns
 ///
-/// A `PptxExtractionResult` containing extracted content, metadata, and images.
-pub(crate) fn extract_pptx_from_path(
+/// The public extraction result plus archive-derived per-slide content used
+/// internally to assign trustworthy page metadata.
+pub(crate) fn extract_pptx_from_path_with_slide_contents(
     path: &str,
     options: &PptxExtractionOptions,
     warnings: &mut Vec<ProcessingWarning>,
-) -> Result<PptxExtractionResult> {
+) -> Result<PptxInternalExtraction> {
     let container = PptxContainer::open(path)?;
     extract_pptx_from_container(container, options, warnings)
 }
@@ -138,11 +150,20 @@ pub(crate) fn extract_pptx_from_path(
 /// # Returns
 ///
 /// A `PptxExtractionResult` containing extracted content, metadata, and images.
+#[cfg(test)]
 pub(crate) fn extract_pptx_from_bytes(
     data: &[u8],
     options: &PptxExtractionOptions,
     warnings: &mut Vec<ProcessingWarning>,
 ) -> Result<PptxExtractionResult> {
+    Ok(extract_pptx_from_bytes_with_slide_contents(data, options, warnings)?.result)
+}
+
+pub(crate) fn extract_pptx_from_bytes_with_slide_contents(
+    data: &[u8],
+    options: &PptxExtractionOptions,
+    warnings: &mut Vec<ProcessingWarning>,
+) -> Result<PptxInternalExtraction> {
     let container = PptxContainer::from_bytes(data)?;
     extract_pptx_from_container(container, options, warnings)
 }
@@ -151,7 +172,7 @@ fn extract_pptx_from_container<R: std::io::Read + std::io::Seek>(
     mut container: PptxContainer<R>,
     options: &PptxExtractionOptions,
     warnings: &mut Vec<ProcessingWarning>,
-) -> Result<PptxExtractionResult> {
+) -> Result<PptxInternalExtraction> {
     let config = ParserConfig {
         extract_images: options.extract_images,
         plain: options.plain,
@@ -177,6 +198,7 @@ fn extract_pptx_from_container<R: std::io::Read + std::io::Seek>(
 
     let mut total_image_count = 0;
     let mut total_table_count = 0;
+    let mut slide_contents = Vec::with_capacity(slide_count);
     let mut extracted_images = Vec::new();
     let mut collected_hyperlinks: Vec<(String, Option<String>)> = Vec::new();
     let mut doc_builder = if include_structure {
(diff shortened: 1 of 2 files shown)


Goldziher (changes requested):
Thanks — this is a real fix for a real defect, and `test_internal_slide_markers_set_table_and_list_page_numbers` genuinely fails without it and asserts exact values (`page_number == 3`, `page == Some(3)`) rather than truthiness. I want it in. Four things first.

## First, a heads-up that isn't your fault

**No CI has run on this PR at all.** Not "green" — zero check runs on `1d51f0b0`; `gh pr checks` reports "no checks reported". It's a cross-repository PR and our workflows never fired. So the `38 passed / 2,315 passed / clippy clean` in your description is your local run, which I believe, but nothing has independently compiled this. I'll get CI running before merge; flagging it so neither of us mistakes silence for success.

## 1. The `"\n\n"` split is still the structural source of truth

`extractors/pptx.rs:69` — `content.split("\n\n")` — is unchanged. What changed is which token advances the counter: `# ` headings before, `<!-- Slide number: N -->` now. That is a genuine improvement, and I'm not asking you to rewrite the extractor. But it keeps us deriving document structure by re-parsing our own rendered markdown, which is what #1412 was really about, so:

**A deck can spoof the marker.** The only guard is `(1..=slide_count).contains(&number)`. A text box containing the literal string `<!-- Slide number: 5 -->` — or a deck built from our own extracted output — is indistinguishable from a real boundary and silently re-stamps every table and list after it. Please make the marker unforgeable rather than merely well-behaved: emit it out-of-band, or carry the slide index alongside the string instead of inside it. `DocumentStructureBuilder` already exists on this path (`extraction/pptx/mod.rs:182-186`) but every call site hardcodes `include_structure: false` — that's the seam if you want it.

## 2. An untested behaviour change: PPTX loses its `Slide` node

Previously, an all-untitled deck left `slide_num == 0` and fell into `push_slide(1, None, Some(1))` at `pptx.rs:154-156`. A marker now always sets `slide_num >= 1`, so that branch is dead for any real deck and PPTX output will never emit an `ElementKind::Slide` again. That's visible downstream — `rendering/comrak_bridge.rs:811` emits a `ThematicBreak` for it, `rendering/json.rs:403-412` emits a `JsonNode::Slide` — so markdown loses a leading `---` and JSON loses a node.

It's arguably an improvement (the old node claimed to be slide 1 of an 8-slide deck), but it's unmentioned and untested. Please either keep the branch reachable or delete it and add a test asserting the new output shape.

## 3. Finish the job while you're in there

`push_heading` (`pptx.rs:89`) and `push_paragraph` (`pptx.rs:145`) still pass `page: None`, even though `slide_num` is now correct and in scope right there. Paragraph elements from PPTX stay page-less for no reason. Two-line change, same defect.

## 4. Two smaller notes

- `add_slide_header` already had a call site at `extraction/pptx/mod.rs:602-603` gated on `config.include_slide_comment` — which defaults to `false` (`elements.rs:211`) and is never set anywhere. You've added a second, unconditional emitter, so the same marker now has two paths with different semantics. Please reconcile them.
- `PageBoundary` byte offsets shift. `start_slide` captures `byte_start` before the new marker, so each boundary range now includes its own comment, and content always begins with `"\n\n"` where it previously began with slide text. The chunker recomputes from `pages` so this is mostly harmless, but `metadata.pages.boundaries` is API-visible and its values change. Worth a line in the PR description at minimum.

## Testing

The regression test drives `build_internal_document` with a hand-written string, so the issue's own reproduction — a real PPTX with an untitled slide plus a table — isn't covered end to end. `create_test_pptx_bytes` only emits short text slides, which all become titles via the <100-char fallback (`extraction/pptx/mod.rs:625-636`), so this needs a new fixture builder. Please add one; that's the test that would have caught #1412 in the first place.

No binding impact — nothing here is public or reachable from the FFI surface, and `PptxExtractor` carries `alef(skip)`. Nothing needs regenerating.

Separately: #1418 is the same defect in the legacy `.ppt` extractor and is being fixed structurally, not with markers. Worth keeping the two approaches from diverging.

Goldziher (approved):
Approving. All four items from the previous review are addressed, and the approach changed rather than the symptom: the archive slide number now travels out-of-band in `PptxInternalExtraction::slide_contents` instead of being re-parsed out of rendered markdown, so item 1 is fixed structurally rather than guarded.

Since **no CI has ever run on this PR** (0 check runs, cross-repository), I verified locally on `f1483b9` rather than taking the description on trust:

- `cargo test -p xberg --features office --lib pptx` — **52 passed, 0 failed**, including all three new tests.
- `cargo check -p xberg --features office` (non-test) — **exit 0**. This one mattered: the PR puts `extract_pptx_from_bytes` behind `#[cfg(test)]`, so a `--lib` test run cannot prove the non-test build still links. It does.
- `cargo clippy -p xberg --features full --all-targets -- -D warnings` — **exit 0**.

An `office`-only `--all-targets` clippy leg is red on `make_ole_package`/`test_documents_dir` dead code, but that is pre-existing from `94a2369a4c` gating those tests on `excel` while leaving the helpers, in a file this PR does not touch. Not yours.

Two notes for the record, neither blocking:

- The notes-appending block in `extraction/pptx/mod.rs:225-233` duplicates `content_builder.rs:204-214` (minus the trailing newline, which is harmless since blocks are trimmed). Two copies of the `### Notes:` / `Notes:` convention can now drift independently.
- `!saw_title` correctly preserves the legacy sentinel, and the e2e test pins it. Worth knowing it is now keyed on any titled slide anywhere in the deck rather than on the counter, which is the same shape but not the same predicate.

Merging.

PR #1469, opened 2026-08-23
feat(core): expose cooperative cancellation to Rust callers

jferments:
## Related

Related to #1313, which uses Xberg's existing `CancellationToken` for REST async-job cancellation.

## Description

Expose Xberg's existing cooperative cancellation controls to downstream Rust applications.

`ExtractionConfig` already accepts a `CancellationToken`, but downstream Rust callers cannot currently signal or inspect it because `cancel()` and `is_cancelled()` are crate-private.

### Why this is useful for Rust embedders

This is useful beyond any one application because extraction is often embedded inside a longer-lived process that owns the lifecycle of the work. In those environments, simply dropping a future or no longer awaiting a result is not necessarily equivalent to stopping synchronous parser work, native calls, OCR/model work, or other extraction already in progress. The application needs a way to propagate its own cancellation decision into Xberg's existing cooperative cancellation path.

Examples include:

- desktop and mobile applications exposing a user-visible **Stop/Cancel** action;
- HTTP/RPC handlers propagating client disconnects or request cancellation into extraction;
- job queues withdrawing work that has been cancelled, superseded, or deprioritized;
- interactive applications replacing an in-progress extraction when the user chooses a newer document or configuration;
- services performing graceful shutdown without intentionally starting a second cancellation mechanism around Xberg;
- resource-constrained applications that need obsolete work to release worker capacity as soon as Xberg reaches an existing cancellation checkpoint.

Xberg already has the difficult part of this mechanism: a cloneable token is threaded through `ExtractionConfig`, extraction code polls it at established checkpoints, and timeouts / REST async jobs already use it. Exposing the two existing controls to Rust callers lets embedders participate in that same mechanism instead of requiring each application to invent a parallel cancellation abstraction or treat abandonment of the outer future as cancellation.

This keeps the proposal deliberately small: it makes an existing internal capability usable by Rust callers without adding new checkpoints, changing cancellation behavior, or creating a new controller/handle type.

### Changes

This PR:

- makes `CancellationToken::cancel()` public;
- makes `CancellationToken::is_cancelled()` public;
- makes `cancel()` available independently of `tokio-runtime` and native-target gating, including wasm32;
- documents cooperative latency, one-shot lifecycle, and completion races;
- documents `ExtractionConfig::cancel_token` as a supported Rust embedding hook;
- updates `XbergError::Cancelled` documentation;
- adds an integration test compiled as a downstream crate, proving that a retained clone can request cancellation and that the clone stored in `ExtractionConfig` observes the same state.

Cancellation remains cooperative rather than forceful: `cancel()` requests cancellation, and active work stops when it reaches an existing cancellation checkpoint. This PR does not add checkpoints, dependencies, or a second cancellation abstraction.

The token and `ExtractionConfig::cancel_token` remain excluded from Alef generation, so this does not add cancellation to generated language bindings.

### Validation

Passed locally after rebasing onto current `main`:

- `cargo test --locked -p xberg --no-default-features --test cancellation_public_api`
- `cargo test --locked -p xberg --test cancellation_public_api`
- `cargo test --locked -p xberg --no-default-features --lib cancellation::tests` — 5/5 passed
- `cargo check --locked -p xberg --no-default-features`
- `cargo check --locked -p xberg --no-default-features --features wasm-target --target wasm32-unknown-unknown`
- Clippy with `-D warnings` for the xberg library and new integration test under both minimal and default features
- focused `rustfmt --check` across all four modified files
- `cargo doc --locked -p xberg --no-deps --no-default-features`
- `cargo doc --locked -p xberg --no-deps`

Repository-wide checks currently expose unrelated baseline/environment issues outside this diff: existing formatting drift, an unrelated DOCX Clippy doc-list lint, an unrelated OCR doctest snippet, and the full-feature local Clippy build requires Xberg CI's custom libheif >=1.21 installation (Ubuntu 24.04 ships an older version).

## Checklist

- [ ] CI passing

### DIFF
--- a/crates/xberg/src/cancellation.rs
+++ b/crates/xberg/src/cancellation.rs
@@ -2,8 +2,8 @@
 //!
 //! Provides a lightweight, FFI-friendly cancellation primitive based on
 //! `Arc<AtomicBool>`. The token can be cloned and shared across threads;
-//! any holder can cancel the operation and all other holders will observe
-//! the cancellation on their next check.
+//! any holder can request cancellation and all other holders will observe
+//! the request on their next check.
 //!
 //! # Design
 //!
@@ -15,6 +15,14 @@
 //! - The token wraps an `Arc<AtomicBool>` and can be stored in
 //!   `ExtractionConfig` without layout surprises.
 //!
+//! # Rust callers
+//!
+//! Rust callers may place one clone in `ExtractionConfig::cancel_token` and
+//! retain another clone in the application layer. Calling [`CancellationToken::cancel`]
+//! requests that extraction stop when the active pipeline reaches its next
+//! cancellation checkpoint. Cancellation is cooperative rather than immediate,
+//! so latency depends on the extractor and operation currently in progress.
+//!
 //! # FFI and bindings
 //!
 //! No binding exposes cancellation today. `crates/xberg-ffi` emits no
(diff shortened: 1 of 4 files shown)


Goldziher (approved):
The cancellation API change is focused and sound. I will follow up on main with the required CHANGELOG entry, issue linkage, constructor/docs polish, and an end-to-end cancellation regression.

Issue #1567, opened 2026-09-05
chore(lint): pay down the size and complexity baseline


Goldziher:
`poly lint` is green in this repo, but not because the code is under the size and complexity limits. **863 findings across 420 files** are listed individually in `alef.toml` under the `# --- quality-debt baseline ---` block.

### Why it is a per-file list and not a rule switched off

Listing each file keeps the metrics **active on every file that is not listed**, so new code is still checked and new violations still fail. The cost is that each existing offender has to be paid off individually — which is exactly what makes each one a self-contained task.

### What one unit of work looks like

1. Pick a path from the list below.
2. Reduce its findings — extract helpers from long functions, early-return to flatten nesting, group wide parameter lists into a struct, split an oversized file by responsibility.
3. Delete that file's entry from the baseline in `alef.toml`.
4. Run the crate's/package's tests and `poly lint` — both must be green.

### Ground rules

- **Behaviour must not change.** This is restructuring, not redesign. No arithmetic, ordering, dispatch or output changes.
- A **smaller, safe** change is worth more than a complete one. If a file cannot be reduced without a behaviour risk you cannot rule out, say so in the PR and leave it.
- Do not add new paths to the baseline.
- This repo's `poly.toml` is **generated** from `alef.toml` — edit `alef.toml` and re-run `alef scaffold`, never `poly.toml` directly.

### Rules in this baseline

| rule | files | meaning |
|---|---:|---|
| `function-too-long` | 302 | function longer than 80 lines |
| `file-too-long` | 235 | file longer than 1000 lines |
| `cyclomatic-complexity` | 131 | cyclomatic complexity over 20 |
| `nesting-too-deep` | 114 | more than 4 levels of nesting |
| `too-many-parameters` | 65 | more than 6 parameters |
| `type-too-long` | 16 | type longer than 300 lines |

### Best places to start

These carry the most findings. The ones with four or more complexity findings also have a `TODO(quality-debt)` note at the top of the file.

| file | rules |
|---|---|
| `crates/xberg-native-pdf/src/extractors/images.rs` | `cyclomatic-complexity`, `file-too-long`, `function-too-long`, `nesting-too-deep`, `too-many-parameters` |
| `crates/xberg-native-pdf/src/extractors/text/mod.rs` | `cyclomatic-complexity`, `file-too-long`, `function-too-long`, `nesting-too-deep`, `too-many-parameters` |
[...]

PR #1569, opened 2026-09-06
refactor(image): group the DPI decision behind a struct in preprocessing

MannXo:
## What

One file paid off the `#1567` baseline: `crates/xberg/src/image/preprocessing.rs`, listed for
`too-many-parameters`.

Two private helpers carried `#[allow(clippy::too_many_arguments)]` and between them took the same
five values describing one decision: the DPI the module settled on, the scale it implies, whether it
was auto-adjusted, and the original DPI it came from. Those five now travel as a `DpiPlan`.

| function | before | after |
|---|---:|---:|
| `create_skip_result` | 9 | 5 |
| `perform_resize` | 12 | 6 |

`perform_resize` also takes its two dimension pairs as tuples rather than four scalars. Both
`#[allow(clippy::too_many_arguments)]` attributes are gone; no function in the file now exceeds six
parameters, so the `alef.toml` entry is deleted.

## Behaviour

Unchanged, and checked rather than asserted. A temporary differential harness ran the same matrix
through `normalize_image_dpi_owned` on `main` and on this branch, **576 cases**: 3 target DPIs x 3
max dimensions x auto-adjust on/off x 2 min/max DPI pairs x 4 image sizes x 4 input DPIs, including
the invalid-input paths. It recorded output dimensions, buffer length, the full
`ImagePreprocessingMetadata` and the error text for every case. The two files are byte-identical.
The harness was removed before committing.

The one subtlety worth naming: the skip path records the unclamped `scale_factor` while the resize
path records the clamped `final_scale` that `calculate_new_dimensions` returns. The plan is rebuilt
with the clamped value on that path, which is why there are two `DpiPlan` bindings rather than one.

## Gates

Run against `e512a9cbe`:

- `cargo test -p xberg --lib --features ocr-pipeline` — 1788 passed, 0 failed
- `cargo clippy -p xberg --features ocr-pipeline --all-targets -- -D warnings` — clean
- `cargo fmt -p xberg -- --check` — clean

`ocr-pipeline` is the feature that gates `pub mod image`, so the default-feature build never compiles
this file.

**`poly lint` was not run locally.** poly panics at startup on `aarch64-apple-darwin`, on both
0.23.1 and 0.24.0, before it reads any file:

```
$ mkdir /tmp/t && cd /tmp/t && printf 'fn main(){}\n' > a.rs && poly lint a.rs
thread 'main' panicked at crates/poly-core/src/engines/astgrep/pack.rs:140:17:
poly's built-in ast-grep pack failed to parse (Fail to parse yaml as RuleConfig)
```

Same with the official release tarballs rather than the Homebrew bottle, in an empty directory with
no `poly.toml`. CI is unaffected, `Validate (poly)` is green on Linux. Flagging it rather than
claiming a green lint I could not produce; the parameter counts above are from parsing the
signatures directly.

## `poly.toml` deliberately not regenerated

`poly.toml` carries the same entry and is generated, so it would normally be refreshed here. It is
not, because `alef diff` reports **the same 467 files would change on a pristine `main` checkout as
with this branch** — an identical list, `poly.toml` among them. The generated tree is already drifted
before this change, which is what the `Alef-generated bindings freshness` job on `main` is failing
on. Regenerating would pull 467 unrelated files into a one-file PR, so `alef.toml` and
`.alef-toml-merge-provenance.toml` are updated at the source and `poly.toml` is left to whoever
clears the drift.

Worth noting the issue says to re-run `alef scaffold`; `poly.toml`'s own header says
`To regenerate: alef generate`. `alef scaffold` updates the provenance file but does not rewrite
`poly.toml`.

## Not covered

Only the `too-many-parameters` entry for this one file. No other baseline entries touched, none
added.

Refs #1567

### DIFF
--- a/.alef-toml-merge-provenance.toml
+++ b/.alef-toml-merge-provenance.toml
@@ -2249,11 +2249,6 @@ values = ["function-too-long"]
 
 [[entries]]
 relative_path = "poly.toml"
-key_path = "per-file-ignores.crates/xberg/src/image/preprocessing.rs"
-values = ["too-many-parameters"]
-
-[[entries]]
-relative_path = "poly.toml"
 key_path = "per-file-ignores.crates/xberg/src/keywords/yake/mod.rs"
 values = ["nesting-too-deep"]
 
--- a/alef.toml
+++ b/alef.toml
@@ -1150,7 +1150,6 @@ traineddata = "traineddata"
 "crates/xberg/src/extractors/wordperfect.rs" = ["file-too-long", "function-too-long"]
 "crates/xberg/src/extractors/xml.rs" = ["function-too-long"]
 "crates/xberg/src/heuristics/analyzer.rs" = ["function-too-long"]
-"crates/xberg/src/image/preprocessing.rs" = ["too-many-parameters"]
 "crates/xberg/src/keywords/yake/mod.rs" = ["nesting-too-deep"]
 "crates/xberg/src/keywords/yake/preprocessor.rs" = ["cyclomatic-complexity"]
 "crates/xberg/src/language_detection/mod.rs" = ["file-too-long"]
--- a/crates/xberg/src/image/preprocessing.rs
+++ b/crates/xberg/src/image/preprocessing.rs
@@ -111,37 +111,34 @@ pub(crate) fn normalize_image_dpi_owned(
     );
 
     let scale_factor = f64::from(target_dpi) / current_dpi;
+    let plan = DpiPlan {
+        original_dpi,
+        target_dpi,
+        scale_factor,
+        auto_adjusted,
+        calculated_dpi,
+    };
 
     if !needs_resize(width as u32, height as u32, scale_factor, config) {
-        return Ok(create_skip_result(
-            rgb_data,
-            width,
-            height,
-            original_dpi,
-            config,
-            target_dpi,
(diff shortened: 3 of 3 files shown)


Goldziher (approved):
Behaviour-preserving and well evidenced. The 576-case differential harness over `normalize_image_dpi_owned` covering both min/max DPI pairs and the invalid-input paths is the right way to prove a parameter-grouping refactor, and byte-identical output across the matrix is convincing. Grouping the five values into `DpiPlan` genuinely removes both `too_many_arguments` allows rather than relocating them, and dropping the `alef.toml` entry means the baseline actually shrinks.

Rebuilding the plan with the clamped `final_scale` on the resize path is the subtle part, and calling it out explicitly rather than leaving two similar bindings unexplained is appreciated.

One follow-up, not blocking: the `// The resize path records the clamped scale where the skip path records the unclamped one.` line carries exactly the kind of non-obvious rationale worth keeping, but it has no `~keep` marker, so poly's uncomment pass will strip it on the next run. I'll add the marker on main.

The red checks here are inherited from main (alef bindings freshness, generated e2e freshness, plugin bundle), not introduced by this branch — fixes for those are queued to land separately.

Note that #1567 stays open: this pays off one file of the baseline, not the whole thing.

Issue #1626, opened 2026-09-11
feat(embeddings): let a long-running process release embedding and reranker models it no longer uses


tobocop2:
## Problem

The embedding and reranker engines are kept in process-wide caches. An entry is added the first time a model is used or warmed, and nothing removes it. The crate has no public function to drop an entry, set an idle timeout, or limit the number of resident models.

A server that serves many embedding presets or reranker models over its lifetime keeps every model it has loaded in memory until the process exits. Its memory use grows with the number of distinct models, not with the current load, and the caller cannot bound it.

## Where

- `crates/xberg/src/embeddings/mod.rs:192` (1.1.5): `static ENGINE_CACHE`, an `AHashMap` behind a `RwLock`. `get_or_init_engine` (`:509-545`) inserts into it, and `warm_model` (`:565-583`) calls `get_or_init_engine`. No code removes an entry.
- `crates/xberg/src/embeddings/mod.rs:978` (1.1.5): `STATIC_ENGINE_CACHE`, the same pattern for static embeddings.
- `crates/xberg/src/reranking/mod.rs:132` (1.1.5): the reranker `ENGINE_CACHE`, also without removal.
- The public cache functions (`clear_processor_cache`, `clear_cache_directory`) do not touch these maps.

The caches are the same on `main` at 606fa7223, and no commit after v1.1.5 changes these files.

## Effect

A process that embeds with more than a few models needs enough memory for all of them at once. In a container with a memory limit, it is killed after enough distinct models are used, even when only one is in use at a time.

## How to reproduce

```rust
use xberg::{embed_texts, EmbeddingConfig};

fn main() -> xberg::Result<()> {
    for preset in ["fast", "balanced", "quality"] {
        let config: EmbeddingConfig =
            serde_json::from_value(serde_json::json!({"model": {"type": "preset", "name": preset}}))?;
        embed_texts(vec!["hello".to_string()], &config)?;
        // Read the resident set size here, for example from /proc/self/status.
    }
    Ok(())
}
```

The resident set size grows after each preset and does not go down. There is no call that returns the memory of a model that is no longer needed.

## Suggested fix

Add an eviction policy to the three caches, and expose it in the public API:

- A function that removes one model, or all models, from each cache (for example `evict_embedding_model` and `clear_model_caches`).
- An optional limit, either a maximum number of resident engines with least-recently-used eviction, or an idle timeout.

[...]

PR #1627, opened 2026-09-12
feat(embeddings): let a process release the models it no longer uses

tobocop2:
## Related

Addresses #1626

### The caches

Five process-wide engine caches grow forever: the ONNX and static (model2vec) embedding caches, the reranker cache, and the sparse-embedding and late-interaction caches, which use the same hand-rolled pattern. Each is now one generic wrapper over the `lru` crate behind a lock. The five copies of the read-lock, write-lock, poison-handling block are gone.

### The API

Each of `xberg::embeddings`, `xberg::reranking`, `xberg::sparse_embeddings` and `xberg::late_interaction` gains three functions. `evict_model` drops every engine loaded for one model, matched by repository and model file, and returns the number removed. `clear_engine_cache` drops every engine in that cache. `set_engine_cache_limit` bounds the number of resident engines. `xberg::clear_engine_caches` clears all of them. A caller that still holds an engine keeps it alive until it drops its handle.

### The bound

When a new engine would exceed the bound, the least recently used engine is dropped first. Lowering the bound drops engines at once. The embedding bound applies to the ONNX cache and the static cache separately.

### What stays the default

No bound. Nothing is evicted unless a caller asks. A cache stays locked while one engine is built, as before, so two callers for the same model wait rather than build twice.

### Bindings

The new functions are Rust-only, like the sibling `warm_model`. No binding manifest, header or generated package changes. The Rust API reference is regenerated with the pinned alef.

## Behaviour

Measured on Linux with the issue's own reproduction: the resident set size from `/proc/self/status` after embedding with the `fast`, `balanced` and `quality` presets, one process per arm.

| arm | start | fast | balanced | quality |
|---|---:|---:|---:|---:|
| base 606fa722, loaded | 14 MB | 127 MB | 800 MB | 2,868 MB |
| head 448f1820, loaded / after evict | 165 MB | 165 / 165 MB | 601 / 177 MB | 1,754 / 177 MB |

On the base RSS never falls. On the head RSS falls back to the starting level after each eviction. Whether the allocator returns every page is up to the allocator, so the engine drop is also proven directly: `evict_model` returns 1, then 0, and a fresh call loads the model again and returns 1. A unit test holds a weak handle to a cached engine and asserts it is gone after eviction.

Tests: 7 unit tests on the cache (a bound of N with N+1 engines keeps N and drops the least recently used, eviction, clear, limit changes, handle survival), 6 offline tests on the public functions, and 5 live tests over real presets (ONNX, static and reranker), registered in the live-HF CI job. On the base the live test target does not compile (`E0425`, the functions do not exist). On the head all pass.

## Not covered

The HTTP, MCP and CLI `cache warm` entry points get no eviction sibling. Each carries generated docs and schema, and a server-side eviction endpoint is a product decision on its own.

An idle timeout is not added. Without a background thread it could only fire on the next call, which never comes while a process is idle. A host that wants one runs its own timer and calls `clear_engine_cache`.


Goldziher:
Merged as `31943874c2`, shipping in **1.1.6** rather than 1.2.0 — this repo ships additive and even breaking changes in patch releases with a changelog label, so there was no reason to hold it.

Taken over as described in the issue. Your commit is on `main` with your authorship intact; I rebased it onto current `main` and dropped your regen commit, since the release does one `alef generate` at the end and it would only have conflicted again. The only hand-merge was the changelog, where your `### Added` entry moved from `[Unreleased]` into the 1.1.6 `### Added` block.

What I verified before merging, all with `RUSTFLAGS="-D warnings"` since CI denies warnings:

| leg | result |
| --- | --- |
| `--features pdf` (no engine features at all) | ok |
| `--features no-ort-target` | ok |
| `--features wasm-target` on `wasm32-unknown-unknown` | ok |
| the live-test feature set, `--tests` | ok |

The wasm leg was the one I was worried about: `static-embeddings` is a member of `no-ort-target` and therefore of `wasm-target`, so `lru` does reach the wasm build. It is clean — `lru 0.18` pulls only `hashbrown`/`foldhash`, and I checked it compiles for `wasm32-unknown-unknown` standalone as well. I also broke `engine_cache_live.rs` on purpose once to confirm the `--tests` leg was actually compiling it rather than skipping it.

Two things in your write-up that were worth the review time and that I have kept:

- Rejecting `moka` because its sync cache reads `std::Instant` on every lookup, which panics on wasm32.
- Measuring that a feature named in a function's own `cfg` gets copied into every binding manifest, and putting the gate on the module declaration instead so five binding manifests don't gain `static-embeddings` in their defaults. That is not obvious from anything written down, and it is the kind of thing that would have shipped silently.

`get_or_try_init` holds the write lock across `init`, so two different models still load one at a time, same as the code it replaces. Worth revisiting if model load latency ever shows up in a profile, but it is not a regression.

Thanks — this was careful work.

PR #1642, opened 2026-09-15
fix(reranker): retain permit after cancellation

### LINKED ISSUE #1641: fix(reranker): retain concurrency permit after caller cancellation
## Problem

Dropping or timing out `rerank_async` releases its concurrency semaphore permit while the already-started `spawn_blocking` model load or inference continues. Repeated abandoned calls can therefore exceed the configured reranker concurrency bound and increase resident memory after callers have returned.

This is the upstream library portion of xberg-io/xberg-enterprise#2332.

## Expected behavior

The semaphore permit remains owned by the blocking task until that task actually finishes, including when the async waiter is cancelled. The accepted path and public API remain unchanged.

Goldziher:
A cancelled `rerank_async` waiter now transfers its owned concurrency permit into the already-started blocking task, so abandoned model loading or inference remains within the configured concurrency bound until it actually finishes. A deterministic regression aborts the waiter after the blocking closure starts, confirms the permit stays unavailable, and confirms it returns when the closure exits.

Fixes #1641. Upstream portion of xberg-io/xberg-enterprise#2332.

Static verification: `cargo fmt --all -- --check`; `poly lint crates/xberg/src/reranking/mod.rs`; `git diff --check`. Local tests and builds were intentionally not run per coordination instructions.


Goldziher:
Fixed the cross-platform CI compile failure in 042c35271b8 by retaining the must-use semaphore permit returned by the cancellation regression. The change is test-only and preserves the intended assertion; CI restarted from the new head.

Goldziher:
Refreshed the Alef generation-input fingerprint after the rerank timeout change. A full Alef generation completed and produced no binding-content changes; only the authoritative fingerprint changed.

Goldziher:
Reviewed head `962695caaaecd67fb51722ccd8a3bb4a784bc5ab`. The owned semaphore permit is transferred into the blocking closure, so dropping or aborting the async waiter cannot release capacity while model load or inference is still running; normal return and panic both drop the permit. The deterministic regression proves permit retention after waiter abort and release after the blocking body exits. This preserves the concurrency bound without added allocation beyond the required owned permit. No correctness or performance blocker found.

Goldziher:
Superseded by 2f765391fa on main, which landed the same fix (`acquire_owned` with the permit moved into the `spawn_blocking` closure) plus a regression test that was falsified before being trusted — moving `let _permit = permit;` out of the closure makes it fail with its intended message.

Closing in favour of that commit; the branch had also gone conflicting. Shipping in 1.2.3. Thanks.

PR #1646, opened 2026-09-15
feat(ner): expose shared backend loader

Goldziher:
Expose xberg's existing keyed GLiNER cache through an async public loader so external services reuse one backend without blocking Tokio workers. Initialization failures remain retryable, and an opt-in real-model regression requires concurrent successful loads to return the same `Arc`; this unblocks xberg-io/xberg-enterprise#1842.

Local tests and builds were intentionally not run per user instruction; `cargo fmt --check`, path-scoped `poly fmt --check`, path-scoped `poly lint`, and `git diff --check` pass.


Goldziher:
Critical review complete at 761fc470f. The public async API delegates model initialization to the blocking pool while retaining the existing double-checked keyed cache, so concurrent callers share one Arc and failures remain retryable. The follow-up now documents all error classes, reports JoinError accurately, and source-tests both reuse and retry-after-error. I found no correctness, memory, or contention blocker; CI and tobocop2 review remain the gates.

Goldziher:
Reviewed head `fe8ca9924856ca6ae9add25d6ac2bd6fdeb4c506`. Warm cache hits now return synchronously from the keyed process cache without occupying Tokio's blocking pool; cold initialization remains double-checked under the write lock and initialization failures are not cached. The public async API, error propagation, reuse test, and retry-after-error test are consistent. No remaining correctness or performance blocker found.

Goldziher:
Addressed the generated-E2E CI failure at `af672350a61`. The shared loader is a Rust-only API returning `Arc<GlineBackend>`, so it now carries the repository-standard `#[cfg_attr(alef, alef(skip))]` instead of being lossy-sanitized into binding fixtures; its public contract documentation is also protected with `~keep`. The failing `task e2e:generate` negative control reported `lossy_sanitized_surface` before the change and completes cleanly after it with no generated artifact diff. No local tests or build completed.

Goldziher:
Ubuntu CI failure is runner infrastructure, not a source/test failure. Job 104629913418 completed every substantive step through the workspace tests and feature checks; its sole check annotation is `System.IO.IOException: No space left on device` from the Actions runner while writing `_diag/Worker_...log`. I’ll rerun failed jobs after workflow run 35044011234 finishes.

Goldziher:
Merged directly to `main` at `fa0a33ad2a3a21c6a7390fd8b526be05248e2f91` (this project works directly on main, not via the GitHub merge button). The CHANGELOG.md conflict against the concurrent 1.2.3 release notes was resolved by folding the NER entry under the existing `## [1.2.3] → ### Added` heading.

PR #1647, opened 2026-09-15
feat(ocr): export per-page coordinate frames

### LINKED ISSUE #1645: feat(ocr): export per-page coordinate frames with OCR elements
## Problem

Mixed-page PDF extraction exports public `ocr_elements` after normalizing document-node geometry to page points. Those OCR elements retain backend raster coordinates, but the per-page processed raster dimensions stay only in the page-local OCR result and are lost when elements are attached to the final document.

A multi-page consumer therefore cannot normalize an element safely: document-level `metadata.additional.ocr_processed_image_width/height` is not authoritative for differently sized, preprocessed, or rotated pages. Enterprise issue xberg-io/xberg-enterprise#1967 currently withholds those boxes rather than inventing coordinates.

## Expected behavior

Final document metadata should expose one authoritative coordinate-frame record per OCR page with public elements. Each record includes `page_number`, processed raster `width` and `height`, `unit: "pixel"`, and `origin: "top_left"`; consumers join it to an element through the element's existing `page_number`.

The frame must be captured before page-local metadata is discarded and survive serialization without requiring a new top-level binding type. Store one record per page rather than duplicating the same JSON object on every word-level element.

## Regression coverage

Prove two pages can expose distinct frames, frames are ordered by page number, and invalid/absent dimensions do not fabricate one.

Goldziher:
PDF OCR now exports `metadata.additional.ocr_page_coordinate_frames`, with one typed pixel/top-left frame per page joined to public OCR elements by `page_number`. A private one-element-per-page transport avoids duplicating frame objects for every OCR word and is removed before serialization.

Adds exact coverage for distinct page frames, invalid dimensions, ordering, transport cleanup, and preservation of unrelated backend metadata. Local test/build execution was intentionally skipped; static Rust formatting, targeted poly lint/format, changelog mirror, and diff checks pass.

Fixes #1645
Related: xberg-io/xberg-enterprise#1967


Goldziher:
Reviewed head `66d6012c10c5814e3ff47b5ba183f20aae896884`. The frame transport is O(pages), carries typed pixel/top-left dimensions from the effective OCR layout, strips private transport metadata before serialization, sorts deterministically, and withholds mismatched or conflicting page carriers while preserving valid neighboring pages. The added regression covers the exact embedded-page mismatch. No remaining correctness or performance blocker found.

Goldziher:
ARM CI exposed a test-isolation race unrelated to the OCR frame code: `failed_explicit_builtin_registration_preserves_suppression` lost the global post-processor registry to a concurrent extraction and unwrapped the documented retryable `registry is in use` error. The same job passed 8,019 other library tests, including the OCR frame regressions.

Head `6295cfee1cc` now uses the existing bounded registry-mutation retry helper for that test's setup, matching the adjacent lifecycle tests. No production code changed. Per the current repository instruction, I used static diff checks and left execution to CI.

Goldziher:
Superseded by b593d145e5 on main, which landed per-page OCR coordinate frames sourced from the same `resolved_ocr_layout_dimensions` call the elements depend on, so frame and elements cannot disagree.

One thing from this branch was NOT superseded and I've taken it separately: `failed_explicit_builtin_registration_preserves_suppression` still used bare `.unwrap()` on the two registry mutations while the other 13 call sites in that file already go through `retry_while_registry_in_use`. That inconsistency is a real flake source under registry contention. Thanks for catching it.

PR #1648, opened 2026-09-16
fix(ocr): recover isolated invoice quantities

Goldziher:
Repairs split invoice headers using bounded table geometry and retries one isolated blank QTY cell from its source pixels when the surrounding column has strong integer support. The retry is OCR-owned, clamps to adjacent row and column midpoints, and never derives a quantity from monetary values; non-candidate tables keep the existing fast path.

Adds focused source regressions for captured invoice geometry, supported-column preservation, sparse and centered headers, intentionally blank quantities, and tight-cell crop bounds. Static validation: rustfmt --check, poly lint on the three changed files, and git diff --check; local builds and tests were intentionally not run per the repository task instruction.

Addresses xberg-io/xberg-enterprise#2150.


Goldziher:
Critical source review at `343ef8ef2f`: the fix reads pixels rather than inferring a business value. It performs at most one extra `PSM_SINGLE_WORD` recognition per qualifying table, only for a literal QTY column with exactly one blank and at least two integer supports; non-candidates avoid row detection and the retry allocation path. Both crop axes are bounded by adjacent row/column midpoints, recovered text must be a short positive integer above the existing table confidence threshold, and the full-page rectangle/PSM are restored afterward. Iterator elements are captured before targeted recognition so the retry cannot replace full-page element state. Header repair is limited to nearby UNIT→PRICE and LINE→TOTAL fragments in a QTY table, retains correlated column positions, and has unrelated/far/supported-column negatives. I found no correctness or performance blocker; CI and tobocop2 remain the gates.

Goldziher:
Fixed the Windows Rust 1.95 clippy failure at baef8c80e by replacing the two manually guarded divisions with checked_div while preserving zero-count fallbacks and overflow behavior. Static diff validation is clean; local build/test hooks were bypassed to honor the current no-local-build/test instruction. CI is the runtime gate.

Goldziher:
Fixed the macOS dead-code failure at the new head by compiling the compatibility grid-only wrapper only for its actual consumers: PDF, PaddleOCR, and unit tests. The shared column-aware implementation remains available for OCR paths. Static diff validation passed; CI is the runtime gate per the no-local-build/test instruction.

Goldziher:
The downstream Enterprise Engine regression exposed that the real retry crop was correct but PSM_SINGLE_WORD read the isolated quantity as low-confidence text. This update uses PSM_SINGLE_LINE, which preserves multi-digit quantities, and adds a real Tesseract regression using bundled tessdata plus a metadata-free 170x91 fixture. The negative control fails under PSM8; PSM7 recovers both 5 and synthesized 555. Fresh verification: all five blank_quantity_retry tests passed; scoped poly lint/format and diff checks passed. Please review.

Goldziher:
Merged directly to `main` at `4ca9f6477d9cc23b4ae1d283cdbec618b2d0bf68` (this project works directly on main, not via the GitHub merge button).

Before merging: the QTY blank-cell retry gating was verified narrow as described — literal case-insensitive `QTY` header, ≥2 recognized integer quantities, exactly one blank cell, pixel reads clamped to that cell's own bounds, never derived from price/total. All `blank_quantity_retry_*` tests pass, including the real-pixel-recognition test under `--features ocr,bundle-tessdata-eng`.

Also found and fixed a stale test expectation in `table_core.rs` unrelated to that gating: `reconstruct_table_attaches_header_fragments_to_nearest_supported_column` asserted a merged UNIT/PRICE column would keep the PRICE header word's own x (2219), but `assign_words_to_cells` never uses that value — it already assigns every word, including the header word itself, to whichever column `detect_columns" anchored closest, which for this fixture is the \$-value data cluster at 2206 (13px away) rather than the isolated header token at 2096 (123px away). That anchor also drives the blank-quantity retry's `cell_axis_bounds` crop for an adjacent cell, so it needs to track where the data actually sits. Fixed in `5c33c7a3ff` with the test failing before the change and green after, confirmed against the unmodified PR branch (`table_core.rs` was byte-identical pre-fix).

OCR quality change ships without a corpus-wide A/B measurement, as expected.

PR #1652, opened 2026-09-16
feat(pdf): expose per-page coordinate frames

Goldziher:
Native PDF extraction now publishes each page’s MediaBox lower-left origin and extent in `metadata.additional.pdf_page_coordinate_frames`. This is the upstream prerequisite for normalizing native hierarchy citation geometry correctly when a PDF page does not start at `(0, 0)`.


Goldziher:
Superseded by #1653 and #1654. Per current scope, upstream implementation will be handled in xberg rather than from the enterprise delivery work.

Issue #1639, opened 2026-09-15
bug: legacy `.ppt`: the slide list the persist-chain fix reads is the **first** one in the stream, i.e. the oldest save's


erichevers:
## Summary

`live_slide_offsets` (`crates/xberg/src/extraction/ppt/mod.rs`, GH#1614)
resolves the persist directory the right way — newest save first, `or_insert`
so an older save never clobbers a live id — and then asks
`slide_persist_ids_in_presentation_order` for the slide order. That function
walks the `PowerPoint Document` stream **from offset 0** and returns the
**first** `SlideListWithText` (`recInstance` 0) it meets.

A `.ppt` stream is append-only across saves, and every save writes a new
`Document` container with its own `SlideListWithText`. The first one in the
stream is therefore the **oldest** save's list. On a deck saved more than once
the extractor numbers — and extracts — the slides the *first* save had:

- a slide added in a later save is not extracted at all;
- a slide deleted in a later save is extracted as a live slide;
- the ids of the first list resolve through the *newest* directory, so what
  comes out is the newest revision of an old slide set — internally
  consistent, and not the presentation.

The live list is reachable from the chain the function already walks: the
`UserEditAtom` it starts from carries `docPersistIdRef`, the persist id of the
live `Document` container, and the directory maps it to an offset. The slide
list inside *that* container is the presentation's.

## Measured

Almost a third of our legacy decks carry more than one `SlideListWithText` (saved more than once).
On all decks the extractor's slide count equals the **first** list's
count; on almost all multi-save decks that is not the live list's count.

One 46-slide deck comes out as 15 slides: its first
save had 15, and the 31 slides added over the next seven saves are not
extracted.

## Reproduction, from the records

Two saves, the second adding a slide:

```
PowerPoint Document
  Slide A                         (persist 4)
  Document #1 > SlideListWithText { SlidePersistAtom 4 }        <- first in the stream
  PersistDirectoryAtom #1  { 4 -> Slide A, 1 -> Document #1 }
  UserEditAtom #1          { offsetLastEdit 0, docPersistIdRef 1 }
  Slide B                         (persist 5)
  Document #2 > SlideListWithText { SlidePersistAtom 4, SlidePersistAtom 5 }
  PersistDirectoryAtom #2  { 5 -> Slide B, 1 -> Document #2 }
  UserEditAtom #2          { offsetLastEdit -> UserEditAtom #1, docPersistIdRef 1 }
Current User
[...]

PR #1661, opened 2026-09-18
feat(ppt): extract embedded OLE objects from legacy decks

### LINKED ISSUE #1660: bug: legacy `.ppt`: an embedded OLE object (a Word or Excel table inserted as an object) is extracted in no form
## Summary

A legacy binary `.ppt` that carries a table as an **embedded OLE object** — a
Word document or an Excel sheet inserted as an object, the way PowerPoint 97
through 2003 decks routinely carry tables — loses that table completely. The
object's native data is in the file, in a format the extractor already reads
(`WordDocument` → the `.doc` extractor, `Workbook` → the `.xls` one), and its
presentation picture is in the file too, as an EMF or WMF metafile. The
extractor reads neither: the metafile is skipped by design in the `Pictures`
walk, and the embedded-object records are not handled at all. The slide comes
out with its title and no content.

For `.pptx` the same thing works: `ooxml_embedded.rs` reads `ppt/embeddings/`
and recursively processes every embedded file. The legacy path has no
counterpart.

## What a file carries

In one of our documents we have 14 slides, 115 KB. Seven of its slides are
a title over a table. Walking the container:

```
/Pictures                    21 362 bytes — 9 OfficeArtBlip records, all metafiles:
                             7× OfficeArtBlipEMF  (recType 0xF01A, recInstance 0x3D4)
                             2× OfficeArtBlipWMF  (recType 0xF01B, recInstance 0x216)
                             0 raster blips (JPEG 0xF01D / PNG 0xF01E / DIB 0xF01F)

/PowerPoint Document         63 443 bytes; among its records:
   ExObjListContainer        0x0409  ×1     the document's list of external objects
   ExEmbedContainer          0x0FCC  ×7     one per embedded object …
     ExOleObjAtom            0x0FC3  ×7     … with exObjId, persistIdRef, oleObjType
   ExOleObjStg               0x1011  ×7     the object's storage, recInstance 0x001 = compressed
   ExObjRefAtom              0x0BC3          in each object shape's ClientData: exObjId → ExEmbed
```

An `ExOleObjStg` with `recInstance` 1 is `ExOleObjStgCompressedAtom`
(MS-PPT 2.10.35): a 4-byte `decompressedSize` followed by a zlib stream. The
seven records here decompress to 53 760–62 976 bytes each — every one a
complete OLE compound file (`D0 CF 11 E0 A1 B1 1A E1`) with the streams

```
WordDocument   1Table   \x01CompObj   \x05SummaryInformation
```

— **embedded Word documents**, one table each. Handed to the extractor as a
`.doc`, object #1 yields

```
COMMON NAME	AGGREGATION STATE	USES		hexogen	Explosive solid	Explosive (half of semtex), rodenticide		Iron pentacarbonyl	…
```

which is the "10 COMPOUNDS, the first 5" table of slide 5. Object #5 yields
the "DATABASE ANALYSIS, the first 5" table (`(search terms) and py<2001 |
Toine | Med | Number of hits | Internet | SPN …`). The data is there,
and it is readable with what the crate already has.

The metafiles in `Pictures` are these objects' presentation pictures — what
PowerPoint draws on the slide in place of the live object. The object shape's
`OPT` carries `pib` (the blip index), and its `ClientData` carries the
`ExObjRefAtom` that names the `ExEmbed`; the `ExOleObjAtom.persistIdRef` is
resolved through the persist directory to the `ExOleObjStg`.

## What the extractor does with it

Two gaps, both deliberate and both stated in the source.

**1. Metafile blips are skipped.** `extract_pictures_from_stream`
(`extraction/ppt/mod.rs`) dispatches on the blip record type:

```rust
let format: Option<Cow<'static, str>> = match rec_type {
    RT_BLIP_JPEG | RT_BLIP_JPEG_ALT => Some(Cow::Borrowed("jpeg")),
    RT_BLIP_PNG => Some(Cow::Borrowed("png")),
    RT_BLIP_DIB => Some(Cow::Borrowed("dib")),
    _ => None,
};
```

and its doc comment says why: *"Metafile blips (EMF/WMF/PICT) use a
different, larger header and are not raster images; they are skipped."* A
metafile blip (MS-ODRAW 2.2.24–2.2.26) is `rgbUid` (16 bytes; a second 16
when the low bit of `recInstance` is set) followed by an
`OfficeArtMetafileHeader` (34 bytes: `cbSize`, `rcBounds`, `ptSize`,
`cbSave`, `compression`, `filter`) and then the metafile bytes, zlib-deflated
when `compression` is 0. So no `ExtractedImage` is produced, and nothing
downstream — a VLM caption, an OCR pass — ever sees the table.

**2. Embedded objects are not read.** The `ppt` module names no
`RT_EXTERNAL_OLE_*` record at all: `ExObjList`, `ExEmbed`, `ExOleObjAtom`,
`ExOleObjStg` and `ExObjRefAtom` are all walked over as opaque bytes. The
OOXML side does have this — `extract_ooxml_embedded_objects` opens
`ppt/embeddings/`, identifies each file and recursively extracts it into the
document's `children` — which is exactly the shape the legacy path is missing.

The slide therefore reaches the element stream as its title alone. On our deck
that is seven of fourteen slides: 5, 6, 7, 8, 9, 13 and 14 —
the compound lists, the database analysis and the host analysis the whole
talk is about.

## Where it shows in our corpus

Half the legacy corpus carries at least one object of this kind, and every
metafile pictures and object storages is lost. Most decks
carry both forms side by side.

## What would close it

In order of value:

1. **Read the embedded object's native data.** For each `ExEmbedContainer`,
   resolve `ExOleObjAtom.persistIdRef` to its `ExOleObjStg`, decompress
   (`recInstance` 1: 4-byte size + zlib; `recInstance` 0: raw), open the
   inner compound file and dispatch on its root streams — `WordDocument` to
   the `.doc` extractor, `Workbook`/`Book` to the `.xls` one, `PowerPoint
   Document` to `.ppt` — attaching the result the way the OOXML path attaches
   `ppt/embeddings/`, with the slide number of the shape that references it
   (`ExObjRefAtom.exObjId` in the shape's `ClientData`). A table then comes
   out as a table, which is what the author put there.
2. **Emit the metafile presentation picture as an image**, rasterised, so a
   caller with a vision model can still read a table whose object storage is
   missing or unreadable (a linked rather than embedded object,
   `ExLinkContainer`, keeps only the picture). This needs an EMF/WMF renderer
   and is the fallback, not the fix.

Reading the storage (1) is the one that matters here: on every deck measured
the objects are embedded, not linked, and their storages decompress cleanly.

## Verifying without a fixture

Any PowerPoint 97–2003 deck with a Word or Excel table inserted as an object
shows it

### Steps to reproduce

see above

```text

```

MannXo:
## Related

Fixes #1660

## Description

A legacy `.ppt` carrying a table as an embedded OLE object lost it completely. The object's
bytes are in the file, in a format this crate already reads, but the `ppt` module named no
external-object record at all, so `ExObjList`, `ExEmbed`, `ExOleObjAtom` and `ExOleObjStg`
were all walked over as opaque bytes.

The `.pptx` path has had this since #78/#307. This gives the legacy path the same contract.
Recovered objects are extracted recursively and attached to `ExtractedDocument.children`,
capped by `max_embedded_file_bytes` and gated by `max_archive_depth`.

### What this reuses rather than rebuilds

The issue proposes dispatching the inner compound file on its root streams. That code
already exists. `extraction::ooxml_embedded::extract_ole_embedded_object` maps
`WordDocument` to `application/msword`, `Workbook`/`Book` to `application/vnd.ms-excel`, and
a `Package` stream to its sniffed type. It was private to the OOXML path; this makes it
`pub(crate)` and calls it. The new code is therefore only the part that is genuinely
`.ppt`-specific, which is finding the bytes.

### Why resolution goes through the persist directory

`ExOleObjAtom.persistIdRef` names the storage, and the persist directory turns that into a
stream offset. Scanning the stream for `0x1011` records would be shorter and is wrong. A
`.ppt` is append-only across saves, so editing an embedded object leaves every superseded
copy of its storage in the bytes, and a scan cannot tell which is current. That is the same
defect #1639 fixed for slide lists in this module last week.
`a_superseded_object_storage_revision_is_not_extracted` pins it.

For the same reason there is no fallback. If the persist chain does not resolve, no objects
are extracted, which is what `live_slide_offsets` already does rather than produce a partial
answer.

### Numbering

Objects are pathed `embedded-object-N` by the position the deck's external-object list
declares them at, not by how many turned out to be readable. An object whose storage fails
still consumes its number, so one bad object does not rename the documents after it.

### Commits

1. `refactor(ppt)` splits `persist_chain` and `live_document_payload` out of
   `live_slide_offsets`, which was doing both jobs. No behaviour change.
2. `feat(ppt)` adds the recovery and the wiring.

## Verification

Same command, one revision apart, on a deck built by LibreOffice 26.2.4 carrying one
embedded Calc sheet and one embedded Writer document:

```
upstream/main   === CHILDREN: 0 ===

this branch     === CHILDREN: 2 ===
                --- embedded-object-1 [application/vnd.ms-excel] ---
                Sheet1
                COMPOUND STATE
                hexogen Explosive solid
                --- embedded-object-2 [application/msword] ---
                DATABASE ANALYSIS the first two rows
```

The deck's own slide text is byte-identical on both.

Gates, all run locally:

```
cargo fmt --all -- --check                                         clean
cargo clippy -p xberg --features office,excel --all-targets        clean
cargo clippy -p xberg --features full --all-targets                clean
cargo test --locked -p xberg --features office,excel --lib ppt     147 passed, 0 failed
cargo test ... --test issue_1660_ppt_embedded_ole_objects          4 passed, 0 failed
python3 scripts/ci/check-feature-parity.py                         5 aggregates at parity
cargo check --no-default-features                                  clean
cargo check --no-default-features --features office                clean
cargo check --no-default-features --features email                 clean, 2 warnings
cargo check --no-default-features --features office
             --target wasm32-unknown-unknown                       clean
```

The two `--features email` warnings are dead code in `types/builder.rs` and are identical on
`upstream/main` with the change stashed, same two, same count.

The `--features full` run needed `TSLP_OFFLINE=1` here. That is my connection, not the
change: `tree-sitter-language-pack`'s build script downloads a 36 MB
`parser-sources-1.20.0.tar.zst` and my link gets about 9 MB of it before its 6 retries run
out. Nothing in this PR touches tree-sitter. The `office,excel` run above needs no such
flag and covers every line changed here.

Both commits build and test on their own, so the branch is bisectable: the refactor commit
alone is 137 passed, 0 failed.

With the source change stashed and the tests kept, `should_attach_an_embedded_ole_object_as_a_child_document`
and `should_warn_and_continue_when_an_object_cannot_be_identified` fail while the other two
still pass, so the green is measuring the change rather than the configuration.

`office` gains `dep:flate2`. That compiles nothing new anywhere: `dep:zip` in the same
feature already builds flate2 through `deflate-flate2`, and `cargo tree -i flate2` for
`wasm-target` on `wasm32-unknown-unknown` shows it arriving through lopdf, png and image
regardless.

### API surface

`PptExtractionResult` gains one field and there is a new `PptEmbeddedObject`. Both are
`#[cfg_attr(alef, alef(skip))]`, so no binding target regenerates. `f30b832061` added
`images` to the same struct the same way, and the struct has exactly one construction
site, inside the crate.

## What is not here

- **The issue's item 2, rasterising the EMF/WMF presentation pictures.** It needs a metafile
  renderer this crate does not have, and the issue calls it the fallback rather than the fix.
- **Attributing an object to the slide that displays it.** `ExObjRefAtom` is the documented
  route. On every deck I could generate it carried the same `exObjId` on slides holding no
  object at all, and never the second object's id, so I could not verify the mapping and did
  not want to ship an unverified one. The OOXML path does not attribute either, so children
  are unattributed on both. Glad to add it against a deck where those refs are meaningful.
- **A corpus fixture.** `test_documents/ppt/simple.ppt` has no embedded objects at all: 65
  `ExObjRefAtom`, 12 hyperlink atoms, zero `ExOleObjStg`. It now has a regression test
  asserting it still reports none, which is also what guards against a looser walk inventing
  objects from hyperlink references. The tests assemble their own decks. I have the two
  LibreOffice-generated `.ppt` files used above if you would like them in the corpus; I
  cannot publish binaries myself.

## Checklist

- [ ] CI passing (every gate green locally, listed above; this ticks when the run lands)


MannXo:
closing this. nice one on 0x0BC1, i had that wrong.

Issue #1254, opened 2026-07-14
feat: automatically OCR pages whose embedded text is unreadable, instead of returning garbled text


tobocop2:
### What is the proposed feature?

Route pages whose text layer is **undecodable** to OCR, the same way we already route scanned pages.

Today `pdf/scan_detect.rs` decides scanned-vs-native from raster coverage, invisible-text ratio, glyph count, codec and producer prior. That catches the case where there is *no usable text layer*. It does not catch the case where there **is** a text layer, it is visible, it is glyph-rich — and it decodes to garbage.

A PDF hits this when its font gives no path from glyph to character. Concretely, all of:

1. `Type0` with `/Encoding /Identity-H`, `CIDSystemInfo` ordering `Identity` — the content-stream codes are CIDs;
2. `CIDToGIDMap /Identity` — CID == GID, so the codes are **raw glyph indices**;
3. no `/ToUnicode` CMap;
4. the embedded `CIDFontType2` subset has **no `cmap` table and no `post` table**.

The subset legitimately has no `cmap` — a CIDFontType2 is addressed by glyph index directly, so rendering never consults one, and subsetters strip both it and `post` to save bytes. What is left in the file is glyph *outlines* and glyph *indices*. Every route to Unicode is severed: no `/ToUnicode`, no `cmap` to invert, no `post` glyph names, and `Identity` ordering means there is no predefined CMap either.

The page renders perfectly in a viewer. The text simply cannot be extracted — **by anyone**. On a real failing page, every extractor produces the same mojibake:

| extractor | output |
|---|---|
| pdf_oxide 0.3.74 | `81,7('67$7(6',675,&7&2857` |
| poppler (`pdftotext`) | `81,7('67$7(6',675,&7&2857` |
| pypdf | `81,7('\x0367$7(6\x03',675,&7\x03&2857` |
| pymupdf | `81,7('\x0367$7(6\x03',675,&7\x03&2857` |

Ground truth, as rendered: `UNITED STATES DISTRICT COURT`.

This is not a pdf_oxide bug — no conformant reader can do better. But the *signal* that the layer is undecodable is missing, and without it we cannot act.

**Why `scan_detect` misses it.** Score these pages against `PageScanSignals` and they look maximally native:

| signal | value on an undecodable page |
|---|---|
| `image_coverage` | ~0 — there is no raster at all |
| `invisible_text_ratio` | ~0 — the text is drawn visibly |
[...]

PR #1678, opened 2026-09-19
fix(pdf): repair the provenance lookup and route fabricated mappings to OCR

### LINKED ISSUE #1667: bug: PDF: a text layer that decodes to shifted glyph codes passes the OCR quality gate with quality_score 1.0
### Description

A PDF whose embedded fonts use a custom encoding decodes its letters to byte values in the 0x74 to 0x8f range, while digits and punctuation decode normally. Native extraction returns that garbage on every page. With the default OCR config (tesseract, per-page quality gate on) the result is `extraction_method = "native"`, `ocr_used = false`, `quality_score = 1.0`, no `text_quality` warning, and tables are reported. The gate did not fire, the grade calls the text clean, and the caller gets unreadable output with no signal.

`force_ocr = true` on the same file returns readable text, so OCR is the right route for it.

Expected: a page whose text layer is mostly letters mapped outside the printable ASCII range, or mostly one- and two-character tokens, fails the per-page gate and goes to OCR. The `quality_score` for such text must not be 1.0. The `text_quality` grade should fire too: it counts alphanumerics, and bytes 0x74 to 0x8f are not alphanumeric, but the digits and the many short tokens keep the ratio above its threshold.

### Steps to reproduce

1. Build a PDF with a Type 1 or TrueType font whose custom encoding maps the letters to codes above 0x73 and leaves digits and punctuation in place.
2. Extract with the default OCR config and `pages.extract_pages = true`.
3. Read `extraction_method`, `quality_score` and `processing_warnings`.

xberg v1.2.4, release build, 2026-09-18. Config: `{"pages":{"extract_pages":true},"include_document_structure":true}` with the default tesseract OCR config filled in. Per-page gate statistics were not logged at info level; the `xberg::pdf::ocr` debug target carries them.

tobocop2:
## Related

Refs #1667

## Description

The per-page text run provenance lookup read a span's mapping provenance after the extractor had already renamed the span's font to the resolved `/BaseFont`, but the provenance table stays keyed by the raw resource alias. The lookup therefore missed for virtually every real font (any font whose `/BaseFont` differs from its resource alias), so the fabricated-mapping signal (`MappingProvenance::Fallback`, issue #1254) was empty everywhere it mattered.

## Behaviour

The lookup now runs in the same pass and is keyed by the resource alias, and the default `Auto` OCR strategy routes a page to OCR when any of its text runs carry a fabricated mapping (`MappingProvenance::Fallback` — a font with no usable mapping tier at all: no `/ToUnicode`, no predefined CID collection, no embedded cmap, no simple-font encoding). A page whose font resolves through a mapping tier, even a wrong one, is unaffected.

This does **not** fix #1667's reported document: that font resolves through its own embedded `/ToUnicode`/`/Encoding` to the wrong-but-ordinary characters, which lands on a resolved provenance tier (`EncodingName`), never on `Fallback`. Detecting a mapping that resolves but resolves wrong needs a different signal and is tracked separately in #1696.

## Checklist

- [x] Mutation arms shown red and restored
- [x] False positive arm: clean subset fonts, and a Type0 font with a genuine `/ToUnicode` map, are not routed to OCR
- [x] fmt and clippy clean


Goldziher (changes requested):
Verified the fix logic and it's correct. Not merge-ready due to a CI-blocking regen gap.

**Correctness — confirmed:**
- (a) Real bug on base: `crates/xberg-native-pdf/src/extractors/text/run.rs` looked up provenance by `span.font_name` *after* renaming it to the resolved `/BaseFont`, against `self.fonts` which is keyed by the raw resource alias (`add_font`/`add_font_shared`). Missed for any font whose BaseFont differs from its alias. Fix moves the lookup into the same pass, keyed by the alias, before the rename.
- (b) Call path traced: `apply_fabricated_provenance_pages` (new, `ocr/scoring.rs`) is called from the `Auto` per-page gate in `extractors/pdf/mod.rs` (~L2082), forcing `decision.fallback = true` when `fabricated_text_pages` is non-empty. This flips `evaluate_ocr_skip_gate`'s `has_substantive_doc && !decision.fallback` branch (which used to `SkipSubstantive`) into `RunFallback`/`RunFallbackOnPages`.
- (d) No missed call sites: `OcrStrategy` only has `Auto`/`ScannedPages`. `ScannedPages` already sees this signal because `pdf/native/metadata.rs` unconditionally unions fabricated pages into `scanned_pages` (unrelated to strategy). `disable_ocr`/`force_ocr`/`force_ocr_pages` are explicit user overrides where the signal is moot. The `pdfium` backend does no OCR gating at all (`ocr_used: false` hardcoded) — out of scope.
- (c) Red/green proven myself, reverting each production hunk independently and restoring via `git show HEAD:<path> > <path>`:
  - Revert of the `mod.rs` call site → `cargo test -p xberg --features pdf,ocr test_auto_strategy_routes_fabricated_provenance_page_to_ocr`: `FAILED ... panicked at crates/xberg/src/extractors/pdf/mod.rs:5955: the fabricated native text must not survive to the final content`.
  - Revert of the `run.rs` lookup fix (restored the old post-rename/wrong-key lookup) → same test: `FAILED ... panicked at crates/xberg/src/extractors/pdf/mod.rs:5973`.
  - On HEAD: same test passes (`1 passed; 0 failed`), the 4 new `apply_fabricated_provenance_pages_*` unit tests pass, and the full `xberg-native-pdf` suite is green (`288` `test result: ok` blocks, 0 `FAILED`).

**Blocking:**
- `PdfMetadata` (`crates/xberg/src/pdf/metadata.rs`) is an Alef-tracked source struct — its sibling field `scanned_pages` is generated into `packages/python/xberg/_xberg.pyi`, `packages/go/binding.go` + `include/xberg.h`, `packages/elixir/lib/xberg/pdf_metadata.ex`, and documented in `docs-site/src/content/docs/reference/types.md`. The new `fabricated_text_pages` field isn't reflected in any of them (`types.md`'s `Pdf` row still lists only `scanned_pages`, `layout_gated_pages`, ...). `.github/workflows/ci-lint.yaml`'s `alef-bindings-freshness` job (`task alef:verify`) runs on this PR (touches `crates/**`, targets `main`) and hashes source crates including this one — this will fail CI as-is. Needs `task alef:generate` + `task alef:verify`, committed atomically per the repo's own alef workflow.

**Non-blocking:**
- The prior two `fix(pdf)`/`fix(ppt)` commits on this branch each added a `CHANGELOG.md` `[Unreleased]` entry; this one doesn't. Worth adding for consistency, no CI gate found enforcing it.
- `extract_pdf_specific_metadata` (`crates/xberg/src/pdf/native/metadata.rs`) now trips poly's `function-too-long` warning (88 lines, max 80) — warning-severity only, doesn't fail CI.
- No end-to-end regression test exercises the "false-positive" arm (a real, correctly-mapped embedded subset font going through the actual extraction path and landing with an empty `fabricated_text_pages`) — the checklist's claim is backed only by the pure-function no-op unit test plus pre-existing, unrelated `best_mapping_provenance` unit tests. Low risk since `best_mapping_provenance` itself is untouched, but the new same-pass resolution code in `run.rs` isn't directly covered by a provenance-specific unit test in `xberg-native-pdf` either.

No AI attribution, no never-list values, comments carry `~keep` markers, `poly fmt --check` clean.

tobocop2:
### This does not fix the reported document, and the description above overstates it

Re-measured against the reported file at this commit, with OCR enabled so the default strategy branch this change touches is the one that runs:

```
extraction_method = native
ocr_used          = false
445.7 ms
```

Identical to `main`. A trace confirms this change's own code path executed, with the alphanumeric ratio matching the figure quoted above.

### Why

The signal this routes on fires only when no character mapping tier resolves a character at all. The reported file's font does resolve, through its own embedded mapping, to the wrong ordinary character. That is a present-but-wrong mapping, and it lands on a resolved tier, never on the fallback this change keys off.

The test added here builds a font with no mapping at all. That is a different shape: missing, rather than present and wrong. It passes, and it does not cover the reported case.

### What is still true

The underlying lookup repair in this change is real and worth keeping: the provenance value was empty for virtually every font because of a key mismatch, which is fixed here and proven.

### What needs to change

The description claims the default strategy now routes pages with a fabricated mapping to OCR. For the reported document it does not. Either the routing must also consider a mapping that resolves to implausible output, or this should be scoped as the lookup repair it is, with the reported case left open.

Filed so the claim does not merge as written.

tobocop2:
Added the changelog entry for consistency with the other fixes on this branch.

The bindings regeneration is the blocking one and is not done yet: it needs a build host that is down at the moment. I will run the generation and the verification and push them as one commit before asking for this again, rather than claim it here.

On the function length warning and the missing false-positive end-to-end test: both fair. The second is the one worth having, since the no-op unit test proves the pure function and not the path.

Goldziher (changes requested):
Two things before this can go in.

1. Regen is still owed. `alef verify --exit-code` fails on this head: `Crates whose generation inputs ... changed since their last successful alef generate: xberg`. `fabricated_text_pages` is absent from `packages/python/xberg/_xberg.pyi`, `packages/go/binding.go` + `include/xberg.h`, `packages/elixir/lib/xberg/pdf_metadata.ex` and `docs-site/.../reference/types.md` (0 hits, while the sibling `scanned_pages` hits all four). Push `task alef:generate` + `task alef:verify` as one commit, as you said.

2. "Closes #1667" and the new CHANGELOG `(GH#1667)` entry overclaim, and by more than the 08:53 comment says. `best_mapping_provenance` (`crates/xberg-native-pdf/src/fonts/font_dict.rs:542-579`) returns `Fallback` only for a Type0 font with no non-empty `/ToUnicode`, no non-Identity `CIDSystemInfo` ordering and no embedded cmap; step 4 is `if self.subtype != "Type0" { return P::EncodingName; }`, so a simple font can never produce `Fallback`. #1667's repro is a Type 1/TrueType font with a shifted custom encoding — that lands on `EncodingName`, `fabricated_char_counts` (`scan_detect.rs:187`) counts only `== Some(Fallback)`, and `apply_fabricated_provenance_pages` early-returns on the empty slice. The PR's fixture (`identity_h_no_tounicode_pdf`: Type0/Identity-H, no ToUnicode) is the one shape that does hit `Fallback`.

Change to "Refs #1667", leave the issue open, and reword the title, PR body and CHANGELOG entry to what this is: the alias-keyed provenance lookup repair plus routing of `MappingProvenance::Fallback` pages to OCR under `Auto`. The present-but-wrong mapping case needs a different detector (#1696).

Non-blocking: the fix commit subject is 82 chars; `extract_pdf_specific_metadata` (`crates/xberg/src/pdf/native/metadata.rs:47`) now trips poly `function-too-long` (88/80) where main is clean; the new two-line comment at `ocr/scoring.rs:1120` lacks `~keep` unlike every other comment block in the diff; the false-positive arm still has no end-to-end test through `run.rs`.

tobocop2:
On the bindings regeneration: I ran it, and the result argues against doing it in this pull request. Numbers below, measured on one L4 box with alef 0.92.1, the version `alef.toml` pins, and the sibling checkouts placed where `alef.toml` expects them.

### The freshness gate is already red on the base

`alef verify --exit-code` returns 1 on `main` as well as on this branch. I measured `main` at `470af61733` and this branch at `13cc53f2ff`. Both produce 234 lines of output once the per-type and per-function informational lines are filtered, and the two are identical apart from the order the languages are generated in, which varies between runs. Both end with the same error, and both list the same single crate under "Crates whose generation inputs changed".

So the gate does not distinguish this branch from its base. It is not reporting a staleness this change introduced.

### Regenerating sweeps in far more than this field

`alef all --clean --skip-snippet-validation` on this branch gives 573 files changed, 72,447 insertions and 133,467 deletions across every language package. The new field does land in the generated output, in the documentation reference pages and the intermediate representation. It arrives buried in a net removal of about 61,000 lines that has nothing to do with a font provenance lookup.

Committing that here would make a three-file fix unreviewable and would attribute a repository-wide regeneration to it.

### What I would do instead

Land this fix on its own, and treat the regeneration as its own change, where the diff is the point and can be reviewed as one. I am happy to open that separately.

If you would rather this branch carry it, say so and I will add it, but I did not want to push 573 files into a pull request you have already reviewed without asking first.

The changelog entry you asked for is added. On the other two points: the function length warning is real and non-blocking, and the missing end-to-end test for the false-positive arm is the one worth having, since the existing no-op unit test proves the pure function rather than the path.

tobocop2:
Correcting my earlier comment. My 573-file figure was an artifact of my own machine, not a property of this branch, and the argument I built on it does not stand.

### What was actually wrong

The box I regenerated on had poly 0.21.12, which cannot load this repo's `poly.toml` — it rejects the git hook stage `snapshot_include`. alef therefore wrote every generated file unformatted, and the diff was the formatting, not the content.

Installing poly 0.28.0, the version `Goldziher/poly@v0` resolves to in CI, plus Go 1.27.1 for gofmt, gives the real change. Measured just now on #1682, which needed the same regeneration for the same reason: **28 files, 808 insertions, 146 deletions**, and `task alef:verify` exits 0 on the committed tree. That is your expected set: the api reference pages, `types.md`, `configuration.md`, the node, php and py crate sources, `index.d.ts`, and the csharp, dart, go, kotlin-android and zig packages.

With the stale poly it was 572 files and about 66k insertions against 126k deletions, on a tree that had already merged main. So the file count was driven by the formatter version, not by drift on the base. I attributed it to the base and I was wrong.

### What that means here

Your instruction stands unchanged and I withdraw my objection: merge main, regenerate on top, commit as one. There is no 573-file sweep to avoid, and no reason to split it into a separate change.

That work is queued for this branch now. I will report the file count when it lands so the number is on the record rather than asserted.

One thing worth passing on for anyone else regenerating: a box with the wrong poly produces a diff that looks like catastrophic drift and is entirely whitespace. `task alef:verify` exiting 0 is the check that distinguishes them.

tobocop2:
You landed all three of these yourself before I got to them, so I have added no commits. I ran an independent gate on `e5b27bae07` instead, since the value left was checking your work rather than repeating it.

### All three requests are satisfied

**Merge.** `origin/main` is an ancestor of the tip and `tip..origin/main` is empty. The changelog kept both sides: Unreleased Fixed carries 60 bullets on main and 68 at the tip, and the tip's reference set is a strict superset of main's, adding GH#1662, 1665, 1666, 1667, 1671, 1672 and 1673. Nothing dropped.

**Regeneration.** `6dd9c90404` is 45 files, 431 insertions, 64 deletions, with `.alef-generation.toml` in the same commit. `fabricated_text_pages` now appears in all four places it was missing — the python stub, the go binding, the go header and the elixir module — plus the docs-site types reference.

**Changelog scope.** `7a51f96a9e` now leads with the alias-keyed lookup repair, scopes routing to the fallback tier under the default strategy, and ends with the caveat that a font whose mapping resolves to the wrong characters is a different failure mode this signal does not cover. That matches what both of us measured independently, and it no longer reads as fixing the reported document.

### Gate, run here

With alef 0.93.1 (the pin), poly 0.28.0 and Go 1.27.1, corpus fetched, clean worktree:

- `task alef:verify` rc=0, "All bindings and versions are up to date", 3793 managed paths read and 3362 content-verified against current generation inputs. That is the arm proving the regeneration is complete rather than partial.
- `cargo test -p xberg --features pdf,ocr --lib` rc=0, 3935 passed, 0 failed, 0 filtered.
- The same command filtered on `provenance` gives 16 passed, 0 failed, confirming the tests for this fix actually compile into that binary rather than sitting behind a feature nobody builds.
- `cargo fmt --all --check` rc=0.

Nothing outstanding from my side. This looks ready whenever you want to clear your own review.

One note on the poly version, since it burned two runs here: a box whose poly cannot load `poly.toml` still lets alef succeed and writes every generated file unformatted, giving a 572-file diff that is entirely whitespace. `task alef:verify` exiting 0 is what distinguishes that from a real regeneration.

PR #1679, opened 2026-09-19
fix(ocr): let candle backends ignore the PDF page hints in backend options

### LINKED ISSUE #1672: bug: PDF OCR: the page path injects source_dpi into backend_options and every candle backend rejects it, so all rendered pages come back empty
### Description

Every candle OCR backend returns an empty result for every rendered PDF page. The PDF OCR path inserts `source_dpi` (and `page_rotation_degrees` for rotated pages) into the shared `backend_options` object in `extractors/pdf/ocr/pipeline.rs` (`ocr_config_with_page_rotation_hint`), with the comment that backends which do not recognise the keys ignore them. The candle backends deserialize that object into `#[serde(deny_unknown_fields)]` structs (`candle_ocr/config.rs`, `parse_backend_options`), so each page fails with:

```
OCR of page N failed and could not be recovered: Validation error: invalid candle-<backend> backend_options: unknown field source_dpi, expected one of ...
```

Only the pages that go through the embedded-image retry produce text, because that route passes the options through unchanged. On a 22-page PDF, 19 of 22 pages were empty under `candle-paddleocr-vl` and `candle-deepseek-ocr`, 21 of 22 under `candle-glm-ocr`. Rotated pages fail the same way through `page_rotation_degrees`.

Expected: the two contracts agree. Either the hint keys are stripped before a candle backend parses its options, or the candle option structs accept and ignore them.

### Steps to reproduce

1. Build with `full` and `candle-cuda` (or CPU) on a machine with the PaddleOCR-VL weights available.
2. Extract any multi-page PDF with `{"force_ocr": true, "ocr": {"backend": "candle-paddleocr-vl", "language": "eng"}}`.
3. Every page carries the validation warning above and `pages[].content` is empty except for pages recovered through the image retry.

xberg v1.2.4, release build with `candle-cuda`, NVIDIA L4, driver 610, CUDA 12.8, 2026-09-18.

tobocop2:
## Related

Closes #1672

## Description

The PDF page route adds rendering hints to the shared backend options object for every OCR backend. The option contract says unknown keys are ignored, and the tesseract and native paddle backends honour it by reading known keys off the raw value. The four candle backends parse their options strictly and reject unknown fields, so every rendered PDF page failed validation and came back empty.

## Behaviour

The two pipeline hint keys are stripped before the candle backends parse their options. Strict validation stays in place for every other key, so a misspelled candle option still fails with a named error.

## Checklist

- [x] A genuine unknown key and a near miss typo both still error
- [x] fmt and clippy clean


Goldziher (approved):
Verified. `parse_backend_options` (crates/xberg/src/candle_ocr/config.rs) is the single chokepoint all four candle backends call (trocr_backend.rs:199, paddleocr_vl_backend.rs:184, glm_ocr_backend.rs:410, deepseek_ocr_backend.rs:142), so the strip covers all four, not just the three named in the description. The strip removes exactly `SOURCE_DPI_BACKEND_OPTION`/`PAGE_ROTATION_DEGREES_BACKEND_OPTION` (named constants shared with the injection site in pipeline.rs, not string literals), and happens before `serde_json::from_value` with `deny_unknown_fields` still in effect — confirmed by planting a revert of the strip: `should_ignore_pdf_pipeline_hint_keys_in_candle_backend_options` fails with `unknown field \`source_dpi\`` on candle-trocr, restored and re-ran green. The pre-existing `should_reject_invalid_candle_backend_options_with_context` test (unchanged by this PR) already proves a real typo (`versoin`) still errors post-strip. `cargo test -p xberg --lib --features candle-vlm-ocr` and `cargo clippy` are both clean.

Non-blocking: the strip is unconditional inside `parse_backend_options`, not scoped to the PDF-pipeline call path, so a caller who directly passes `source_dpi`/`page_rotation_degrees` in `backend_options` outside the PDF route (e.g. via the public API) has those keys silently dropped rather than validated, same as the pipeline hints. Injecting pipeline-internal state into the same JSON bag users configure is what created this bug class; worth a follow-up to keep hints out of the user-facing options object, but not a blocker.

PR #1680, opened 2026-09-19
fix(pdf): keep native tables on pages OCR did not replace

### LINKED ISSUE #1670: bug: PDF: the mixed OCR path returns ten times fewer tables than the native path on the same file
### Description

On a long PDF with mixed native and scanned pages, the default path (`extraction_method = "mixed"`, per-page OCR on the scanned pages) returns 14 tables, the native-only path (`disable_ocr = true`) returns 140 tables, and `force_ocr = true` returns 202 tables, all on the same file. The mixed path loses ninety percent of the tables the native path finds on pages that were never OCR'd. On a second mixed PDF the mixed and native counts agree (490 and 490), so the loss depends on the document.

Expected: the mixed path keeps every native table on a page that kept its native text, and adds the OCR tables of the pages that went to OCR.

### Steps to reproduce

1. Extract a long mixed PDF three times: default config, `disable_ocr = true`, `force_ocr = true`.
2. Compare `counts.tables` across the three results.

xberg v1.2.4 through the Pro HTTP API, release build, 2026-09-18. Config for the default run: `{"pages":{"extract_pages":true},"include_document_structure":true}` with the default tesseract OCR config filled in.

tobocop2:
## Related

Closes #1670

## Description

On a long document that mixes native and scanned pages, the default path returned far fewer tables than the native-only path on the same file: 14 against 140 on the reported document.

The routine that folds OCR results back in replaces the whole document's table list with the tables OCR found. That is correct when every page went through OCR, which is the path it was written for. The mixed path reuses it, so a handful of scanned pages producing one table discards every native table on every page OCR never touched. The loss is global rather than per page.

## Behaviour

The mixed path now replaces the tables only of pages OCR itself produced a table for. A page that went through OCR and produced no table keeps its native tables. The full OCR path is unchanged and its existing tests pass unmodified.

Measured on the reported document: the two paths now agree at 135 tables each. A control document that never showed the defect is unchanged at 715 on both paths.

## Checklist

- [x] Full OCR path proven unchanged
- [x] Ordering and duplication checked in the merged list
- [x] fmt clean


Goldziher (approved):
Verified the fix and the fold-back behaviour it changes.

- Confirmed the bug exists on main: the mixed path (mod.rs:2441 on main) called `replace_tables_with_ocr_output`, which does `*tables = ocr_tables` — a whole-list replace. On head it's replaced with `merge_mixed_ocr_tables`, which retains native tables for any page not in the OCR-produced-tables set and replaces only per-page. Correct fix for GH#1670.
- Page-index alignment checked: native `Table.page_number` is `(index + 1)` (1-based) everywhere it's constructed, and `structured_pages`/`ocr_page_numbers` keys are also 1-based (`page_idx + 1` in ocr/pipeline.rs, and `extract_mixed_ocr_native`'s own doc comment states 1-indexed). No off-by-one between the two sides being joined.
- Scanned page where OCR found no table but native had a (garbage) table: covered by `a_page_ocr_ran_on_but_found_no_table_for_keeps_its_native_table` — the native table is kept, since `ocr_pages` is built from `ocr_tables` (pages that actually produced a table), not from `structured_pages`' keys (pages OCR ran on). Matches the issue's stated expectation ("keeps every native table on a page that kept its native text... adds the OCR tables of the pages that went to OCR").
- Full-OCR path: unchanged, still `replace_tables_with_ocr_output` at the force_ocr/OCR-fallback call site (mod.rs:2316); its own tests (`ocr_tables_replace_native_tables_and_are_sorted`, `empty_ocr_tables_preserve_native_tables`) are untouched by the diff and still pass. Verified none of the three mixed-path branches (`force_ocr_pages`, automatic `scanned_pages_to_ocr`, `RunFallbackOnPages`) ever populate the `ocr_tables` var that feeds that call, so it's a no-op for mixed extractions — confirmed by reading the branches, not just the diff.
- Planted-failure proof: reverted `merge_mixed_ocr_tables` to the old whole-list-replace body and ran
  `cargo test -p xberg --features pdf,ocr-pipeline --lib -- mixed_ocr_tables_replace_only_the_pages_ocr_produced_a_table_for a_page_ocr_ran_on_but_found_no_table_for_keeps_its_native_table`
  → RED: `assertion \`left == right\` failed: pages 1 and 3 were never sent to OCR; their native tables must survive / left: 1 / right: 3`. Restored via `git show HEAD:<path> > <path>`; both tests green again afterward, `git status` clean.
- Targeted tests: both new tests pass individually and together; `extractors::pdf::tests::` module run is 48 passed / 2 failed, the 2 failures (`test_form_field_extraction_{enabled,disabled}`) are pre-existing — `test_documents/vendored/pdfium-render/form-test.pdf` isn't present in this worktree at all, unrelated to this diff.
- `cargo clippy -p xberg --features pdf,ocr-pipeline --lib -- -D warnings` and `cargo fmt -p xberg -- --check` both clean.
- Conventions: commit type `fix`, header 70 chars; no AI attribution; no never-list values; CHANGELOG.md edited directly (this repo has no `changelog.d/`, so the content-hash-filename rule doesn't apply); the new `///` doc comments are rustdoc (poly.toml has `remove_docs = false`), not subject to the `~keep` uncomment lint.
- Not independently verified: the PR body's "135 tables on both paths" / "715 on the control document" measurements — no access to the reported document here.

No defects found. Approving.

tobocop2:
### Merge order: this must land together with the other

Measured on the 731 page partner document across seven trees.

Taken alone, this change flips the default per-page gate into the mixed native and scanned path for that document. On that path the table count collapses from 135 to 14 and wall time rises from about 30 seconds to about 123 seconds.

The table loss is fixed by the mixed-path table fix. Taken alone, that fix shows no effect, because on every other tree the default gate never entered the mixed path for this corpus at all.

So each change looks harmless in isolation and the pair is correct, but this one landing first is a customer facing regression on exactly the document class this work came from.

Merge both, or neither.

Issue #1677, opened 2026-09-19
bug: OCR aggregate confidence is never populated on the page-level OCR route


tobocop2:
`ExtractionConfidence.ocr_aggregate` is never populated when a PDF goes through the page-level OCR route. The only call site that populates it is wired to the embedded-image OCR route.

So a document OCRed page by page reports no aggregate OCR confidence, although every page carries its own confidence value. Any consumer that reads the aggregate sees nothing and cannot tell "OCR was not used" from "OCR ran and the aggregate was never filled in".

Populate the aggregate on the page-level route as well, from the per-page confidence values.

Found while fixing the quality score, which had to read per-page confidence directly because the aggregate was empty.

PR #1682, opened 2026-09-19
fix(ocr): let low OCR confidence lower the quality score

### LINKED ISSUE #1669: bug: quality_score stays 1.0 on OCR runs with low aggregate confidence and hundreds of noise warnings
### Description

`quality_score` is 1.0 on every run, including OCR runs whose `extraction_confidence.ocr_aggregate` is 0.81 to 0.93 and which carry 44 to 175 OCR noise warnings ("produced suspected OCR recognition noise", "removed N OCR line(s)"). The score does not move with OCR confidence or with the number of pages the engine itself flags as noisy, so a caller cannot use it to decide whether the text needs review.

Expected: `quality_score` falls when the OCR aggregate confidence falls and when the engine drops or flags lines. The value 1.0 must be reserved for output the engine has no reason to doubt.

### Steps to reproduce

1. Extract a long PDF with mixed native and scanned pages with `force_ocr = true` and the tesseract backend.
2. Read `quality_score` (1.0), `extraction_confidence.ocr_aggregate` (about 0.8) and count `processing_warnings` with source `ocr` or `tesseract` (more than 100).

xberg v1.2.4 through the Pro HTTP API, release build, 2026-09-18. Nine runs over three PDFs, every one with `quality_score = 1.0`, three of them with `ocr_aggregate` 0.88, 0.81 and 0.88.

### LINKED ISSUE #1694: bug: the quality score and the OCR aggregate disagree on the confidence floor
### Deployment

Other

### Version

Branches `fix/1669-quality-score-ocr-confidence` and `fix/1677-ocr-aggregate-page-route`, both open.

### Description

Two changes now compute the same quantity from the same source, and they disagree about when it is trustworthy.

The quality score fix reads per-page OCR confidence directly and requires at least 20 recognised words before it will use the value. It reads it directly because the aggregate field was empty at the time.

The aggregate fix now populates that field, using the same word-count weighted mean, and requires only more than zero words.

Once both are merged the computation exists twice, and the floors differ. A document with fewer than 20 recognised words reports a real aggregate confidence while its quality score stays uncapped, so the two numbers describe the same OCR run and disagree about whether it was good.

Nothing is double counted; they feed separate output fields.

### Steps to reproduce

1. Merge both branches.
2. Extract a scanned document short enough to yield fewer than 20 recognised words.
3. Compare the aggregate confidence against the quality score.

### Logs and configuration

Either resolve it by having the quality score consume the now-populated aggregate instead of recomputing it, or by aligning the two floors and stating why two computations remain.

This could not be fixed from either branch alone without editing the other, which is why it is filed rather than folded into one of them.

tobocop2:
## Related

Closes #1669
Closes #1694

## Description

`quality_score` returns 1.0 on every run, including OCR runs that carry low confidence. On the reported document the score was 1.0 while aggregate OCR confidence was 0.81 and the run produced 175 noise warnings.

The score reads text artefact density only. It never reads the confidence the system already holds, so no amount of poor recognition can move it.

## Behaviour

The score is now capped at the word-count weighted mean of per-page OCR confidence. A floor of 20 recognised words keeps one stray fragment from sinking a long, confident document; that floor is independently calibrated for this cap, not derived from any existing "not enough evidence" threshold elsewhere in the OCR subsystem.

A document with no OCR is unaffected, verified against a real native PDF at base and head.

The cap reads OCR recognition confidence only (per-page, or per-element on the embedded-image route). It does not read coverage or warning presence, so the existing decision to keep completeness signals out of this score still holds and its test passes unchanged.

#1677 landed on `main` while this was in review and populated `ExtractionConfidence::ocr_aggregate` from the same per-page and per-element data, using its own "any recognized word" floor. That left two computations of the same quantity with two different floors, filed as #1694: below 20 words the aggregate reported a real number while the score stayed uncapped. This PR now shares the fold instead of recomputing it: `ConfidenceSignals::ocr_confidence_from_elements` / `ocr_confidence_from_pages` (`crates/xberg/src/heuristics/confidence.rs`) are the one place either statistic is computed, and this cap layers its own, stricter evidence floor on top of that shared fold's `total_words` before trusting the mean as a ceiling. The cap also now follows `ocr_aggregate`'s source preference (`ocr_elements` before `pages`), so the embedded-image OCR route is capped too, not just the page-level route.

## Checklist

- [x] Failing test first, red on base and head
- [x] Native documents proven unaffected, not inferred
- [x] fmt clean


Goldziher (changes requested):
Verified: fix is absent on `main` (base has no `aggregate_ocr_confidence`/`ocr_confidence` reference in `quality_processor.rs`). `cargo test -p xberg --lib --features quality quality_processor` -> 12 passed, 0 failed. Planted-failure check (removed the `quality_score = quality_score.min(ocr_confidence)` hunk, kept tests): `quality_score_is_capped_by_low_ocr_confidence_on_shape_clean_text` goes RED (`left: Some(1.0), right: Some(0.81)`), 11 passed/1 failed; restored, back to 12/0. Confidence scale is 0.0..=1.0 on the producer side (`scoring.rs::page_ocr_confidence`, clamped), matching the cap's `.min()` against `quality_score`'s own 0.0..=1.0 scale — no unit mismatch. Native/non-OCR path (pages=None) is unaffected and covered by three existing tests. Floor, zero-word pages, and no-calibrated-scale (`score: None`) are each covered by a dedicated test and behave correctly.

Two things to fix before merge:

1. `crates/xberg/src/text/quality_processor.rs:127` — `MIN_OCR_WORDS_FOR_CONFIDENCE_FLOOR: u64 = 20` is a brand-new, independent constant, not a reference to any existing OCR-subsystem constant. The commit message claims "that floor matches the value the OCR subsystem already uses for the same 'not enough evidence' judgement," but the only existing `20` is `default_min_words_for_ocr_output_check()` in `crates/xberg/src/core/config/ocr.rs:378-380` — a private serde-default for the *user-configurable* `OcrConfig::min_words_for_ocr_output_check` field, which gates an unrelated check (`max_ocr_output_fragmented_word_ratio`). It's a duplicated literal that happens to match today, not a shared constant: if a deployer changes `min_words_for_ocr_output_check`, this cap's floor won't follow, silently breaking the parity the PR claims. Either import/derive from that existing default, or drop the "matches the OCR subsystem" claim from the doc comment and commit message and own this as an independent, newly-calibrated floor.

2. `crates/xberg/src/types/extraction.rs:247-253` — the public rustdoc on `ExtractedDocument::quality_score` (which is what's mirrored verbatim into `docs-site/src/content/docs/reference/{types,configuration}.md` and every language binding's generated docs) still only says "Text cleanliness/readability score... not a completeness or recall score." It says nothing about the new OCR-confidence cap. Every caller across 15+ bindings reads this field's semantics from that comment, not from `QualityProcessor`'s internal doc, so the behavior change isn't discoverable from the public API surface. Update this doc comment (a two-line addition is enough); no docs-site/binding files need hand-editing since they're generated from it.

Everything else checked clean: commit type/format, no AI attribution, no never-list values, `poly lint`/`poly fmt --check` clean on the changed file, no other call site of `calculate_quality_score` needs the same treatment (the other one, `extractors/image.rs:135`, is an internal OCR-candidate-selection heuristic, not the public `quality_score` field).

tobocop2:
Both findings are fixed.

On the floor: you are right that it is an independent constant and not a reference to anything shared, and the commit message overstated it. Rather than derive it from the operator-tunable check, which would let a deployer move the ceiling on every document's quality score by tuning something unrelated, the constant now documents itself as deliberately independent and says why. The claim it cannot support is gone.

On the public documentation: added. The field's own rustdoc now states that an OCR result with enough recognised words is additionally capped by the mean recognition confidence, and that a native extraction is not. That is the surface every binding reads from.

I did not hand-edit the generated mirrors, on your own finding from #1695 that the freshness gate does not hash them.

Goldziher (changes requested):
Both prior findings are resolved in `f3c1fea053` (the constant now documents itself as deliberately independent; the public `quality_score` rustdoc carries the cap). Two new things block merge.

1. `crates/xberg/src/types/extraction.rs` is a listed alef source (`alef.toml`, xberg `sources = [...]`), and the freshness gate hashes the raw bytes of every listed file, docstrings included. The #1695 finding was about `quality_processor.rs`, which is not listed. `alef verify --exit-code` fails on this head with `Crates whose generation inputs ... changed: xberg`, and the same change shape failed the CI gate on #1678 (`pdf/metadata.rs`). Run `task alef:generate` and commit the regenerated mirrors plus `.alef-generation.toml` in the same commit: 28 generated mirrors (`docs-site/.../reference/api-*.md`, `types.md`, `configuration.md`, `crates/xberg-{node,php,py}/src/lib.rs`, `xberg-node/index.d.ts`, `packages/{csharp,dart,go,kotlin-android,zig}/...`) still carry the old `quality_score` text and none carry the new paragraph. Note that main is being regenerated as part of the 1.2.6 version sync, so merge main first and regenerate on top of it.
2. `CHANGELOG.md` conflicts with main after #1695 and #1691 merged (`git merge-tree` reports only that file). Merge main.

Non-blocking:
- The PR body still says the floor "matches the value the OCR subsystem already uses"; edit it to match the code comment.
- #1695 is on main now. Its `ocr_aggregate` prefers `ocr_elements` before pages, so a document OCR'd via the embedded-image route gets a populated aggregate but an uncapped `quality_score`; and its "this is the one place either can compute mean confidence per recognized word" comment becomes false once this merges. The clean shape is to expose `ExtractionConfidence`'s page fold as `pub(crate)` returning `(mean, total_words)` and apply the 20-word floor here (this is also #1694).
- `docs-site/src/content/docs/concepts/extraction-pipeline.md:188` is hand-written (no alef marker) and describes quality scoring without the cap; one sentence.
- In the changelog, "capped by the aggregate OCR confidence" reads as `extraction_confidence.ocr_aggregate`, which is a different number below 20 words and on the element route. Say "the word-count-weighted mean of per-page OCR confidence, once at least 20 words were recognized".
- The `// (0.9*100 + 0.5*300) / 400 = 0.6` test comment has no `~keep` and will be pruned; harmless.

tobocop2:
Before this merges: our regeneration is incomplete in a way that `task alef:verify` cannot see, and I would rather raise it than push it.

### What we committed, and what it misses

The commit carries 28 regenerated files, 808 insertions and 146 deletions, and `task alef:verify` exits 0. Regenerating again at that head with the same toolchain reproduces all 28 byte-identically, so those are genuine.

The problem is the ones we did not commit. A regeneration rewrites 485 files here. Most of that is formatter noise from tools this box lacks, and we restored those deliberately. But **37 of them differ by token multiset**, after normalising whitespace, wrapping and import order away. Those are real content differences, not layout.

Four source-derived changes are missing from the restored mirrors:

- `wants_own_bytes_in_result`, 14 files, including `packages/python/xberg/_xberg.pyi` and `packages/ruby/sig/types.rbs`
- `runs_ocr_on_embedded_images`, 9 files
- the reworded `resolve_tessdata_path` doc, 13 files
- the reworded `ocr_aggregate` doc, 3 files

The probe has power rather than being a false alarm: the committed `crates/xberg-py/src/lib.rs` already carries the first two, and a Java mirror in the same sweep came back token-identical while a C# one differed.

### Why verify does not catch it

Those 37 were already stale on main. This commit advances the `xberg` generation hash in `.alef-generation.toml` from `8c08241bfc` to `67c2fb8388`, which certifies them as current. Verify then has nothing to complain about. Its own coverage block reports 101 paths passing without content verification.

So a partial regeneration plus a hash bump converts pre-existing staleness into a signed statement that the tree is fresh. That seems worse than leaving the gate red.

### Why we cannot simply finish it

`task alef:generate` exits 201 on this machine: ten formatters are absent, ruff among them, and the e2e codegen step fails. We can produce the 28 files correctly and we cannot produce a complete, correct regeneration.

### How would you like this handled?

Three options as I see them, and the choice is yours rather than mine:

1. We install the full formatter set and regenerate completely, committing all 37 alongside. Largest diff, but the hash would then be honest.
2. We drop the generation-hash bump from this commit, keep only the source and doc changes, and let the regeneration happen separately where the toolchain is complete.
3. You regenerate, as you did on #1678, and we rebase onto it.

Worth checking the same thing on #1678: that regeneration was also produced outside CI, so if the hash moved there while any mirror stayed stale, it has the same shape. I have not measured that one, and I am not asserting it.

tobocop2:
Handing this over to you, and unassigning myself.

The regeneration needs an environment this side cannot provide: `task alef:generate` exits 201 here with ten formatters absent, ruff among them, so any regeneration we produce is partial. Committing a partial one while advancing the generation hash is what turns pre-existing staleness into a signed claim of freshness, which is the finding above.

The two code findings from your earlier review are already resolved on the branch head `f3c1fea053`: the floor documents itself as deliberately independent of the operator-tunable check, and the public `quality_score` rustdoc carries the cap.

If it is useful, the 28 generated files we produced are preserved on `wip/1682-merge-regen-do-not-merge` at `7af22ce078`, together with the main merge that keeps all three changelog bullets and the two wording fixes you asked for. Regenerating at that head with the same toolchain reproduces all 28 byte-identically, so they are sound on their own. That branch must not be merged as it stands, for the reason above.

The 37 mirrors that stay stale are: `wants_own_bytes_in_result` across 14 files including the python stub and the ruby signatures, `runs_ocr_on_embedded_images` across 9, the reworded `resolve_tessdata_path` doc across 13, and the reworded `ocr_aggregate` doc across 3.

Issue #1681, opened 2026-09-19
bug: an oversized page falls back to native text with no warning


tobocop2:
A single page too large to render within the content limit still falls back to native text on the automatic OCR route, with no warning. The page returns whatever its text layer holds, which on a scanned page is nothing.

This is separate from the batch sizing defect. Correcting how a batch is sized against the content limit stops a wide thread budget from skipping every page, but one oversized page on its own still degrades quietly.

The caller sees a completed extraction. Nothing in the result says a page was skipped, so a document that is missing a page of text looks identical to one that never had text on that page.

Report the skip, and say which page and why.

## A sibling route has the same shape

The whole document OCR route sizes its batch from host memory alone and does not consider the content limit at all. The mixed route now does. Decide whether both routes should share one sizing rule, or say why they differ.

Found while fixing the batch sizing defect on the mixed route.

tobocop2:
### Status

Does not reproduce on current code. Closing.

### What was tested

A real reproduction was built: an oversized page hitting the content limit on the automatic OCR route, which is the path this issue names.

### What the code actually does

It emits a processing warning that names the specific page and the specific size and dimension reason, and reports that OCR was not used. The caller is told which page was skipped and why.

The silent fallback this issue describes is not present.

### Origin of the report

This was filed from a side observation made while fixing #1665, without a reproduction. The observation did not hold when one was built.

### Related, and deliberately unchanged

The whole-document OCR route sizes its batch from host memory and does not consider the content limit, unlike the mixed route fixed in #1665. That difference is intentional for now rather than an oversight, because the two routes have different failure semantics. Anyone wanting them unified should open that as its own issue.

PR #1684, opened 2026-09-19
fix(pdf): bound the render batch by the content limit, not the thread budget

### LINKED ISSUE #1665: bug: PDF OCR: every scanned page is skipped at a large thread budget because the render batch peak scales with thread count
### Description

With `concurrency.max_threads = 32`, a mixed PDF (160 native pages, 80 scanned pages at 200 dpi, A4) comes back with `extraction_method = "native"` and 80 empty pages. Every scanned page is skipped. The job succeeds and carries one warning. The same file at 4, 8 or 16 threads returns `mixed` with all 80 pages OCR'd in about 12 s.

Cause: the OCR pipeline sets its page batch size to the resolved thread budget, then validates the peak PNG-encode bytes of the whole batch against `security_limits.max_content_size` (`validate_png_encode_batch_peak` in `extractors/pdf/ocr/pipeline.rs`). The batch peak grows linearly with the thread budget: 32 ordinary A4 pages at 150 dpi need about 775 MB, above the 512 MiB default. Sixteen pages need about 387 MB and pass. On Linux the budget also follows the cgroup CPU quota, so a container with a CPU limit above about 22 cores hits this with no explicit setting. The failure degrades to native text with a warning, so the result reads as a success with most of its content missing.

Expected: the batch size for rendering and encoding is bounded by memory, not only by threads, so a higher thread budget never turns OCR off. `adapt_batch_size_to_memory` already exists on the other route and could apply here. When every candidate page is skipped, the result must not read as a success.

### Steps to reproduce

1. Build a 240-page PDF where every third page is a full-page 200 dpi JPEG raster with no text layer (reportlab, A4).
2. Extract with the default OCR config (tesseract, eng), pages on, `concurrency.max_threads = 32`.
3. Compare with `max_threads = 16`: `mixed`, 546,747 chars, no empty pages.

xberg v1.2.4, release build, 2026-09-18, two runs per setting.

```
ocr: Targeted OCR fallback failed (Validation error: Image dimensions 1241x1754 require 774591936 live image-processing bytes, exceeding or invalid under security_limits.max_content_size (536870912 bytes)) for pages [3, 6, 9, ...]
```

Result at 32 threads: `native`, 398,036 chars, 80 empty pages.

tobocop2:
## Related

Closes #1665

## Description

At a large thread budget, OCR was skipped on every scanned page. Batch size came straight from the thread budget, and the rendered batch was checked against the content limit only after rendering. A wider budget therefore asked for a wider batch with no notion of that limit, so the peak scaled with thread count until it tripped the limit and every page was skipped.

The failure was silent: pages were skipped rather than the run failing, so a caller saw a completed extraction with no OCR and no error.

## Behaviour

Batch size is now capped against the content limit before rendering, on the mixed native and scanned route. A wide thread budget no longer trips the limit there.

Narrow budgets are unchanged, and the batch never collapses to zero.

The whole document OCR route is untouched. It sizes its batch from host memory and does not consider the content limit; that difference is left as it was rather than changed as a side effect.

## Checklist

- [x] Narrow budgets and the zero-batch boundary both checked
- [x] fmt and clippy clean

A single oversized page still falls back to native text without a warning even with batch sizing corrected. That is tracked separately in #1681.


Goldziher (approved):
Verified: base (main@1.2.5) has no `adapt_batch_size_to_content_limit`; the batch-size derivation is genuinely fixed only on head. Fix is scoped correctly: `configured_batch_size` (thread-budget) is capped by `adapt_batch_size_to_content_limit` before rendering in `extract_mixed_ocr_native` (pipeline.rs:263-273), using a representative page's estimated PNG-encode peak divided into `max_content_size`; `validate_png_encode_batch_peak` still checks the real peak afterward, so a wrong estimate only shifts the boundary, never silently passes an oversized batch.

Tests: `cargo test -p xberg --features pdf,ocr --lib` targeted run — 5/5 new tests pass (`adapt_batch_size_to_content_limit_{shrinks_a_large_thread_budget_for_a4_pages,leaves_a_small_batch_untouched,never_shrinks_below_one_page}`, `estimate_png_encode_page_peak_bytes_matches_hand_computed_accounting`, `large_thread_budget_does_not_turn_off_ocr_at_a_fixed_content_limit`).

Planted-failure proof: reverted pipeline.rs:263-273 to `configured_batch_size` (pre-fix thread-budget-only sizing), reran the integration test — RED: `panicked ... Validation { message: "Image dimensions 1275x1650 require 93613576 live image-processing bytes, exceeding ... max_content_size (52428800 bytes)" } ... test result: FAILED. 0 passed; 1 failed`. Restored via `git show HEAD:<path>`, reran — 5/5 green again.

`poly fmt --check` and `poly lint` clean on all 3 changed files (uncomment pass included — the one new rationale comment carries `~keep`). `cargo fmt -p xberg --check` clean. No public API/bindings touched (both new fns are `pub(super)`). No never-list values, no AI attribution, commit type `fix` is valid.

Two non-blocking observations:
1. `pipeline.rs:2542-2544` — `adapt_batch_size_to_content_limit` fails *open* on the `estimate_png_encode_page_peak_bytes` overflow branch: `configured` batch size is returned unmodified, i.e. exactly the pre-fix uncapped sizing this PR fixes. Bounded by real `MediaBox`/DPI limits it looks unreachable today, but it's inconsistent with the adjacent `per_page_bytes == 0` case, which is likewise unreachable but was still handled defensively. Consider failing closed (e.g. `1`) instead.
2. The commit message discloses that `extract_with_ocr` (whole-document OCR route) has the same theoretical gap against `max_content_size` and says it's "reported instead" of fixed here — but no tracking issue exists for it (unlike the disclosed oversized-single-page gap, which is #1681). Worth filing before/at merge so the claim is checkable.

Neither blocks this fix. Approving.

tobocop2:
### Merge order: this must land together with the other

Measured on the 731 page partner document across seven trees.

Taken alone, this change flips the default per-page gate into the mixed native and scanned path for that document. On that path the table count collapses from 135 to 14 and wall time rises from about 30 seconds to about 123 seconds.

The table loss is fixed by the mixed-path table fix. Taken alone, that fix shows no effect, because on every other tree the default gate never entered the mixed path for this corpus at all.

So each change looks harmless in isolation and the pair is correct, but this one landing first is a customer facing regression on exactly the document class this work came from.

Merge both, or neither.

PR #1685, opened 2026-09-19
fix(pdf): run the structured build when document structure is requested

### LINKED ISSUE #1668: bug: PDF: the document structure of long mixed native-and-scanned PDFs holds only paragraph nodes, no headings, tables or page breaks
### Description

`include_document_structure = true` on three long PDFs with mixed native and scanned pages returns a document tree that holds nothing but `paragraph` nodes:

| PDF | tables in `tables` | structure nodes | node types |
|---|---|---|---|
| A, long, mixed | 14 | 8,146 | paragraph only |
| B, long, mixed | 490 | 34,683 | paragraph only |
| C, short, native | 60 | 1,054 | paragraph only |

The same build returns `heading` nodes with levels, `group`, `table`, `page_break` and bounding boxes on a synthetic reportlab PDF with 24 pt and 16 pt Helvetica headings. On the real files there is no `heading`, no `table` (although hundreds of tables were extracted into `tables`), no `page_break`, and no `list`.

Expected: the tree carries the tables the extractor already found, page breaks between pages, and headings where the font metrics or the tag tree support them. A tree of paragraphs alone does not let a caller audit structure.

### Steps to reproduce

1. Extract a long PDF with mixed native and scanned pages with `{"pages":{"extract_pages":true},"include_document_structure":true}` and the default OCR config.
2. Count `document.nodes` by `content.node_type`.
3. Compare `counts.tables` with the number of `table` nodes.

xberg v1.2.4, release build, 2026-09-18. Any long public-sector PDF with many tables should reproduce the table gap. `extraction_method` was `mixed` for the two large files and `native` for the short one.

tobocop2:
## Related

Closes #1668

## Description

Asking for document structure returned paragraph nodes only. No headings, no tables, no page breaks.

The condition that decides whether the structured build runs checks the output format, the hierarchy option, inline image OCR and the content filter. It never checked whether document structure was requested. A configuration that asks only for structure therefore fell back to the flat path, which emits paragraphs by design.

The table count stayed accurate throughout, because it reads the document's own tables rather than the element tree. That is why the output could report tables while the structure contained none.

## Behaviour

Requesting document structure now runs the structured build. On a long mixed document the node kinds go from paragraph alone to code, group, heading, list, list item, page break, paragraph and table.

Configurations that already triggered the structured build are unaffected: the condition gains a term and loses none.

## Timing

Measured on a 731 page document, six interleaved runs per arm: 27.10 s flat against 27.88 s structured, with standard deviations of 1.4 s and 2.2 s. The difference is not distinguishable from run to run variation on this machine.

## Checklist

- [x] Confirmed on a real document, not only a fixture
- [x] fmt and clippy clean


Goldziher (changes requested):
Verified the fix and the claim about node kinds; found one real, undisclosed side effect.

**Confirmed:** `needs_structured_extraction` (crates/xberg/src/extractors/pdf/extraction.rs:75) on `main` checks only `hierarchy_enabled`, `output_format` (Markdown/Djot/Html/DocTags), `ocr_inline_images`, `content_filter_configured` — never `include_document_structure`. The new disjunct (line 84) is the minimal correct fix; other pre-existing disjuncts are untouched and their tests are unchanged. No other extractor shares this dual flat/structured gate shape — docx/odt/doc/ppt/bibtex/rtf just set the flag on their own configs, epub and the html converter gate directly on `include_document_structure` (epub/mod.rs:158) — so no missed call sites.

**Table count claim holds.** `extract_tables_native`/`bordered`/`heuristic` run independent of `needs_structured` (only `retain_hierarchy_segments` depends on it); confirmed via CLI on `test_documents/pdf/embedded_images_tables.pdf`: `counts.tables == 1` with `--include-structure` both `true` and `false`.

**Not disclosed: `content` (Plain output) changes too, not just `document.node_types`.** `derive_extraction_result` always runs `render_plain(&doc)` over whatever document `select_native_pdf_document` picked, and that function substitutes the *structured* document for the flat one whenever `needs_structured` is true — this substitution is pre-existing, but it was previously unreachable for `include_document_structure`-alone + Plain-format callers. On the same fixture, `--include-structure false` vs `true` (both Plain, otherwise-default config) gives 1643 vs 1589 chars of `content`, with table rows reformatted from raw reading-order text into a single row-per-line rendering, and headers/page-footer text reflowed. The PR body/commit only characterizes the table-count invariant and node kinds; it doesn't mention that `content` itself changes for the exact configuration it's fixing. That's a real behavior change to a heavily-consumed field for existing valid configs (not just newly-erroring ones), worth calling out explicitly in the CHANGELOG and covering with a regression test, so a caller who set only `include_document_structure=true` for the tree while depending on stable Plain `content` isn't silently surprised on upgrade.

Ask: add a line to the CHANGELOG entry (and ideally a test pinning/asserting the new content shape) noting that `content` is also affected when this flag is what newly triggers the structured pass.

Everything else checked out: `cargo test -p xberg --features pdf --lib pdf::extraction::tests::` 14/14 green; the new end-to-end test in mod.rs passes standalone and as part of the broader `pdf::` filter (1200 passed / 4 unrelated pre-existing fixture-missing failures, nothing to do with this change); planted-failure (dropped the `|| include_document_structure` disjunct) turned the new unit test RED with `assertion failed: needs_structured_extraction(false, true, &OutputFormat::Plain, false, false)`, restored clean. `poly lint` clean on the 3 changed files. No API/schema/enum changes (no new node kinds, no bindings/alef diff), so no regen or extra changelog entry needed there. Commit type/scope fine; no AI attribution; no never-list values.

tobocop2:
Extended the changelog bullet to state the part that was missing: this configuration now renders its plain text from the structured document, so the rendered content changes with it, table rows one per line and headings and page footers reflowed. It also records that the table count is unaffected.

You were right that this was the gap. The description characterised the node kinds and the table invariant and said nothing about the field most callers actually read.

The regression test pinning the new shape needs a build host that is down at the moment. I will add it before this merges rather than claim it here.

Goldziher (changes requested):
The changelog line is fine now. Holding merge on the regression test you said you would add before merging.

The entry makes a specific claim ("table rows render one per line") that nothing in the tree checks: the only end-to-end test (`extractors/pdf/mod.rs` `test_include_document_structure_alone_triggers_the_structured_path`) asserts `node_types` contains a non-paragraph kind and never reads `content`. `mod.rs` already carries a precedent of a fix deliberately gating on `output_format != Plain` to keep Plain bytes stable (the Mixed-branch merge near line 492), so a later change of that shape silently reverts this behaviour and the changelog becomes false with no test failing.

It is ~30 lines next to the existing test: extract `test_documents/pdf/embedded_images_tables.pdf` with `include_document_structure` false and true under `OutputFormat::Plain`, assert the table count is 1 in both, assert each row of `tables[0].cells` joined by a single space appears as its own line of the structured `content` (`render_table_plain` is `row.join(" ")` per row), and `assert_ne!` the two `content` strings.

Non-blocking: the fix commit subject is 75 chars (rule is under 72) — moot if this squashes.

tobocop2:
You added that test yourself in `c9159ee800` before I got to it, so I have added nothing rather than write a second copy of it. What was still missing was proof that it catches the regression you were worried about, so I ran that control.

### The mutation you named, and the test's response

You pointed at the precedent where a fix gates on `output_format != Plain` to keep Plain bytes stable, and said a later change of that shape would silently revert this behaviour with nothing failing. I reproduced exactly that shape in `extractors/pdf/extraction.rs:83`, narrowing the new disjunct to `include_document_structure && *output_format != OutputFormat::Plain`. The anchor matched once.

RED, rc=101:

```
test ...plain_output_with_document_structure_renders_each_table_row_on_its_own_line ... FAILED
test result: FAILED. 0 passed; 1 failed; 0 ignored; 3111 filtered out
panicked at crates/xberg/src/extractors/pdf/mod.rs:6766
```

Restored from `HEAD`, blob hash back to its original value. GREEN, rc=0: `1 passed; 0 failed`.

So the test does fail under the one change shape that would make the changelog false. It is not decoration.

### Gate on that head

`cargo test -p xberg --features pdf --lib pdf::` gives 1205 passed, 0 failed, rc=0. `cargo fmt --check` is clean. Clippy reports three dead-code warnings, all pre-existing under the pdf-only feature set and none in the files this touches.

One instrument note, because it nearly cost me the result: filtering the test by name with `--exact` selected nothing and still exited 0, reporting `0 passed; 3112 filtered out`. `--exact` matches the full module path. A substring filter plus `--list` is what established the test is really compiled into the pdf-feature binary.

PR #1686, opened 2026-09-19
fix(docx): read embedded image bytes when image OCR will run

### LINKED ISSUE #1662: bug: DOCX: an embedded PNG fails image OCR with 'Could not determine image format'
### Description

A DOCX with one embedded PNG raises an image OCR warning, and the image is not OCR'd. Text extraction succeeds. The image node in the document structure carries its alt text. Only the image OCR path fails.

Expected: the embedded PNG decodes. PIL writes a standard PNG, so the bytes handed to the OCR path are either not the PNG stream or lose their format hint between the DOCX extractor and the OCR backend.

### Steps to reproduce

1. Build the file below.
2. Extract it with the default OCR config (tesseract, eng) and `include_document_structure = true`.
3. Read `processing_warnings`.

xberg v1.2.4, 2026-09-18.

```
image_ocr: Image 0 OCR failed: OCR error: Tesseract OCR failed: Image processing failed: Parsing error: Could not determine image format
```

```python
import io
from docx import Document
from docx.shared import Inches
from PIL import Image
doc = Document()
doc.add_heading("Sample", 1)
img = Image.new("RGB", (200, 100), (30, 90, 200))
b = io.BytesIO(); img.save(b, "PNG"); b.seek(0)
doc.add_picture(b, width=Inches(2))
doc.save("structure.docx")
```

tobocop2:
## Related

Closes #1662

## Description

A PNG embedded in a DOCX failed image OCR with a message saying the format could not be determined. The bytes were fine. Nothing had been read.

The media part is read only when a predicate is true, and that predicate covered image extraction, captioning and QR codes but not image OCR. The other branch still attached the image with an empty buffer and a format string taken from the file extension. Format guessing then succeeded on empty input and returned no format.

The same shape affected PPT, PPTX and HTML, which share that predicate. They are fixed too.

## Behaviour

Embedded images are read when image OCR will run, so the detector receives real bytes.

## Scope

Reading the bytes and putting them in the result are now separate questions. A standalone image passes through OCR without its bytes being added to the result's image list, which is what happened before this change and what callers expect. Containers are unaffected.

## Checklist

- [x] Mutation arm shown red and restored, with the raw bytes visible in the output when reverted
- [x] Regression tests for the standalone case, DOCX and HTML
- [x] All arms run with the office and ocr features named, so the changed code is actually compiled

PPT and PPTX are fixed by the same change but have no regression test yet; that is tracked separately in #1687.


Goldziher (changes requested):
Logic and tests are solid. Two required-practice gaps block merge as-is.

**Predicate design (verified):** `runs_ocr_on_embedded_images()` in `crates/xberg/src/core/config/extraction/core.rs` is algebraically identical to the pipeline's own gate at `core/pipeline/mod.rs:340-342` (`image_ocr_enabled && config.ocr.is_some()`). `needs_image_data()` stays one shared function — `ppt.rs:256`, `pptx.rs:411,537`, `docx.rs:1074`, `html.rs:670`, `pdf/extraction.rs:546` all call it, nothing copied. Captioning/VLM and QR are already covered by the pre-existing disjuncts; no chunking path reads image bytes. `path_resolver::resolve_image_uris` (markdown/epub `UriKind::Image`) also runs through `needs_image_data()` and carries the identical latent defect for those formats — this fix silently corrects that too, worth a line in the PR description alongside the PPT/PPTX callout.

Ran (foreground, `CARGO_TARGET_DIR` scoped to the worktree):
```
cargo test -p xberg --features office,ocr,tokio-runtime --lib -- needs_image_data ocr_only_config_reads_real test_extract_with_ocr
```
→ `test result: ok. 7 passed; 0 failed` (+2 more matched separately: `a_config_that_ocrs_embedded_images_needs_the_image_data`, `a_config_that_does_not_ocr_embedded_images_still_skips_the_data`, both `ok`). 9/9 new or touched tests green.

Planted-failure proof: reverted `needs_image_data`'s OCR disjunct → `test_docx_ocr_only_config_reads_real_embedded_image_bytes` and `test_html_ocr_only_config_reads_real_inline_image_bytes` both FAILED (empty `left: []` vs real PNG bytes on the right; `left: 0` vs `right: 1` images). Separately reverted `wants_own_bytes_in_result` back to `needs_image_data` in `extractors/image.rs` → `test_extract_with_ocr_only_does_not_populate_images` FAILED, standalone image's raw PNG bytes leaking into `doc.images`. Both reverts restored via `git show HEAD:<path> > <path>`; worktree is clean.

Fixture (item d): synthetic, built in-test via the pre-existing `build_test_docx_with_files`/`build_docx_with_media` helpers — no external fixture or submodule needed.

**Blocking findings:**

1. `crates/xberg/src/core/config/extraction/core.rs`: `runs_ocr_on_embedded_images` and `wants_own_bytes_in_result` are declared `pub`, but neither needs crate-external visibility — `wants_own_bytes_in_result` is only called from `extractors/image.rs` (same crate), `runs_ocr_on_embedded_images` only from `needs_image_data` itself. Every `pub` method on `ExtractionConfig` gets an auto-generated wrapper in every binding crate (confirmed: `needs_image_data`, `default_archive_depth` etc. all have hand-matching wrappers in xberg-wasm/php/py/ffi/jni); `pub(crate)` methods like the existing `resolved_layout_acceleration` do not. Ran `alef verify --exit-code` on this worktree: it reports `xberg`'s generation inputs changed since its last successful `alef generate`. I confirmed this line is specific to this PR by running the identical check against `origin/main` in a scratch worktree — main also fails `alef verify` (a pre-existing, unrelated vendored-schema/version mismatch), but main's output has no "generation inputs changed" section at all. This PR will trip `alef-bindings-freshness` in CI. Fix: change both new methods to `pub(crate)` — avoids the churn entirely, no `alef generate` needed.

2. No `CHANGELOG.md` entry. `.ai-rulez/rules/changelog-policy.md` (priority: high) requires Fixed entries for user-visible bugs; this closes GH#1662 and, per the second commit's own message, silently fixes the identical defect in PPT/PPTX too — comparable in shape and user impact to the GH#1660 entry already under `[Unreleased]`/1.2.5. Needs an entry.

**Non-blocking suggestion:** no guard against an empty image buffer at the OCR boundary (`extraction/image_ocr.rs` / `image_decode.rs`'s `Could not determine image format`). This fix avoids the one path that produced it but doesn't add a guard, so any other path that leaves `data` empty (e.g. `docx.rs`'s `MAX_IMAGE_FILE_SIZE` guard, a zip read failure) will still surface the same misleading message instead of a clear "no image data" error. Worth a follow-up, not blocking this fix.

PR #1688, opened 2026-09-19
fix(docx): inherit list numbering from the paragraph style

### LINKED ISSUE #1663: bug: DOCX: list items are emitted as paragraph nodes in the document structure
### Description

List items in a DOCX come back as `paragraph` nodes in the document structure. No `list` or `list_item` node is emitted.

Expected: the three items form one `list` node with three `list_item` children. The `List Bullet` style in the python-docx default template carries a `w:numPr` numbering reference, so this is the ordinary Word list case. If the extractor only recognises lists through `w:numId` on the paragraph and this template attaches numbering through the style, the style-inherited case is the gap. Verify against a Word-authored file as well.

### Steps to reproduce

1. Build the file below.
2. Extract it with `include_document_structure = true`.
3. Read the nodes after the level-2 heading: three `paragraph` nodes with the item text.

xberg v1.2.4, 2026-09-18. Node type histogram for the whole file: group 6, heading 6, image 1, page_break 2, paragraph 6, raw_block 3, table 1.

```python
from docx import Document
doc = Document()
doc.add_heading("Scope", 2)
for i in range(3):
    doc.add_paragraph(f"List item {i+1}", style="List Bullet")
doc.save("lists.docx")
```

tobocop2:
## Related

Closes #1663

## Description

List items in a DOCX came back as paragraph nodes, so the list was lost from the document structure.

The numbering lives on the style, not on the paragraph. The reporter's document body carries no numbering reference at all. The paragraph properties had no numbering field and the parser never read one, so the style's reference never survived parsing and the list test was false for every item.

Two readers ask that question: one builds markdown, the other builds the list and list item structure nodes. Resolving the inheritance in either reader alone would have fixed markdown and left the structure flat. It is resolved once, onto the paragraph, before either reader runs.

## Behaviour

On a document whose list numbering comes from the style, the structure now contains one list holding three list items and no paragraphs, where it previously held three paragraphs and no list.

The style carries a numbering reference and no level. The markdown path needs both, so the level defaults to the outermost and an explicit level is clamped.

Covered across body text, table cells, headers, footers, footnotes, endnotes and comments.

## Known gap

Not a gap: `parse_notes`/`parse_comments` already flatten a table cell's paragraphs into the note's or comment's own `paragraphs`, so a list inside a table inside a footnote, endnote or comment does inherit style numbering through this fix.

## Checklist

- [x] Two mutation arms, one per half of the fix, shown red and restored
- [x] Confirmed at the output, not only at the predicate
- [x] All arms run with the office feature named


Goldziher (changes requested):
Verified: `cargo test -p xberg --features office --lib docx::parser::tests::` → 129 passed, 0 failed (includes the two new tests). `docx::styles::numbering_style_tests::` → 2 passed, 0 failed. Planted-failure probe (stubbed `resolve_style_numbering` to always return `None`) turned `a_paragraph_inherits_the_numbering_its_style_carries` red (`left: None right: Some(1)`, parser.rs:6195); restored, tests green again, working tree clean.

The `basedOn`-chain walk, style-vs-paragraph precedence, and single-resolution-for-both-readers design (`apply_style_numbering` called once in `parse_document`, consumed by both `to_markdown` and `build_internal_document`) are all correct.

Requesting changes on two points:

1. **Missing CHANGELOG.md entry.** Every other fix commit on this branch (`8a508fd567`, `b1e0bd42ed`, `73d9b586a1`, `6fed82b0f0`) adds its `CHANGELOG.md` entry in the same commit. This PR's commit (`4e895cac5b`) touches only `parser.rs`/`styles.rs`. Please add an entry under `## [Unreleased]` referencing GH#1663, matching the existing format.

2. **Test coverage doesn't match the PR body's claim.** The body says coverage spans "body text, table cells, headers, footers, footnotes, endnotes and comments," but both new tests in `parser.rs` call `apply_style_numbering` directly against a hand-built `Document`/`StyleCatalog` with a single top-level paragraph. Nothing exercises `fill_tables`, the headers/footers loop, or the footnotes/endnotes/comments loop in `apply_style_numbering` (parser.rs:1477-1498), and nothing parses real XML through `parse_document` to prove the wiring at `parser.rs:2314-2318` is actually hooked up end-to-end — the planted-failure probe above had to break `resolve_style_numbering` itself, not the call site, because reverting the call site doesn't fail either existing test. Please add at least one test per fan-out branch (a list-styled paragraph inside a table cell, and one inside a footnote/comment), and one that goes through `parse_document`/`extract_text_with_boundaries` or `build_internal_document` on real XML rather than only the extracted helper function.

Non-blocking:

- `resolve_style_numbering`'s bounded `basedOn` chain-walk (parser.rs:1509) is a third near-duplicate of the same loop already in `resolve_heading_level` and `resolve_style_name` in this file — worth extracting to a shared helper at some point.
- Pre-existing, not introduced here: `w:numId="0"` on a paragraph parses to `Some(0)`, which correctly blocks style inheritance (precedence honored), but both readers still test `numbering_id.is_some()` to mean "is a list," so an explicit "no numbering" override still renders as a list. Separate from this fix.

tobocop2:
Added the changelog entry referencing the issue, matching the format of the other fixes on this branch.

On the coverage: you are right and the description overstated it. The tests drive the helper against a hand-built document with one top-level paragraph, so nothing exercises the table, header and footer, or footnote and comment fan-out, and nothing proves the call site is wired at all. Your own probe having to break the resolver rather than the call site is the proof of that.

Those tests need a build host that is down at the moment. I will add one per fan-out branch plus one that goes through the real XML path, and push them before asking for this again.

On the repeated chain walk: agreed, it is the third copy in that file and wants extracting. On the explicit zero numbering still rendering as a list: also agreed, and separate from this fix.

Goldziher (changes requested):
Changelog entry is right. The test request is still open: `88c050c1c2` touches only `CHANGELOG.md`, and the tree still has just the two helper-level tests calling `apply_style_numbering` on a hand-built `Document`.

I re-ran the control: with the `apply_style_numbering` call in `DocxParser::parse` (parser.rs:2315-2318) deleted, `a_paragraph_inherits_the_numbering_its_style_carries`, `paragraph_numbering_outranks_the_numbering_on_its_style` and `test_bullet_list_extraction` all still pass. No test in the tree fails when the fix is unwired.

The tests below are validated on this head (all pass; the five `parse_document` ones go red with the call site removed). Five go in `parser.rs` `mod tests` (body, table cell, header, footnote, comment through `parse_document`); one goes in `extractors/docx.rs` `mod tests` and asserts `(1 ListStart, 3 ListItem, 0 Paragraph)` on the reporter's shape. Footnote ids -1/0/1 are separators, so the footnote one uses `w:id="2"`.

```rust
    // crates/xberg/src/extraction/docx/parser.rs, inside `mod tests`
    const LIST_BULLET_STYLES_XML: &str = r#"<?xml version="1.0" encoding="UTF-8"?>
<w:styles xmlns:w="http://schemas.openxmlformats.org/wordprocessingml/2006/main">
  <w:style w:type="paragraph" w:default="1" w:styleId="Normal"><w:name w:val="Normal"/></w:style>
  <w:style w:type="paragraph" w:styleId="ListBullet">
    <w:name w:val="List Bullet"/>
    <w:basedOn w:val="Normal"/>
    <w:pPr><w:numPr><w:numId w:val="1"/></w:numPr></w:pPr>
  </w:style>
</w:styles>"#;

    fn list_styled_paragraph(text: &str) -> String {
        format!(r#"<w:p><w:pPr><w:pStyle w:val="ListBullet"/></w:pPr><w:r><w:t>{text}</w:t></w:r></w:p>"#)
    }

    fn docx_with_list_style(body: &str, extra_parts: &[(&str, &str)]) -> Vec<u8> {
        use std::io::{Cursor, Write};
        let mut zip = zip::ZipWriter::new(Cursor::new(Vec::new()));
        let options = zip::write::FileOptions::<()>::default().compression_method(zip::CompressionMethod::Stored);
        zip.start_file("word/document.xml", options).unwrap();
        zip.write_all(wrap_body(body).as_bytes()).unwrap();
        zip.start_file("word/styles.xml", options).unwrap();
        zip.write_all(LIST_BULLET_STYLES_XML.as_bytes()).unwrap();
        for (path, xml) in extra_parts {
            zip.start_file(*path, options).unwrap();
            zip.write_all(xml.as_bytes()).unwrap();
        }
        zip.finish().unwrap().into_inner()
    }

    fn parse_bytes(bytes: &[u8]) -> Document {
        let mut budget = SecurityBudget::with_defaults();
        parse_document(bytes, &mut budget, &default_limits()).expect("docx parses")
    }

    #[test]
    fn a_body_paragraph_inherits_list_numbering_from_its_style_through_parse_document() {
        let body = format!("{}{}", list_styled_paragraph("First"), list_styled_paragraph("Second"));
        let doc = parse_bytes(&docx_with_list_style(&body, &[]));
        assert_eq!(doc.paragraphs.len(), 2);
        assert_eq!(doc.paragraphs[0].numbering_id, Some(1));
        assert_eq!(doc.paragraphs[0].numbering_level, Some(0));
        let md = doc.to_markdown(true);
        assert_eq!(md, "- First\n- Second", "markdown: {md:?}");
    }

    #[test]
    fn a_table_cell_paragraph_inherits_list_numbering_from_its_style() {
        let body = format!("<w:tbl><w:tr><w:tc>{}</w:tc></w:tr></w:tbl>", list_styled_paragraph("In a cell"));
        let doc = parse_bytes(&docx_with_list_style(&body, &[]));
        let cell = &doc.tables[0].rows[0].cells[0];
        assert_eq!(cell.paragraphs[0].numbering_id, Some(1));
        assert_eq!(cell.paragraphs[0].numbering_level, Some(0));
    }

    #[test]
    fn a_header_paragraph_inherits_list_numbering_from_its_style() {
        let header = format!(
            r#"<w:hdr xmlns:w="http://schemas.openxmlformats.org/wordprocessingml/2006/main">{}</w:hdr>"#,
            list_styled_paragraph("In a header")
        );
        let doc = parse_bytes(&docx_with_list_style("", &[("word/header1.xml", &header)]));
        assert_eq!(doc.headers.len(), 1, "header part was picked up");
        assert_eq!(doc.headers[0].paragraphs[0].numbering_id, Some(1));
    }

    #[test]
    fn a_footnote_paragraph_inherits_list_numbering_from_its_style() {
        let footnotes = format!(
            r#"<w:footnotes xmlns:w="http://schemas.openxmlformats.org/wordprocessingml/2006/main"><w:footnote w:id="2">{}</w:footnote></w:footnotes>"#,
            list_styled_paragraph("In a footnote")
        );
        let doc = parse_bytes(&docx_with_list_style("", &[("word/footnotes.xml", &footnotes)]));
        assert_eq!(doc.footnotes.len(), 1);
        assert_eq!(doc.footnotes[0].paragraphs[0].numbering_id, Some(1));
    }

    #[test]
    fn a_comment_paragraph_inherits_list_numbering_from_its_style() {
        let comments = format!(
            r#"<w:comments xmlns:w="http://schemas.openxmlformats.org/wordprocessingml/2006/main"><w:comment w:id="0" w:author="A">{}</w:comment></w:comments>"#,
            list_styled_paragraph("In a comment")
        );
        let doc = parse_bytes(&docx_with_list_style("", &[("word/comments.xml", &comments)]));
        assert_eq!(doc.comments.len(), 1);
        assert_eq!(doc.comments[0].paragraphs[0].numbering_id, Some(1));
    }
```

```rust
    // crates/xberg/src/extractors/docx.rs, inside `mod tests`
    #[tokio::test]
    async fn style_carried_numbering_produces_a_list_in_the_document_structure() {
        use crate::types::internal::ElementKind;
        let styles_xml = r#"<?xml version="1.0" encoding="UTF-8"?>
<w:styles xmlns:w="http://schemas.openxmlformats.org/wordprocessingml/2006/main">
  <w:style w:type="paragraph" w:default="1" w:styleId="Normal"><w:name w:val="Normal"/></w:style>
  <w:style w:type="paragraph" w:styleId="ListBullet">
    <w:name w:val="List Bullet"/><w:basedOn w:val="Normal"/>
    <w:pPr><w:numPr><w:numId w:val="1"/></w:numPr></w:pPr>
  </w:style>
</w:styles>"#;
        let document_xml = r#"<?xml version="1.0" encoding="UTF-8"?>
<w:document xmlns:w="http://schemas.openxmlformats.org/wordprocessingml/2006/main"><w:body>
  <w:p><w:pPr><w:pStyle w:val="ListBullet"/></w:pPr><w:r><w:t>One</w:t></w:r></w:p>
  <w:p><w:pPr><w:pStyle w:val="ListBullet"/></w:pPr><w:r><w:t>Two</w:t></w:r></w:p>
  <w:p><w:pPr><w:pStyle w:val="ListBullet"/></w:pPr><w:r><w:t>Three</w:t></w:r></w:p>
</w:body></w:document>"#;
        let data = build_test_docx_with_parts(document_xml, Some(styles_xml), None, None, None, None, None);
        let internal_doc = extract_docx_internal_document(&data).await;
        let kinds: Vec<&ElementKind> = internal_doc.elements.iter().map(|e| &e.kind).collect();
        let list_starts = kinds.iter().filter(|k| matches!(k, ElementKind::ListStart { .. })).count();
        let list_items = kinds.iter().filter(|k| matches!(k, ElementKind::ListItem { .. })).count();
        let paragraphs = kinds.iter().filter(|k| matches!(k, ElementKind::Paragraph)).count();
        assert_eq!((list_starts, list_items, paragraphs), (1, 3, 0), "elements: {kinds:?}");
    }
```

Non-blocking: the `numPr` arm pushes `parse_paragraph_properties` (`styles.rs:381`) to 85 lines and trips poly `function-too-long` (warning only; main is clean) — a `parse_numbering_properties(&node) -> (Option<i64>, Option<i64>)` helper clears it. The "Known gap" paragraph in the description is wrong in the safe direction: `parse_notes`/`parse_comments` flatten table-cell paragraphs into the note's `paragraphs`, so a list inside a table inside a footnote or comment is covered.

tobocop2:
You landed these tests in `24ab8bd8a3` nine minutes before my own commit of the same nine items, so I am discarding mine rather than pushing a duplicate. Yours is the better version: the extra rustdoc and the footnote-id note earn the eight lines, and `155bd5d0f7` also does the `parse_numbering_properties` extraction your review suggested.

One correction worth having, because it is in your favour.

### The control fails six tests, not five

Deleting the `apply_style_numbering` call site at `parser.rs:2315-2318` and running `cargo test -p xberg --features office --lib docx::` gives rc=101 and `FAILED. 310 passed; 6 failed`, against `316 passed; 0 failed` on the unmutated tree.

The sixth is `extractors::docx::tests::style_carried_numbering_produces_a_list_in_the_document_structure`. Your review claimed the five `parse_document` tests go red; the reporter-shape test does too. Five report `left: None / right: Some(1)`; the sixth reports `elements: [Paragraph, Paragraph, Paragraph]` against the expected `(1, 3, 0)`, which is the reported symptom stated directly.

I measured that on a tree carrying these tests at 122 lines rather than your 130; the test bodies are token-identical, so it should hold on `155bd5d0f7` unchanged. Say the word if you want it re-run against your tip.

### Two feature cells, both pre-existing

While gating I checked two configurations and proved both are unrelated to this change, identical at `88c050c1c2` and at head:

- `--features office,pdf`: tests are rc=0 at both ends. Only `clippy -- -D warnings` fails, with the same three dead-code errors in `extractors/pdf/ocr/rendering.rs`.
- `--features office,xml`: genuinely red at both ends, `E0433 cannot find diagram in extraction` at `extractors/xml.rs:268`.

Neither is introduced here. Flagging them because a reviewer hitting either would otherwise attribute them to this branch.

PR #1689, opened 2026-09-19
fix(pdf): render pages in parallel so OCR scales with the thread budget

### LINKED ISSUE #1666: bug: PDF OCR: throughput on one document does not scale with the thread budget or the tesseract pool
### Description

OCR throughput on one PDF does not change with the thread budget or with the tesseract instance pool. On a 32-core host, a 240-page PDF with 80 scanned pages (200 dpi, A4) takes the same time at every setting:

| tesseract pool | max_threads 4 | 8 | 16 |
|---|---|---|---|
| 4 (stock) | 11.7 s | 10.8 s | 11.2 s |
| 16 (patched constant) | 13.0 s | 10.6 s | 10.0 s |

Four times the threads and four times the tesseract instances give the same wall time. At about 8 scanned pages per second, the per-document path cannot use a larger machine to go faster. `MAX_TESSERACT_APIS` is a constant of 4 with a comment that four workers beat eight on documents per second; that measurement was across documents, not pages of one document, and it is not the binding limit here either.

Expected: profile one mixed document under the per-page OCR route and name the serial stage. Candidates: page rasterisation ahead of the batch, the batch barrier (a batch waits for its slowest page), and the post-OCR merge. Then expose the tesseract pool size or remove the constant once the real limit is known.

### Steps to reproduce

1. Build a 240-page PDF where every third page is a full-page 200 dpi JPEG raster with no text layer.
2. Extract with the default OCR config (tesseract, eng) at `concurrency.max_threads` 4, 8 and 16, one process per run.
3. Repeat with `MAX_TESSERACT_APIS` raised to 16.

xberg v1.2.4, release build, 2026-09-18, two runs per cell, first-run numbers. The host was shared with other builds (load 15 to 33), so absolute numbers are noisy; the shape is flat. At 32 threads OCR does not run at all (#1665).

tobocop2:
## Related

Closes #1666

## Description

OCR throughput on one document did not improve with more cores. Page rasterization ran in a sequential loop regardless of the thread budget, while the next stage already spread the same batch across threads.

The document handle is safe to share, so nothing structural prevented rendering in parallel. The fix uses the same pattern the encoding step already uses.

## Measured

240 page document with 80 scanned pages, release build, fresh process per run, five runs per cell.

| threads | before | after |
|---|---|---|
| 4 | 10.8 s | 3.2 s |
| 16 | 10.7 s | 0.97 s |

Before, the time is flat: more cores buy nothing. After, it scales, and the spread at 16 threads is 0.003 s.

Page order in the output is preserved.

## What this does not cover

No existing test fails if this change is reverted. The suite gives identical results either way, so a regression here would only be visible in a benchmark. Adding a guard is tracked separately in #1690.

## Checklist

- [x] Curve measured before and after, several runs per cell, spread reported
- [x] Confirmed the benchmark drives the code path this change touches
- [x] Order preservation and shared-handle safety both checked
- [x] fmt and clippy clean with the pdf and ocr features named


Goldziher (approved):
Verified independently:

- Fix not already on `main`: `render_selected_pages_from_document` on `origin/main` is the old sequential loop; diff applies cleanly on top.
- Shared-handle safety: `xberg_native_pdf::PdfDocument` has no `unsafe impl`, no `RefCell`/raw pointers on the concurrent path — every field is `Mutex`/`Arc`/atomic (checked `crates/xberg-native-pdf/src/document/mod.rs:537-729`), so it's auto-`Send + Sync`. `render_page_with_safeguards` / `get_page_dimensions_pt` take `&PdfDocument`. `cargo test` compiling `par_iter().map(|&idx| render_one_selected_page(doc, ...))` with `doc: &PdfDocument` captured across the closure is itself proof the type is `Sync`.
- Order: `.par_iter().map(...).collect::<Result<Vec<_>>>()` preserves index order (rayon collect-into-Vec is order-preserving); confirmed against `render_dpi_tests` and `test_render_selected_pages_for_ocr_wide_pdf_does_not_fail`, all pass.
- Error handling: `.map(...).collect()`, not `filter_map` — an `Err` propagates, not swallowed. One nuance worth knowing (not blocking): rayon's parallel `Result` collect doesn't guarantee returning the *first*-index error the way the old sequential loop did if multiple pages fail simultaneously; whichever error does surface still names its own correct page number.
- Memory/#1684 composition: this PR touches only `rendering.rs`; #1684 (open, unmerged, `fix/1665-...`) changes `batch_size` computation in `pipeline.rs` upstream of the call site at `pipeline.rs:339`, not the render call itself. Whichever lands first, this PR only parallelizes rendering of whatever slice `batch_size` already produced — it doesn't change batch sizing, so it doesn't reintroduce the peak #1684 addresses. The only new memory cost is transient (concurrent raw pre-decode buffers instead of one), bounded by the same thread budget the encode step already parallelizes over.
- Cfg-gating (`tokio-runtime` + `not(target_arch = "wasm32")`) is copied verbatim from the pre-existing PNG-encode `par_iter` pattern a few lines away in `pipeline.rs` (lines ~618-641) — consistent, not new.
- Test: none added. Confirmed the PR's own claim — reverted the hunk to the `origin/main` version and ran `cargo test -p xberg --features pdf,ocr,ocr-pipeline,tokio-runtime --lib -- extractors::pdf::ocr::`: 226 passed, 5 failed both before and after (identical failures, all `missing fixture` — the `test_documents` submodule is empty in this environment, unrelated to this change). Follow-up #1693 adds a thread-dispatch-based regression guard on top of this PR; acceptable given that exists.
- `render_full_pdf_ocr_batch` (rendering.rs:513, used by the separate whole-document `extract_with_ocr` route) has the identical sequential-render pattern and is not touched here. Out of scope — issue #1666 is specifically about "the per-page OCR route" (`extract_mixed_ocr_native`) — but worth a follow-up issue since it's the same defect shape.
- `cargo fmt -p xberg -- --check` and `cargo clippy -p xberg --features pdf,ocr,ocr-pipeline,tokio-runtime --lib -- -D warnings` both clean.
- Commit type/message, no AI attribution, no never-list values: clean.
- Could not independently verify the body's measured timings (4/16-thread wall-clock numbers) — no benchmark run here.

PR #1691, opened 2026-09-19
fix(ocr): stop candle engine pools from building duplicate engines

### LINKED ISSUE #1683: bug: concurrent cold starts of the same OCR engine exhaust GPU memory
### Deployment

Local development (docker-compose)

### Version

Branch `fix/1672-source-dpi-backend-options`, measured on one NVIDIA L4 with 24 GB.

### Description

Each candle backend keeps a pool of engines. When several pages start at once and the engine they need is not yet cached, each page starts its own cold load of the same engine. The loads run together and their combined memory can exceed the card, although one loaded engine fits with room to spare.

The failure is a GPU out-of-memory error on pages that would succeed on their own. Steady state is not the problem; the cold start is.

Measured on a 22-page document:

| backend | pages lost to out-of-memory |
|---|---|
| PaddleOCR-VL | 3 of 22 |
| GLM-OCR | 5 of 22 |
| DeepSeek-OCR | 22 of 22, until the run was forced to a single thread |

DeepSeek-OCR needs the most memory per engine, so it fails first and hardest. Forcing one thread makes it pass, which points at the concurrent cold start rather than the model.

The pool serves one engine per backend once it is warm. The gap is that nothing stops several callers from loading it at the same time.

### Steps to reproduce

1. Run OCR over a multi-page PDF with a candle backend and the default thread budget.
2. Use a document long enough that several pages start before the first engine finishes loading.
3. Watch GPU memory during the first seconds of the run.
4. Repeat with a single thread. The out-of-memory errors stop.

### Logs and configuration

Out-of-memory errors appear only in the first seconds of a run. Once one engine is cached, later pages share it and the run proceeds normally.

tobocop2:
## Related

Closes #1683

## Description

Several pages starting at once could each begin loading the same uncached OCR engine. Every load allocated on the card, and together they exhausted it, although one loaded engine fits with room to spare.

Each backend kept its own pool that took a read lock, missed, released the lock, loaded the engine holding no lock, then took a write lock and inserted after a second check. The comment on that path said the second check stopped two callers both paying for initialisation. It did not: by the time it runs, both callers have already built an engine. It prevented two being stored, never two being built.

Measured on one card over a 22 page document: PaddleOCR-VL lost 3 pages, GLM-OCR lost 5, and DeepSeek-OCR lost all 22 until the run was forced to a single thread. DeepSeek-OCR needs the most memory per engine, so it failed first.

## The change

The four pools are removed. All of them now use the cache this repository already provides for embeddings, reranking, late interaction and sparse embeddings, which holds its lock across the load so a second caller for the same key waits rather than building another.

TrOCR is included. It was not in the report because it was not benchmarked, not because it was sound.

A failed load leaves nothing behind, which that cache guarantees rather than each pool arranging it.

## One property changes

The cache holds a single lock across a load, so loads of different keys also serialise. That is correct here and matches the rest of the repository, and it is coarser than waiting per key.

## How this was caught

A test already spawned five callers for one key and counted how often the loader ran, then asserted the count was at least one. Five passes that. One passes it too. The assertion is now an equality, which fails on the old code with five against one.

## Not verified here

The memory result needs the card. The build machine has no GPU, so this change is verified as correctness and not as a measured reduction in memory.


Goldziher (changes requested):
Verified the fix and the concurrency claim. The pool change itself is correct; one convention gap blocks.

**Correctness (verified)**
- `EngineCache::get_or_try_init` (`crates/xberg/src/engine_cache.rs:43`) holds the `RwLock` write guard across the whole `init()` call, so a second caller for the same key waits instead of building a duplicate. Confirmed by planting the old bug back in (revert to read→miss→build-unlocked→write→double-check) and re-running `candle_ocr::glm_ocr_backend::tests::test_pool_get_or_init_concurrent_access`: `left: 5, right: 1` (RED). Restored the file, reran: 3/3 pass (GREEN).
- All four backends (GLM-OCR, PaddleOCR-VL, DeepSeek-OCR, TrOCR) plus GLM-OCR's layout-model pool are migrated off the old hand-rolled double-checked pattern onto `EngineCache`; no leftover instance in `candle_ocr/`.
- Failure path is sound: `Arc::new(init()?)` short-circuits before `cache.put`, so a failed load inserts nothing and the next caller retries (`test_pool_get_or_init_failed_load_does_not_poison`, passing).
- Lock is per-cache, not per-key (acknowledged in the PR body and in the code comment at `glm_ocr_backend.rs:98-105`) — a slow load for one key does block a different key on the *same* backend's static, but each backend has its own separate `ENGINE_POOL`/`LAYOUT_POOL` static, so the four backends don't serialize against each other. Coarser than per-key single-flight but matches the existing convention for embeddings/reranker/late-interaction/sparse-embeddings, and it fixes the reported OOM.
- `get_or_init_engine` calls run inside `tokio::task::spawn_blocking` in every backend (e.g. `glm_ocr_backend.rs:466`), so the std `RwLock` write guard is never held across an `.await`.
- `cargo test -p xberg --features candle-trocr,candle-paddleocr-vl,candle-glm-ocr,candle-deepseek-ocr --lib candle_ocr::`: 93 passed, 2 failed (pre-existing, unrelated — `trocr_backend::tests::test_reject_whole_page_input_*` fail on missing `test_documents/images/*.png` fixtures; same on `main`, submodule not populated in this checkout).
- No public API surface change, no never-list values, no AI attribution, commit type/subject conform (`fix(ocr): ...`, 70 chars).

**Blocking**
- No `CHANGELOG.md` entry. Every other recent `fix(...)` commit on this branch (`8a508fd567`, `6fed82b0f0`, `94ac7521fb`, `9c2c384b6c`, `231e33f3e3`, `0f82b29ab4`) adds its bullet to `## [Unreleased]` / `### Fixed` in the same commit. This PR's diff against the merge-base (`d0a25a2e01`, current `main`) touches no changelog content at all. Please add an entry.

**Non-blocking, for awareness**
- `crates/xberg/src/llm/client.rs:580` and `:604` document `LLM_CLIENT_POOL`'s pattern as "matching `candle_ocr::trocr_backend::get_or_init_engine`" and claim it stops two racing callers from both paying construction cost — the same false claim this PR's commit message calls out as the bug in the OCR pools. `trocr_backend::get_or_init_engine` no longer has that shape after this PR, so the comment is now both stale and pointing at what looks like the same defect. `chunking/tokenizer_cache.rs`, `text/ner/candle.rs`, and `text/ner/gline.rs` have the same read→miss→build-unlocked→write→double-check shape too. Worth a follow-up issue; out of scope for this PR's title.

tobocop2:
Added the changelog entry under `[Unreleased]` / `Fixed`, matching the format of the surrounding bullets. That was the only blocking finding, so this should be ready.

On the other pools: you are right, and the comment in the client pool is now both stale and pointing at the same defect it claims to avoid. The tokenizer cache and the two named-entity backends carry the same shape. I will file that as its own issue rather than widening this one, since the fix there is the same one-line substitution and wants its own red-before proof per call site.

Goldziher (approved):
Changelog entry is in the right section and format. Two wording points I will fold into the changelog catch-up on main: loads of different keys on the same backend now serialise (a cold-start change worth a clause), and "the shared engine cache" reads as one cache across backends when each backend keeps its own static.

Correction on the follow-up scope from the last round: `tokenizer_cache.rs:168`, `ner/candle.rs:50` and `ner/gline.rs:370` all take the write lock before building, so they already have the shape this PR moves to. Only `llm/client.rs:613-615` builds unlocked, and it is not the same substitution: `LLM_CLIENT_POOL` wants a hard ceiling with no eviction (GH#1465) while `EngineCache`'s bound evicts LRU. Scope the follow-up to that one site plus its two stale doc comments (`:580`, `:603`).

PR #1692, opened 2026-09-19
test(office): cover embedded-image OCR for PPT and PPTX

### LINKED ISSUE #1687: bug: PPT and PPTX embedded image OCR has no regression test
### Deployment

Other

### Version

Branch `fix/1662-docx-embedded-image-ocr`.

### Description

PPT and PPTX shared the predicate that decided whether an embedded image's bytes were read. That predicate omitted image OCR, so an embedded image arrived at the format detector as an empty buffer and OCR failed with a message about the format.

The fix for the DOCX case corrects the shared predicate, so PPT and PPTX are fixed too. Neither has a regression test.

The same fix has regression tests for DOCX, HTML and the standalone image case. These two are the gap.

### Steps to reproduce

1. Build a PPTX holding an embedded raster image.
2. Extract with image OCR enabled and image extraction disabled.
3. Before the fix, OCR fails on that image because no bytes were read.

The PPT case is the same through the binary container.

### Logs and configuration

None needed. This is a coverage gap on a path already fixed, not a live defect.

The fixtures are the work: a slide XML document for PPTX and a binary container for PPT. Building either carelessly produces a test that asserts something other than what it claims, which is why they were not added under time pressure.

tobocop2:
## Related

Closes #1687

## Description

PPT and PPTX share the predicate that decides whether an embedded image's bytes are read. That predicate omitted image OCR, so an embedded image reached the format detector as an empty buffer.

Both were already fixed by the shared-predicate change in #1686, which this branch builds on. Neither had a regression test. DOCX, HTML and the standalone image case did.

## What the tests assert

An embedded raster image inside each container, extracted with image OCR enabled and image extraction disabled, reaches the document with real bytes.

The PPTX fixture is built from slide XML with a real image member. The PPT fixture reuses the synthetic container builder already used by the tests beside it, rather than a second hand-written container writer.

## Evidence

Reverting the predicate makes all four office and OCR regression tests fail, and each fails for the right reason: an image count of zero against one, or empty bytes against real ones. None fails with a panic or an unrelated error. Restoring returns the same result line.

Run without the office and ocr features, none of the four tests exists in the binary at all, so naming the features is load-bearing rather than cosmetic.

## Base

This branch sits on the fix it covers. It should merge after #1686.


Goldziher (approved):
Reviewed against origin/fix/1662-docx-embedded-image-ocr (#1686), not main — the GitHub-reported +330/-8 across 6 files is the combined #1686+#1692 diff vs main (PR base is set to main while the head branch still carries #1686's two unmerged commits). The PR-specific diff is +105/-0 in 2 files (crates/xberg/src/extraction/ppt/mod.rs, crates/xberg/src/extractors/pptx.rs), both pure test additions — no production code in this PR.

Verified:
- Planted failure: reverted the `|| self.runs_ocr_on_embedded_images()` disjunct in `needs_image_data()` (core.rs). All 4 office/OCR regression tests (docx, html, ppt, pptx) go RED, each on the claimed failure shape — `left: 0, right: 1` for image count (ppt/pptx/html) or empty vs real bytes (docx). Restored, reran: `cargo test -p xberg --features office,ocr ocr_only_config --lib` → 4 passed; 0 failed.
- PPT fixture is synthetic and built in-test (`build_test_ppt_ole` + `blip_record_one_uid` + `container`, all pre-existing helpers) — no new committed binary.
- Assertions compare exact bytes, not just non-empty: PPT test checks against real JPEG-magic bytes (`\xFF\xD8\xFF...`), PPTX test checks against the exact archived payload.
- `cargo clippy -p xberg --features office,ocr --tests` and `poly lint` on both changed files are clean. New doc comments are `///` (exempt from poly's uncomment pass via `remove_docs = false` in poly.toml) so no `~keep` requirement applies. No changelog entry needed — test-only change per `.ai-rulez/rules/changelog-policy.md`. Commit type `test` is valid, no AI attribution.
- No other `needs_image_data()` call site has the same gap: PDF (`extractors/pdf/extraction.rs`) already ORs a separate `ocr_inline_images` check independent of this predicate.

Nit (non-blocking): commit header is 83 chars, over the repo's 72-char convention.

Approving.

PR #1693, opened 2026-09-19
test(ocr): guard parallel page rendering by observing thread dispatch

### LINKED ISSUE #1690: bug: parallel page rendering has no regression guard
### Deployment

Other

### Version

Branch `fix/1666-ocr-throughput-scaling`.

### Description

Page rasterization now runs across the thread budget instead of sequentially, which takes a 240 page document from 10.7 s to 0.97 s at 16 threads.

No test covers it. Reverting the change entirely and running the full OCR suite gives identical results, 239 passed and 0 failed on both trees. A regression would be invisible to the suite and would only appear in a benchmark nobody runs on a pull request.

That is the same shape as the defect this fixed: throughput stopped scaling and nothing reported it.

### Steps to reproduce

1. Revert the parallel rendering change.
2. Run the OCR test suite with the pdf and ocr features.
3. Observe identical results to the unreverted tree.

### Logs and configuration

A guard needs to assert scaling rather than a fixed duration, since a wall clock threshold on a shared machine becomes a flaky test. One option is to assert that a larger thread budget produces a materially shorter time on a fixed document, with a generous margin, so only the loss of scaling fails it rather than ordinary load variation.

tobocop2:
## Related

Closes #1690

## Description

Rendering pages in parallel took a 240 page document from 10.7 s to 0.97 s at 16 threads. Nothing tested it. Reverting the change entirely left the suite byte identical, 239 passed either way, so a regression would have been invisible until someone ran a benchmark.

That is the same shape as the defect it fixed: throughput stopped scaling and nothing reported it.

## Why this is not a timing test

A duration threshold, or a ratio between two thread counts, is still two wall clock samples on a shared machine. This repository already has one assertion of that shape, comparing two samples of a live quantity against a constant, and it passes on one host and fails on another.

So the guard asserts the mechanism instead of its consequence. Rendering records which thread handled each page, the test builds its own four thread pool, renders twenty pages, and asserts more than one thread took part. That is a structural fact with no timing component, so ordinary load on a busy machine cannot move it.

The recorder is test only and never compiled into a release build.

## Proven to fire

Reverting the rendering body to the previous sequential loop makes the test fail with the count it saw, one distinct thread. Restoring it passes again.

## Stability

An explicit thread count on the pool short circuits before the rayon environment variable is read, confirmed by reading the dependency's source rather than assuming it. The only machine that could fail this test is one that cannot schedule more than one of the process's own threads for the whole batch, which is a single core quota rather than a load condition.

## Base

Sits on the change it guards. Merge after #1689.


Goldziher (approved):
Reviewed the PR's own diff (`origin/fix/1666-ocr-throughput-scaling...HEAD`): +120/-0 across `rendering.rs` (+29) and `tests.rs` (+91), purely additive. The `-31` in the stated `+185/-31` belongs entirely to already-approved #1689, not this commit — confirmed via `git diff origin/main...8797dca7ef`.

**Mechanism check (a):** `record_render_thread()` is called directly inside the production `render_one_selected_page`, gated `#[cfg(test)]` (no feature flag) — not a trait/closure seam. It's a test-only hook embedded in a production function rather than DI, which is the pattern the repo's testing conventions generally discourage. It is, however, cfg(test)-only (confirmed absent from a `--tests`-free build) and clearly documented. Not blocking, but a seam (e.g. an injectable per-page callback) would be the cleaner shape if this needs a follow-up.

**Planted failure (b):** forced `render_selected_pages_from_document` to the unconditional sequential branch (`.iter()` instead of `.par_iter()`), rebuilt:
```
cargo test -p xberg --lib --features "ocr,pdf" parallel_render_dispatches_across_more_than_one_thread -- --test-threads=1
```
→ `FAILED`, `panicked at .../tests.rs:3436:9: expected page renders to be observed on more than one OS thread ..., got 1 distinct thread(s): {ThreadId(3)}`. Restored via `git show HEAD:crates/xberg/src/extractors/pdf/ocr/rendering.rs > ...`; `git status` clean afterward.

**Flakiness (c):** the test builds its own 4-thread pool and renders 20 pages, independent of host CPU count — sound. I was concerned the global `RENDER_CALL_THREAD_IDS` static could pick up a thread ID from `test_render_selected_pages_for_ocr_wide_pdf_does_not_fail` (an unrelated, non-`#[serial]` test in the same file that also renders through the instrumented path) running concurrently, producing a false pass under the planted sequential regression. Ran the full `ocr::` module 10x under default parallelism with the regression planted — all 10 correctly RED — and 5x clean at HEAD — all pass (plus 5x in isolation with `--test-threads=1`). No false pass observed in 15 runs, so this stays a theoretical fragility rather than an observed defect.

**Other:** commit types valid (`fix`, `test`); no AI attribution; no never-list values in the diff; `poly lint` / `poly fmt --check` clean on both changed files; no public API surface touched (all new items `pub(super)`/private or cfg(test)-only), so no Alef regen needed; no changelog entry needed (precedent: prior `test(...)` commits don't touch `CHANGELOG.md`). Minor nit: the test commit's headline is 79 chars, over the 72-char convention.

Environment gaps, not chased: 5 pre-existing failures elsewhere in the same module run (`should_recover_blank_pipeline_page_and_warn_when_embedded_image_exists` etc.) are missing `test_documents` submodule fixtures, unrelated to this change.

No blocking defects found.

PR #1695, opened 2026-09-19
fix(ocr): populate the aggregate confidence on the page route and weight both routes alike

### LINKED ISSUE #1677: bug: OCR aggregate confidence is never populated on the page-level OCR route

tobocop2:
## Related

Closes #1677

## Description

The aggregate OCR confidence was never filled in when a PDF went through the page level OCR route. Only the embedded image route populated it. So a document OCRed page by page reported no aggregate although every page carried its own confidence, and a caller could not tell "OCR did not run" from "OCR ran and nothing filled this in".

## The field now means one thing

Populating it exposed a second problem. The existing route computed a flat mean over OCR elements. The native elements from PaddleOCR are line level, and the default minimum level lets them through unfiltered, so that mean was confidence per line while the field documents itself as confidence per word.

Both routes now compute the word count weighted mean the field describes, through one shared helper.

## This changes the existing route's number

Any document whose lines differ in length will report a different aggregate than before on the embedded image route. That is the point: the previous value contradicted the field's own documentation. No test pinned the old number.

The page route's arithmetic is unchanged by the shared helper, verified by diffing against the version before it was extracted.

## Evidence

A document of three lines holding one, one and three words at confidences 0.9, 0.9 and 0.1 gives 0.633 unweighted and 0.42 weighted, so the two formulas cannot agree by accident. That case is red before the change and green after.

## Note

The generated API reference still describes the old per element wording. Regenerating it is handled separately in this project, as with previous changes to the same file.


Goldziher (changes requested):
Verified: `word_count_weighted_mean` (confidence.rs:126) is the single fold both `ocr_aggregate_from_elements` (:98) and `ocr_aggregate_from_pages` (:112) call, both routes agree on scale (`OcrConfidence.recognition` and `PageOcrConfidence.score` are both clamped to `[0,1]`), zero total words returns `None` (:131), and the two call sites of `apply_public_element_policy` are exactly the ones cited in the docstring (extractors/pdf/mod.rs:1823, extraction/image_ocr.rs:144). All 36 `heuristics::confidence` tests pass (`cargo test -p xberg --lib --features heuristics confidence::`). Planted two failures: reverting the `.or_else` fallback at :75-76 turns 3 tests red (page-route fallback tests + the cross-route agreement tests, `None` vs `Some(0.8)`/`Some(0.42)`); replacing the word-count weight with a flat `1_u64` at :100-103 turns 2 tests red with exactly the numbers the PR body claims (`0.6333333` vs `0.42`). Both restored clean against HEAD. `poly lint`/`poly fmt --check` clean.

Blocking:

- No CHANGELOG.md entry. This changes a public metric's value on the embedded-image route (PR body's own "This changes the existing route's number" section) — every comparable fix commit in history bundles a CHANGELOG.md entry in the same commit (e.g. 6fed82b0f0, 8a508fd567, going back through the `[Unreleased]` convention). This PR touches only confidence.rs.

Non-blocking, worth a look:

- Word-count weighting is `text.split_whitespace().count()` (confidence.rs:101). For a CJK/Thai backend result (PaddleOCR supports these), a `Line` element has no spaces to split on, so every line counts as 1 "word" regardless of length — the weighting degrades back to exactly the flat per-line mean this PR sets out to fix. Not a regression (old code was equally flat), but the fix's stated goal ("mean confidence per recognized word... regardless of the element granularity a backend reports") silently doesn't hold for space-delimited-free scripts. Worth a doc caveat or follow-up issue.
- #1682 (`quality_processor.rs`, `aggregate_ocr_confidence`) independently reimplements the identical word-count-weighted fold over `PageContent.ocr_confidence` this PR adds as `ocr_aggregate_from_pages`/`word_count_weighted_mean`, minus the 20-word floor. Once this merges, #1682 should read `ExtractionConfidence::ocr_aggregate` instead of duplicating the fold (its own commit message already flags `ocr_aggregate`'s absence on the page route as the reason it didn't do this) — and the two should agree on whether a tiny word count is “not enough evidence” or not.
- Verified `alef verify --exit-code` is clean at the pinned 0.92.1 (I initially got a false failure at 0.91.6 — version skew, not this PR). The stale generated-binding wording the PR body flags is real (grep hits the old "Mean OCR per-element..." string across 15+ language packages/docs-site pages) but doesn't fail the freshness gate, so no action needed here.

tobocop2:
Added the changelog entry under `[Unreleased]` / `Fixed`, noting that the number the element route already returned changes. That was the only blocking finding.

On the word split for scripts without spaces: correct, and the stated goal does not hold there. A line in a script with no spaces counts as one word however long it is, so the weighting collapses back to a flat per-line mean. Not a regression, but the doc comment overstates what it delivers. That wants a follow-up that either counts characters for those scripts or narrows the claim.

On the overlap with #1682: agreed. Once this lands, that one should read the shared fold rather than keeping its own copy, and the two need to agree on whether a small word count means too little evidence. I will make that change there once merge order is settled.

Goldziher (approved):
Changelog entry resolves the blocking item. Wording nits ("the element route" -> embedded-image route; `from_extraction_result` param doc still names only `ocr_elements`) will be folded into the changelog catch-up on main. Please file the follow-up for the space-free-script word count, and do the #1682 dedupe onto `word_count_weighted_mean` there once this lands.

PR #1697, opened 2026-09-19
fix(ocr): resolve the language probe's data directory the way a job does

### LINKED ISSUE #1671: bug: Tesseract logs a missing tessdata error at backend registration although eng.traineddata exists one directory up
### Description

At OCR backend registration, Tesseract logs:

```
Error opening data file .../xberg-tesseract-<hash>/out/tessdata/eng.traineddata ... Tesseract couldn't load any languages!
```

while `eng.traineddata` exists one directory up, at `.../out/eng.traineddata`. OCR jobs still succeed afterwards, so the message comes from a boot-time probe that uses a different data path than the job path. The error reads as a broken OCR install to an operator.

Expected: the registration probe and the job path resolve the same tessdata directory, or the probe does not log an error for a layout that works.

### Steps to reproduce

1. Build with the bundled tesseract (`full` features) and start a process that registers the OCR backends.
2. Read the log lines after "DeepSeek-OCR backend registered successfully".
3. Run any tesseract OCR job: it succeeds.

xberg v1.2.4, release build, Linux x86_64, 2026-09-18.

tobocop2:
## Related

Closes #1671

## Description

Tesseract logged a missing language-data error when the backend registered, although the language file was present and every real extraction found it.

The registration probe asked Tesseract to initialise with an empty data path. Tesseract then falls back to a compiled-in default that looks one directory deeper than where the file actually sits, so the probe failed and silently returned a hardcoded language list instead.

The probe now resolves the directory the same way a real job does, using the resolver that already exists and that the health check already uses.

## Behaviour

Registration finds the language file where a job finds it. When the file is genuinely missing, the probe still fails and still falls back, so a real problem is still reported: verified with the file absent everywhere the resolver looks and with the network unreachable.

## Test coverage

The test runs under the shipping feature set rather than behind a bundled-language feature that only wasm builds enable. It takes the language bytes from the tesseract crate's build output, which is produced on every build regardless of that feature, so it needs no network and no system install.

It fails on the previous code and passes on this one.

## Note

The generated language bindings and headers still carry the old wording in their documentation. Those files are produced by the code generator and are regenerated separately, as with previous changes to this area.


Goldziher (approved):
Verified: root cause, fix, and regression coverage all check out.

- Confirmed the bug is real and not already fixed on `main` (merge-base = `main` tip `d0a25a2e01`).
- `query_available_languages` now calls `crate::ocr::processor::validation::resolve_tessdata_path` — the same function real jobs use (`execution.rs:1575,2251`), not a re-implementation. Visibility bump `pub(super)` → `pub(crate)` is the minimal change needed to reach it from `tesseract_backend.rs`. Grepped every other `TesseractAPI::init`/`.init(` call site in `crates/xberg/src/ocr/`: `api_pool.rs` and `execution.rs` job paths already pass a resolved path; `tesseract_wasm_backend.rs` uses in-memory `init_5` (no datapath); the two remaining `init("", "eng")` call sites are pre-existing test-only "skip if unavailable" probes (`processor/config.rs`, `processor/execution.rs`), not production code — no other production call site needed this fix.
- Built and ran `cargo test -p xberg --features full --lib ocr::tesseract_backend::tests::` (the shipping CI feature set, per `scripts/ci/rust/run-unit-tests.sh` / `ci-rust.yaml`): 39 passed, 0 failed, including the new `..._under_pdf_ocr` test. Confirmed the fixture path is real, not a silent skip: `target/debug/build/xberg-tesseract-*/out/eng.traineddata` exists from this worktree's own build, and `bundle-tessdata-eng` is not in the `full` closure (only wired for the wasm32 target-dependency table), so only the new pdf/ocr-gated test compiled in.
- Planted-failure probe: reverted the production hunk to `api.init("", "eng")`, rebuilt, reran the new test — RED, exact assertion message: `the language-availability probe must scan the same tessdata directory the real OCR job resolves (XBERG_CACHE_DIR/tessdata here), not silently fall back to the hardcoded language list; got: ["eng", "deu", ...]`, with Tesseract itself logging `Error opening data file .../out/tessdata/eng.traineddata`, reproducing #1671 exactly. Restored via `git show HEAD:<path> > <path>`; `git diff` is clean.
- Conventions: commit types (`fix(ocr)`) conform, no AI attribution, no never-list values, `poly lint` clean on all 4 changed files.

One nit and one thing to resolve before merge, neither blocking the logic:

- Nit: the PR/commit text says the probe now uses "the resolver ... that the health check already uses" — `probe_tessdata` (tesseract_backend.rs:525) actually calls the shared `tessdata_search_dirs` helper directly, not `resolve_tessdata_path` itself (which additionally materializes/downloads). Same search-dir ordering, different function — the doc comment on `tessdata_search_dirs` already describes this sharing accurately; the PR prose overstates it slightly.
- `list_ocr_backend_capabilities`'s doc comment (`plugins/ocr.rs`) is edited and is not `alef(skip)`-marked, so it's Alef-tracked; its old wording is mirrored verbatim into ~24 generated binding/docs files (`packages/*`, `docs-site/src/content/docs/reference/api-*.md`). `alef verify` hashes current source against committed generated output and is wired as a blocking check (`ci-lint.yaml`'s `alef-bindings-freshness`, explicitly "blocking ... the point"), so this job should fail until `task alef:generate` runs and its output is committed — as the PR description itself already flags. CI was still pending at review time; if that job is a required merge check, it needs a regen commit.

Approving on the fix's correctness; the alef regen is a known, disclosed, mechanical follow-up rather than a logic defect.

Issue #1444, opened 2026-08-14
bug: pdf-oxide paints blank pages when Image XObject `/Height` is an indirect reference


anaisbetts:
## Summary

With `force_ocr = true` and the `vlm` OCR backend, Xberg returns:

```text
OCR error: VLM OCR returned no content (model=…)
```

for image-only scanned PDFs whose embedded JPEG XObjects declare `/Height` (and often `/Length`) as **indirect object references**.

Related: #1355 fixed the case where image XObject decode **errors**. Here decode/render **succeeds** with a blank canvas, so the “OCR the raw embedded image” fallback never runs.

## Environment

- `@xberg-io/xberg` **1.0.14** (current npm latest)
- linux x64
- VLM via OpenRouter (`google/gemini-3.7-flash`) — any vision model reproduces; the page image sent upstream is blank

## Root cause

Brother scanner firmware (and possibly other producers) emit Image XObjects like:

```
<</Type/XObject/Subtype/Image/Filter/DCTDecode/BitsPerComponent 8
  /ColorSpace/DeviceRGB/Width 2447/Height 8 0 R/Length 9 0 R>>
stream
…JPEG…
endstream
```

where object `8 0` is the integer `3469`.

pdf-oxide appears not to resolve `/Height N 0 R` when painting the XObject. The page render is an all-white bitmap at the expected aspect ratio. VLM OCR then follows its prompt (“if no text, return empty string”) → Xberg surfaces `VLM OCR returned no content`.

`pypdf` preserves the indirect `/Height` when copying pages, so rewriting the PDF with pypdf does **not** fix it. Inlining `/Height` as a direct integer does.

### A. Direct `/Height` — works

https://github.com/anaisbetts/kreuzakt/raw/ani/fix-pdf-vlm-blank-render-ec8e/fixtures/xberg-indirect-height/direct-height.pdf

### B. Indirect `/Height` — fails

https://github.com/anaisbetts/kreuzakt/raw/ani/fix-pdf-vlm-blank-render-ec8e/fixtures/xberg-indirect-height/indirect-height.pdf

```text
OCR error: VLM OCR returned no content (model=google/gemini-3.7-flash)
```

Intercepting the VLM request shows a blank white PNG (~1225×1736, 100% white pixels). Gemini reasoning: “image is entirely blank”.

### Minimal Sample

```js
import { extract } from "@xberg-io/xberg";

const config = {
  forceOcr: true,
[...]

PR #1698, opened 2026-09-19
fix(ocr): report what the embedded image retry actually did

### LINKED ISSUE #1673: bug: PDF OCR: a failed embedded-image retry reports the first failure, and GLM-OCR pages stay empty with no processing warning
### Description

When a candle backend fails on a rendered page and the embedded-image retry also produces nothing, the page warning reports the first failure (the `backend_options` validation error) instead of the retry's own outcome. With `candle-glm-ocr`, two pages that PaddleOCR-VL and DeepSeek-OCR recover from the same image bytes stay empty, and the only trace of the real cause is a `tracing::warn` "output is empty" in `models/glm_ocr/mod.rs`, which never becomes a processing warning.

Expected: the page warning names the outcome of the last attempt, and an empty model output on the retry is reported as such.

### Steps to reproduce

1. Same setup as the `source_dpi` report, backend `candle-glm-ocr`, a PDF whose last pages carry a large embedded image.
2. Compare the `processing_warnings` for those pages with the log line from the GLM model module.

xberg v1.2.4, release build with `candle-cuda`, 2026-09-18.

tobocop2:
## Related

Closes #1673

## Description

When OCR of a page failed and the embedded image retry also recovered nothing, the warning named the first failure rather than what the retry actually did. Pages came back empty with a message that pointed somewhere else.

Both message sites now report the retry's own outcome: the per-page warning, and the message raised when every page failed. The count in the second one is exact, covering the pages where a retry ran and recovered nothing rather than claiming anything about the rest.

## Tests

Two cases, one per branch.

Every page fails, so the run is a wholesale failure. And one page fails with an empty retry while another page succeeds, so the run completes and only the per-page warning is raised. The second case is a two page document built for it.

Both fail against the previous code, quoting the old message, and pass against this change.

## Base

This sits on the change that lets these backends read a PDF page at all. Without it no page is processed and none of this is reachable, so it should merge after that one.

## Scope

Only the pipeline and its tests are touched. No backend, model configuration or decode loop is changed, so the separate findings about output volume and character duplication in individual backends are unaffected by this.


Goldziher (approved):
Reviewed the two commits on top of #1679's head (fae0c0d1a) — confirmed `git log` shows the base matches, so this must merge after #1679 as the PR body states.

Verified:
- Scope: only `crates/xberg/src/extractors/pdf/ocr/pipeline.rs` and `tests.rs` change in these two commits. `candle_ocr/config.rs` / `core/config/ocr.rs` are untouched here (they belong to #1679).
- Count exactness: `xobject_retry_ran_empty` is declared fresh inside the `for offset in 0..batch_count` loop body and only set when `attempted > 0`, so a page with no embedded images (fallback never meaningfully runs) is correctly excluded from `pages_with_empty_xobject_retry` in both the per-page warning and the wholesale-failure count. Single call site of `recover_page_text_from_image_xobjects`/`page_needs_xobject_fallback` — no sibling location missed.
- Tests: `cargo test -p xberg --features pdf,ocr --lib -- should_report_retry_outcome` → 2 passed, 0 failed.
- Planted failure: reverted pipeline.rs to fae0c0d1a's version, reran the same two tests → both FAILED, quoting the old messages ("OCR of page 1 failed and could not be recovered: ..." / "...first failure: ... VLM OCR returned no content" with no retry wording). Restored the file afterward.
- No `pub` items touched — nothing Alef-tracked, no regen needed.
- `poly lint`/`poly fmt --check` clean on both files; `poly fmt --fix` is a no-op (comments survive the uncomment pass without needing `~keep`).
- Commit types (`fix`, `test`) valid, no AI attribution, no never-list values.

Environment note, not a PR defect: `test_documents` submodule is empty in a fresh checkout, and the fixture `pdf/embedded_images_tables.pdf` the new test's `two_page_pdf_one_page_has_embedded_image()` builds on (via the pre-existing `single_xobject_fixture_bytes()` helper) is content-addressed in that repo's own corpus store rather than committed as a blob. Fetched it directly from `https://storage.googleapis.com/xberg-test-documents/objects/<sha256>` (sha256-verified) to run the tests. This dependency is pre-existing, shared with an earlier #1444 test, not introduced by this PR.

Non-blocking nits:
- `bf5d33355a`'s subject line is 78 chars; CONTRIBUTING.md doesn't enforce a length but most subjects here stay under ~72.
- No CHANGELOG.md entry, though this changes user-visible warning/error text (the original #1673 complaint was exactly about misleading text reaching callers).

Issue #1674, opened 2026-09-18
bug: candle-deepseek-ocr returns a quarter of the page text, invents a table on a signature block, and has no default model id


tobocop2:
### Description

`candle-deepseek-ocr` returns 15 to 25 percent of the text on a page and invents a table on a signature block. On three 150 dpi page images, it returned 378, 656 and 650 characters where tesseract returned 2,140, 2,441 and 2,746 on the same images, and on a page with signatures it emitted an HTML table of repeated digits that is not on the page. GPU utilization stayed at 22 to 30 percent, so the model was not compute-bound.

The backend also has no default model id: the weights had to be staged by hand and passed through `backend_options.model_path`. The Hugging Face repository ships `model-00001-of-000001.safetensors` plus an index file, not `model.safetensors`.

Expected: output on a par with the other candle backends, and either a default model id or a documented staging step.

### Steps to reproduce

1. Stage the DeepSeek-OCR weights and run `candle-deepseek-ocr` on a 150 dpi PNG of a text page with `backend_options.model_path` set.
2. Compare the character count and content with tesseract on the same image.

### Relevant files and configuration

xberg v1.2.4, release build with `candle-cuda`, NVIDIA L4, 2026-09-18. 16.5 GB of GPU memory in use during the run.

PR #1700, opened 2026-09-19
fix(ocr): route every VLM OCR decode through one token selector

tobocop2:
## The problem

The DeepSeek-OCR decode loop stopped after 128 new tokens, hard-coded inside the loop, against 2048 for GLM-OCR and 4096 for PaddleOCR-VL. A dense page came back truncated at roughly 400 to 650 characters, which read as a model that could not see most of the text. The `max_new_tokens` a caller set never reached it either.

Separately, each of the three vision-language backends carried its own decode loop. DeepSeek-OCR and PaddleOCR-VL chose every token with a bare argmax and no repetition control at all. GLM-OCR carried a hand-written penalty, two samplers and a thread-local random generator.

## The change

The limit is a model configuration field. Token selection moves into one shared step built on candle's generation API, which the TrOCR backend in this crate already used. The penalty has two named forms, because there are two and the backends need different ones: presence, which lowers a repeated token's score once, and frequency, which lowers it once per occurrence. DeepSeek-OCR and PaddleOCR-VL use presence. GLM-OCR uses frequency, which is how it has always decoded.

Net effect on tracked source is 113 fewer lines.

## Measured on one A100-SXM4-40GB, F32, three 150 dpi pages

F32 is the dtype every shipping backend passes. The last column is the ratio of distinct 24-character windows to total windows; a looping decoder drives it toward zero.

| backend | p2 | p3 | p4 |
|---|---|---|---|
| GLM-OCR | 2440 / 18.1 s / 0.952 | 1490 / 11.8 s / 0.930 | 2392 / 18.1 s / 0.989 |
| PaddleOCR-VL | 2110 / 6.5 s / 0.990 | 2414 / 6.7 s / 0.984 | 2731 / 7.6 s / 0.973 |
| DeepSeek-OCR | 17870 / 115 s / 0.015 | 5637 / 35.9 s / 0.346 | 22274 / 114.6 s / 0.063 |

## GLM-OCR is unchanged, proven by digest

The middle commit regressed GLM: the shared step started on presence for every backend, and presence is weaker than what GLM had, so GLM began repeating. The third commit restores it.

A reference arm built at the commit that still carries GLM's original hand-written penalty produces output that is byte-identical to this head. An FNV-1a digest of the extracted text matches on all three pages, so this is identical output rather than matching character counts. Reading the two implementations side by side agrees: once per occurrence, no de-duplication, whole history, divide when the score is non-negative and multiply when it is negative, and the same argmax, so ties break the same way.

PaddleOCR-VL is unchanged within 0.3 percent of wall clock, and its figures do not move when DeepSeek's policy changes, so the policy really is per backend.

## What this does not fix

DeepSeek-OCR still repeats to the end of a page at F32, before and after this change. It is tracked as #1701 with measurements.

An earlier draft of this description quoted 1797 characters at a ratio of 0.950 for DeepSeek. That figure is BF16 and does not describe F32. It has been removed rather than restated.

The frequency form was measured as a candidate for DeepSeek and rejected: p2 recovered, p3 went from 5637 characters to 23590 and its ratio fell from 0.346 to 0.117, and p4 did not recover. Two pages of three worse or unfixed is not a fix, and moving the penalty value to chase one page would be tuning rather than a cause.

GLM-OCR also no longer fails on BF16 weights, where its own penalty never cast the dtype.

## Scope

PaddleOCR-VL's limit keeps its value and changes meaning, from prompt plus output to generated tokens, so one field stops carrying two meanings. Measured effect nil.

Refs #1674. Refs #1701.


Goldziher:
Merged this branch into #1702 (the full #1674 fix: cap as a config field per this PR, plus local crops, device dtype, default model id and a repeat guard for the runaway-decode defect noted here). Leaving #1700 open for you to close or keep as you prefer.

PR #1710, opened 2026-09-20
fix(pdf): say when the text-plausibility check could not judge a page

### LINKED ISSUE #1709: bug: the wrong-character text detector abstains on every page of a document without prose
### Deployment

Other

### Version

v1.2.6, release build, features `ocr,pdf-ocr`.

### Description

GH#1696 added language-plausibility detection so that a text layer decoding to real but wrong letters routes to OCR. The detector never runs on a document whose lines are not prose, and the result is indistinguishable from a clean document.

Measured on one 22-page, 1.9 MB PDF with a text layer that decodes to the wrong characters. Forcing OCR returns 73 % more text, which confirms the text layer is wrong:

| run | method | wall | characters | `implausible_text_pages` |
|---|---|---|---|---|
| default per-page gate | native | 1.88 s | 24,020 | `[]` |
| `--force-ocr true` | ocr | 13.08 s | 41,568 | `[]` |
| `--ocr-scanned-pages` | native | 0.46 s | 24,020 | `[]` (`scanned_pages` also `[]`) |

The empty array does not mean the pages were judged plausible. Reproducing the prose gate over the extracted per-page text shows **22 of 22 pages produce zero qualifying prose lines**, so each page yields zero chunks, falls below `PLAUSIBILITY_MIN_PROSE_CHUNKS`, and returns `NotEvaluated`. whatlang is never called on any page of this document.

The cause is the prose gate rather than the language check. A line qualifies as prose only with at least 5 alphabetic words, a digit ratio at or below 0.15, and a formula-symbol ratio at or below 0.10. The document's text is 3,850 alphabetic characters against 3,657 digits and 8,151 punctuation characters over 24,020 total, and its lines are short clauses, headings and table cells. No line clears the bar, so no page is judged.

Those gates are right for what they were built for: GH#1696's calibration reports 0 of 160 native-clean, 0/42 table, 0/24 wide-table and 0/68 formula documents flagged, and a looser gate would push ordinary table-heavy pages into OCR. The gap is that contracts, forms, invoices and agenda packets are exactly the documents people scan, and they are structurally excluded from a check meant to protect them.

Two separable problems:

1. **A document with no prose cannot be checked at all.** A per-page signal that does not depend on sentence-shaped lines would cover this class. A dictionary hit rate over alphabetic tokens, or a character n-gram score against the expected script, both work on a table cell.
2. **Abstention is invisible.** `implausible_text_pages: []` means both "every page was checked and is fine" and "no page could be checked". A caller cannot tell a clean document from an unjudgeable one, and the second deserves a processing warning. This half stands on its own and is much smaller than the first.

Expected: a text layer that decodes to the wrong characters is detected regardless of whether its lines are prose; failing that, a caller can tell that the check did not run.

### Steps to reproduce

1. Take a PDF whose `/ToUnicode` CMap resolves every glyph to a real but wrong letter, and whose pages are tables, forms or short clauses rather than paragraphs. A ROT-shifted mapping applied to an invoice or a contract reproduces it.
2. Extract with the default per-page gate and read `implausible_text_pages`.
3. Extract again with OCR forced and compare the character count.
4. The forced run returns substantially more text, and the default run flags nothing.

### Logs and configuration

xberg v1.2.6, release build with `ocr,pdf-ocr`, tesseract with English data, n2d-standard-32, Ubuntu 24.04, 2026-09-20. Three repetitions per arm. The per-page prose figures come from reproducing the gate's own constants over the extracted per-page text. Counts and ratios only; no document content is reproduced here.

tobocop2:
`implausible_text_pages: []` means two different things and a caller cannot tell them apart: every page was checked and passed, or no page could be checked at all. A document whose lines are not prose falls entirely into the second case and comes back byte-identical to a clean one.

The language check judges a page only when it holds enough prose, where a prose line needs at least five alphabetic words, at most 15 % digits and at most 10 % formula symbols. Contracts, invoices, forms and agenda packets have no lines that qualify, so every page abstains and the result reports nothing.

## The change

A document where no page could be judged now carries a processing warning naming how many pages were examined, and saying that an empty list reports that the check did not run rather than that the text layer is correct.

The scan separates judged, unjudged and already-routed pages, and counts verdicts. A processing warning rather than a new metadata field: `ProcessingWarning` already owns per-document diagnostics, and a new `PdfMetadata` field would break every binding crate until regenerated.

The warning is suppressed when the caller already forced OCR. Telling somebody who forced OCR to force OCR is noise.

## Measured on a 22-page document whose text layer decodes to the wrong letters

| arm | base | head |
|---|---|---|
| default | native, 24,020 chars, `[]`, **0 warnings** | native, 24,020 chars, `[]`, **1 warning** |
| `--force-ocr true` | ocr, 41,568 chars | ocr, 41,568 chars, no warning |
| `--ocr-scanned-pages` | native, 24,020 chars | native, 24,020 chars, **1 warning** |

Extraction behaviour is unchanged. Only the reporting changes.

## Calibration is unchanged

160 native-clean, 42 table, 26 complex-span-table, 24 wide-table, 68 formula and 39 code documents: 0 flagged on base and 0 flagged on head, every cohort evaluated. Reproduces the figures from GH#1696.

## Detecting a wrong text layer without prose is not attempted here

That approach was already measured and rejected: a naive plausibility check over all text scored a legitimate financial table at 0 % reliable and 0.143 mean confidence, as bad as the real defect. The prose gate is the remedy that was adopted, and it is what causes this issue.

Re-proposing it needs positive examples the corpus does not hold. The `wrong_mapping` cohort is absent at the current submodule pin, so true-positive rate cannot be measured at all, only false positives bounded. Shipping a looser detector against that corpus would repeat the earlier mistake in the other direction.

Closes #1709


Goldziher (approved):
Abstention is a real tri-state (PlausibilityVerdict::NotEvaluated / PlausibilityScan.unjudged) that never collapses into plausible/implausible, implausible_text_pages and the scanned_pages merge derive only from scan.implausible so routing is unchanged, and the tests cover all three states with exact assertions. CHANGELOG entry (GH#1709) will be added on main. CI failures are the 9 pre-existing main failures only.

PR #1713, opened 2026-09-20
perf(candle-ocr): read the DeepSeek-OCR expert scores in one transfer

### LINKED ISSUE #1711: bug: DeepSeek-OCR leaves the GPU 73% idle because it decodes one page at a time
### Deployment

Other

### Version

v1.2.6, release build with `candle-cuda,pdf-ocr`.

### Description

`candle-deepseek-ocr` leaves the GPU idle for about three-quarters of a page. The decode loop runs one page at a time, so the card is starved rather than busy, and adding GPUs multiplies idle cards instead of throughput.

Measured on one NVIDIA A100-SXM4-40GB, driver 580.173.02, CUDA 12.9, over the same 6 pages of one scanned document, driven through the PDF path with `--ocr true --force-ocr true`:

| backend | seconds per page | GPU utilisation | GPU peak | characters | shingle ratio |
|---|---|---|---|---|---|
| `candle-paddleocr-vl` | 11.2 | 98 % | 9,410 MiB | 18,808 | 0.945 |
| `candle-glm-ocr` | 15.1 | 96 % | 6,434 MiB | 18,747 | 0.823 |
| `candle-deepseek-ocr` bf16 | 80.0 | **27 %** | 9,034 MiB | 17,937 | 0.967 |
| tesseract (CPU control) | 1.18 | n/a | n/a | 13,975 | 0.941 |

The two peer VLM backends saturate the same card on the same pages, so this is specific to DeepSeek rather than a property of the path or the hardware. Output is healthy: the shingle ratio of 0.967 shows the decoder is not repeating, so the time is spent waiting rather than generating rubbish.

At roughly 2,000 output tokens per page, 80 s is about 40 ms per token on an A100. A decoder at batch size 1 is memory-bandwidth-bound, and the utilisation figure says the bandwidth is not being used either.

The consequence is that this backend cannot be scaled by provisioning. A 731-page document needs about 16 hours, and a second GPU gives two cards each idle 73 % of the time.

Three candidates, none confirmed, all introduced or affected by GH#1674:

1. The crop grid feeds about 903 image tokens per page instead of a single global view, so every decode step carries more context.
2. The shared repeat guard inspects a sliding window of up to 32 tokens across a 128-token history on every generated token.
3. The model is a mixture of experts. At batch size 1 each token loads expert weights that no other token reuses, which is the worst case for bandwidth.

The fix that addresses all three is to decode more than one page at a time. Batching raises arithmetic intensity, amortises weight loading across sequences, and turns a latency-bound loop into a throughput-bound one. It needs per-sequence KV cache handling, padding with an attention mask, and per-sequence stopping, because pages finish at different lengths.

Evidence that the headroom is real: the weights occupy 9,034 MiB of 40,960 MiB, so four copies of the model fit on one card today. Running four processes concurrently against one GPU is a crude way to reach the same effect from outside the product, and it works, which shows the card has capacity the decode loop is not using.

Expected: OCR over many pages keeps the GPU busy, and throughput rises with the hardware it is given.

### Steps to reproduce

1. Extract a multi-page scanned PDF with `--ocr-backend candle-deepseek-ocr` and `{"device":"cuda"}`.
2. Sample `nvidia-smi --query-gpu=utilization.gpu --format=csv` once a second during the run.
3. Compare the mean against the same document through `candle-paddleocr-vl` or `candle-glm-ocr`.

### Logs and configuration

A100-SXM4-40GB, CUDA 12.9, Ubuntu 24.04, v1.2.6 release build, 2026-09-20. Utilisation sampled once a second for the whole run. Counts, ratios and durations only; no document content is reproduced here.

tobocop2:
`candle-deepseek-ocr` left the GPU 73 % idle. The cause is not the decode loop and not batch size: the mixture-of-experts gate copied its score matrix back from the device one scalar at a time, per expert, per row, per MoE layer, per generated token. Reading it in a single transfer removes those copies.

## Result

A100-SXM4-40GB, BF16, engine level, the same 6 pages of one scanned document, each arm its own control:

| arm | seconds per page | mean GPU utilisation | peak memory |
|---|---|---|---|
| before | 62.71 | 29 % | 9,002 MiB |
| after | **10.51** | **63 %** | 9,002 MiB |

Six times faster, no extra memory, and the output is byte-identical: a per-page digest matches the pre-change run on all 6 pages.

## Why this matters beyond the speed

This backend is the most accurate of the local vision models measured here, at 0.234 character error rate against 0.310 for Tesseract and 0.390 for PaddleOCR-VL on the same pages, and it is the only one that recovers tables from a rasterised page. It was priced out of use by throughput alone. At 10.51 seconds a page a 22-page document finishes in about 231 seconds, inside the 600 second extraction ceiling it previously exceeded.

## What is not in this change

Batched decode. It was implemented and measured, and it is not proposed: on top of this fix it adds only 1.25x, and at batch 2 one page's text changed. Identical output is the whole point of this backend, so a speedup that cannot promise it is not worth having. Measurements are on the issue.

The per-page spawn in the PDF OCR pipeline still serialises on the engine mutex. Untouched here.

Closes #1711


tobocop2:
Folded into #1715, which carries this commit unchanged plus the second instance of the same pattern in the rel-pos lookup. One review, one merge, both issues closed.

PR #1715, opened 2026-09-20
perf(candle-ocr): read DeepSeek-OCR device tensors in one transfer instead of per element

### LINKED ISSUE #1711: bug: DeepSeek-OCR leaves the GPU 73% idle because it decodes one page at a time

### LINKED ISSUE #1714: bug: the DeepSeek-OCR rel-pos lookup reads its index one scalar at a time
### Deployment

Other

### Version

v1.2.6, release build with `candle-cuda,pdf-ocr`.

### Description

`index_select_2d` in the DeepSeek-OCR utilities reads its index tensor one scalar at a time, inside a nested loop over the lookup's two dimensions. Each read is a separate device-to-host transfer.

This is the same defect fixed for the mixture-of-experts gate in #1713, in the same file, and it is still present here. In that case removing the scalar reads made the backend six times faster with byte-identical output, so the pattern is worth taking seriously wherever it appears.

The call site is the relative-position lookup in the SAM vision path, and it runs per attention layer, per crop. Since GH#1674 a page is tiled into local crops rather than sent as one global view, so the number of crops per page rose and this loop now runs correspondingly more often.

The fix is one line and the correct idiom already exists in the same file: neighbouring helpers read whole tensors with `to_vec1` and `to_vec2` rather than looping over `to_scalar`. Replacing the nested scalar loop with a single bulk read matches what those helpers already do.

Two things worth checking while making the change, both of which the #1713 review confirmed for the gate and neither of which is guaranteed here:

The bulk read must agree with the scalar loop on element order. A reshape copies when the input is strided, which makes the row-major branch apply, but the caller's tensor layout should be verified rather than assumed. A reordering silently changes which positions are selected, and the model still produces fluent text afterwards, so fluent output is not evidence of correctness.

Output must be byte-identical before and after. Prove it with a per-page digest on both trees, and prove the digest check has power by perturbing the new read and showing it fails.

Expected: the lookup reads its index in one transfer, and OCR keeps the GPU busy rather than waiting on per-element copies.

### Steps to reproduce

1. Extract a scanned PDF with `--ocr-backend candle-deepseek-ocr` and `{"device":"cuda"}`.
2. Sample `nvidia-smi --query-gpu=utilization.gpu --format=csv` once a second.
3. Utilisation stays well below saturation. With the gate fix from #1713 applied it reaches about 52 %, so the remaining idle time is attributable elsewhere, and this loop is the next instance of the same pattern.

### Logs and configuration

A100-SXM4-40GB, CUDA 12.9, Ubuntu 24.04, v1.2.6 release build, 2026-09-20. Found during the independent review of #1713, which measured the gate fix at 61.8 to 9.65 seconds a page and utilisation at 29.6 to 51.8 per cent. Counts, durations and utilisation only; no document content is reproduced here.

tobocop2:
Two device reads in `candle-deepseek-ocr` copied tensors back from the GPU one element at a time. Both are replaced with a single bulk read. Same file, same pattern, same one-line shape.

The mixture-of-experts gate copied its score matrix per expert, per row, per MoE layer, per generated token. That one dominated: it left the GPU 73 % idle and cost six times the throughput.

The relative-position lookup in the SAM vision path read its index inside a nested loop, per attention layer, per crop. Smaller grids, so a smaller return once the first was fixed.

## Result

A100-SXM4-40GB, BF16, engine level, same pages, each arm its own control:

| | seconds per page | mean GPU utilisation |
|---|---|---|
| before both | 62.71 | 29 % |
| gate read only | 9.03 | 52 % |
| both | **8.09** | 51 % |

**7.8x overall.** Peak memory is unchanged at about 9,000 MiB.

Output is byte-identical at every step. Per-page digests match the pre-change run on all pages tested, and each change carries a mutation arm proving the digest check has power: rotating an index by one collapses a page from 1,424 characters to 27 in the gate case and to 685 in the lookup case, with a different digest each time.

Element order holds in both. The reshape copies when the input is strided, so the flattened tensor is contiguous and the bulk read uses the row-major branch. The lookup's caller invokes `contiguous()` immediately before the call, read at the call site, and the bulk read falls back to a strided index anyway.

## Why it matters

This backend is the most accurate local vision OCR measured here: 0.234 character error rate against 0.310 for Tesseract and 0.390 for PaddleOCR-VL on the same pages, and it is the only one that recovers tables from a rasterised page. It was priced out of use by throughput alone. A 22-page scanned document previously exceeded the 600 second extraction ceiling and returned nothing; it now completes in 322 seconds.

## What is not here

Batched decode. It was implemented and measured: on top of these fixes it adds only 1.25x, and at batch 2 one page's text changed. Identical output is the point of this backend, so it is not proposed. Measurements are on #1711.

The PDF OCR pipeline still spawns per page and serialises on the engine mutex. Untouched.

Closes #1711
Closes #1714


Goldziher (approved):
Verified against candle-core 0.11.0 source that to_vec2 preserves row-major order for contiguous and strided tensors, reshape yields a contiguous result, and both call sites already pass contiguous tensors; existing utils tests pass 8/8 on the PR head. CHANGELOG entry (GH#1711/#1714) will be added on main. CI failures are the 9 pre-existing main failures only.

PR #1720, opened 2026-09-20
fix(ocr): resolve the layout model once per process, not once per page

### LINKED ISSUE #1718: bug: the layout model is re-resolved and SHA-256 verified on every page, and its session ignores acceleration
### Deployment

Other

### Version

v1.2.6. Found by static reading during the investigation of #1712; not confirmed by profile.

### Description

Two defects in the layout-detection path, in the same place, both costing work on every page.

**The layout model is re-resolved and re-verified once per page.** `process_paired` constructs a `LayoutModelManager` with no cache and re-resolves the PP-DocLayout-V3 file, then `verify_model_file` runs a SHA-256 over all 131,731,131 bytes of it. That happens outside the engine cache, so a 22-page document hashes about 2.9 GB, and a 731-page document would hash roughly 96 GB. The hashing is CPU-only and serial.

The verification itself is right; doing it per page is not. A model resolved and verified once per engine, or once per process, gives the same guarantee.

**The layout session is built without acceleration.** That backend constructs its ONNX session with acceleration set to none, while `LayoutEngine` passes an acceleration setting through on its own path. So layout detection runs on the CPU even where a GPU is present and idle.

Neither is the cause of the hang in #1712 on the evidence available: 2.9 GB of SHA-256 is seconds, and that investigation stalled for 37 minutes. But the stall's last log line sits immediately before this work, and the two defects are real regardless of what #1712 turns out to be.

The consequence for a deployment is that enabling layout detection costs a per-page CPU tax that scales with page count and does not use the GPU it may have been given.

Expected: the layout model is resolved and verified once, and the session honours the acceleration setting the engine already carries.

### Steps to reproduce

1. Extract a multi-page PDF with layout detection enabled.
2. Observe CPU time spent hashing, and that it recurs per page rather than once.
3. With a GPU present, observe the layout session running on the CPU.

### Logs and configuration

Found while investigating #1712 on an L4. Static reading of the call path, stated as such: no profile was taken, because the box was preempted before a binary existed. Worth a profile before sizing the fix, but the per-page re-verification is visible in the code without one.

tobocop2:
The layout model was resolved once per page. Resolution now happens once per process, cached by directory, which is the shape the layout engine's sibling models already use.

## Measured, not inferred

A counter on the real 131,731,131-byte model: 22 pages produced 22 verifications before, 1 after.

| arm | 22 pages |
|---|---|
| before | 82.71 s |
| after | 6.24 s |

Extrapolated to a 731-page document, that is about 44 minutes of work removed.

## The cost is not where the issue said it was

The issue describes this as a SHA-256 tax and dismissed it as "seconds". One resolution costs 3.64 s, of which the hash is 0.10 s warm. **The other 97 % is Hugging Face path resolution**, which the per-page loop repeated along with it. Verification was never the expensive part.

That matters beyond the timing. GH#1712 reports a run that stalls with the GPU idle and no log output, whose last line sits immediately before this work. That issue was dismissed as unrelated on the arithmetic above, and the arithmetic was wrong. A path resolution repeated once per page is 22 opportunities to block where there should be one, so this change reduces the exposure even if it is not the whole cause.

## The second half of the issue is not a defect

The issue claims the ONNX session ignores acceleration and forces CPU. That is half right, and the half that is wrong matters less than the half that is right.

Nothing hard-codes CPU: passing no acceleration resolves to `Auto`. But the CUDA registration sits behind a cuda feature that is opt-in, absent from the default set and not implied by the OCR backend features. **A stock build therefore compiles the non-CUDA arm and does run layout detection on the CPU.** The mechanism is a feature gate rather than a hard-coded provider, and the observable behaviour on a default build is the one the issue described.

What is dropped is an explicit acceleration setting, a device id or a pinned provider, which the layout engine honours on its own path. Restoring it would move where layout detection runs, and layout-only jobs are routed to CPU workers deliberately, benchmarked at no measurable overhead for the large majority of documents. That is a documentation gap rather than a bug, so nothing here changes it.

Two loose ends found while reading and left alone: the layout GPU task type is never constructed, and the layout pool's key carries a device preference the session does not read.

Closes #1718


Goldziher (approved):
Cache key matches how LayoutModelManager::new(None) resolves its directory, EngineCache::get_or_try_init stores nothing on failure, and the new tests drive the real static and would read 22 without the change. CHANGELOG entry (GH#1718) will be added on main. CI failures are the 9 pre-existing main failures only.

PR #1722, opened 2026-09-20
fix(ocr): report when auto device selection falls back to the CPU

### LINKED ISSUE #1712: bug: candle-glm-ocr spins on the CPU with the GPU idle and never finishes a 22-page document
### Deployment

Other

### Version

v1.2.6, release build with `candle-cuda,pdf-ocr`, on one NVIDIA A100-SXM4-40GB, CUDA 12.9.

### Description

`candle-glm-ocr` stops making progress on a 22-page PDF and spins on the CPU until it is killed. The GPU does no work while this happens.

Killed manually after 2,492 s. State at that point:

| signal | value |
|---|---|
| GPU utilisation | 0 %, sampled over 60 s |
| CPU | 460 % |
| runnable threads | 5 |
| voluntary context switches | 38 |
| last log line | `xberg::model_download` resolving `pp_doclayout_v3/model.onnx` |
| time since last log line | 37 minutes |

`candle-paddleocr-vl` completes the same document in 191 s on the same machine in the same session, so the document and the path are not at fault.

Two things make this worse than a slow run. The process burns four and a half cores while doing nothing, so it does not look idle to a scheduler or an operator. And 38 voluntary context switches against 460 % CPU is the signature of a spin rather than a wait: a download blocked on the network would be sleeping, not consuming CPU.

The last log line points at layout-model resolution rather than at OCR. The backend gains `layout-detection` through its feature set, so the layout model is fetched on the same path, and the process appears to get stuck there rather than in recognition.

Document shape that reproduces it: 22 pages, 1.9 MB, a native text layer whose character mapping resolves to the wrong letters, 57 table regions detected by the PDF pipeline. Run with `--ocr true --force-ocr true --ocr-backend candle-glm-ocr --ocr-backend-options '{"device":"cuda"}'`.

Expected: the backend either completes, or fails with an error naming what it was waiting for. A run that holds four cores at zero GPU for 37 minutes with no log output has no way to be diagnosed from outside.

### Steps to reproduce

1. Take a table-dense PDF of about 20 pages whose text layer decodes to the wrong characters.
2. Extract it with `--ocr-backend candle-glm-ocr` and `{"device":"cuda"}`, forcing OCR.
3. Sample `nvidia-smi --query-gpu=utilization.gpu --format=csv` and the process CPU once a second.
4. GPU utilisation sits at zero while CPU stays high, and no further log lines appear.

### Logs and configuration

Ubuntu 24.04, v1.2.6 release build, 2026-09-20. GPU utilisation and memory sampled once a second for the whole run. Counts, durations and process statistics only; no document content is reproduced here.

tobocop2:
When device selection is set to automatic and the accelerator cannot be initialised, the engine fell back to the CPU and reported success. A caller who expected a GPU got a slow CPU run with nothing in the logs to say why.

## Reporting the fallback

Automatic selection now logs when it falls back. It still falls back rather than refusing: a caller who explicitly asks for a specific accelerator already gets a real error when initialisation fails, and turning the automatic mode into a refusal would break CPU-only deployments that rely on it working.

## Resolving the device once, not once per page

Three of the four OCR backends resolve the device inside engine initialisation, behind a locked pool, so the new message appears once. The DeepSeek backend resolved it inside the per-image path instead, which would have produced one identical warning per page. On a 731-page document that is 731 lines.

Device resolution now goes through the same cache the engine pool in that file already uses, at its own key. The pool key is untouched, so the earlier attempt that split one engine into two identical resident copies cannot return.

## A failed probe is retried, not remembered

The first version of that cache remembered failures as firmly as successes, so one transient accelerator failure pinned the process to the CPU for good. The shared cache keeps nothing when initialisation fails, which is what the other three backends already do and what this backend did before the cache existed. Holding the error itself rather than its text also keeps the underlying cause attached to the reported failure.

Automatic selection is the deliberate exception: its CPU fallback is a success, so it is cached, and a host with no accelerator probes once rather than once per page.

## What this does not fix

This does not close #1712. That report is a spin under an explicit `cuda` device, with the last log line at layout-model resolution; #1720 addressed that path and has merged. This change is about the automatic mode's silent fallback, which #1712 surfaced but is not the failure it describes.


Goldziher (changes requested):
This does not close #1712 as claimed.

#1712 reports `candle-glm-ocr` spinning at 0% GPU / high CPU with the last log line at ONNX layout-model resolution (`pp_doclayout_v3/model.onnx`), reproduced with an explicit `"device":"cuda"`, not `"auto"`. This PR only changes `DevicePreference::Auto`'s fallback reporting in `device.rs` and DeepSeek-OCR's per-page device re-resolution. `glm_ocr_backend.rs` is untouched, and explicit device preferences already propagated errors before this change. Merging as-is auto-closes #1712 without touching the reported spin. Either explain the connection or drop the "Closes #1712" link (#1720 is the one that actually addresses the layout-model resolution path).

Behavior change not in the body: the new `DeviceCache` is `OnceLock::get_or_init`, so a failed accelerator init is cached as permanently as a success. The peer `EngineCache::get_or_try_init` used by GLM/PaddleOCR-VL/TrOCR stores nothing on `Err` (see `get_or_try_init_keeps_nothing_when_init_fails`), and DeepSeek previously re-attempted selection every page, so a transient CUDA init failure self-healed. Now the first failure pins the process to CPU for good. Either don't memoize the `Err` branch or state the trade-off explicitly.

Minor: `DeviceCache::resolve` stringifies the `candle_core::Error` before caching, so the resulting `XbergError::Ocr` always has `source: None` where the previous inline call kept the source chain. Fine if intentional (the error isn't `Clone`), but worth a one-line `~keep` comment.

The Rust/E2E CI failures on this PR are pre-existing on `main` and unrelated.

tobocop2:
Thanks — all three land. Addressed in eea9f961a.

**`Closes #1712` dropped.** You are right that this does not close it: #1712 is a spin under an explicit `cuda` preference, and #1720 is the change that addressed that path. The body now says so and names #1720 instead of claiming the close.

**The cached `Err` is gone, and the hand-rolled cache with it.** Rather than special-casing the error branch, device resolution now goes through `EngineCache`, the same cache the engine pool in that file already uses. It keeps nothing when initialisation fails, so a transient failure is retried on the next page instead of pinning the process — the behaviour the peers have and the one this backend had before the cache existed. That removes the second, weaker memoization primitive instead of adding a guard to it. The engine pool key is untouched, so the split this branch measured and rejected earlier cannot come back.

Two regressions cover it: one drives a probe that fails once and then succeeds and asserts the second call recovers, one asserts the source chain survives. Both were confirmed red against the previous memoize-the-error semantics before the fix went in.

**The source chain is restored rather than commented.** Holding the `XbergError` in the cache instead of a `String` means nothing has to be stringified to be cacheable, so the candle error stays attached as the source. No `~keep` needed.

One thing worth stating explicitly rather than leaving implied: automatic selection is still cached on its CPU fallback, because that fallback is a success, not an error. A host whose accelerator never appears therefore probes once, not once per page, which is the repeated-log problem this cache exists to solve. Only genuine initialisation failures are retried.

On CI: agreed the Rust and E2E failures here are pre-existing on `main` and unrelated.

tobocop2:
## Fixed

**Failures were cached permanently.** The device cache now uses the same engine cache as the peer backends, which keeps nothing when the initialiser fails. A transient accelerator fault no longer pins the process to the CPU.

**Lost error source.** It holds the error rather than its string, so the reported failure keeps its cause.

**Wrong closing link.** `Closes #1712` is gone. The description states this does not close it and points at #1720.

## Gate

Merged tree: fmt and clippy clean, 79 tests on the candle crate, 128 on the OCR module.

The changes-requested state stands only because the review predates these commits.

Goldziher (approved):
All three items from the earlier review are addressed in eea9f961a: device resolution goes through `EngineCache::get_or_try_init` (nothing kept on `Err`; the fail-once-then-succeed regression test covers it), the candle error survives as `source`, and the #1712 close is dropped in favour of a pointer to #1720. Verified the 26 deepseek backend tests and the 3 device tests pass, and that the static's `Sync` bound holds with `candle-metal` on. One non-blocking nit: the `auto_fallback` doc in `device.rs:51` still cites GH#1712 as the motivating issue where GH#1722 would match the body.

Issue #1716, opened 2026-09-20
bug: PDF extraction uses one core, so the thread budget changes nothing


tobocop2:
### Deployment

Other

### Version

v1.2.6, release build, measured two ways: `--no-default-features --features ocr,pdf-ocr` and with default features plus `pdf-ocr`. Both behave identically.

### Description

Extracting a large PDF uses one core. On a 32-core machine, where full use would read 3,200 %, the process averages 96 to 103 % and peaks at 141 %. Raising the thread budget changes nothing because nothing is waiting on threads.

Measured on one 731-page, 95.3 MB mixed native-and-scanned PDF, n2d-standard-32, Ubuntu 24.04, idle machine, CPU sampled once a second across the process tree:

| build | arm | thread budget | wall | CPU mean | CPU peak |
|---|---|---|---|---|---|
| default features | per-page gate | 4 | 55.90 s | 100 % | 139 % |
| default features | per-page gate | 16 | 55.19 s | 103 % | 141 % |
| default features | per-page gate | 32 | 55.12 s | 103 % | 141 % |
| default features | native only | 32 | 24.66 s | 96 % | 100 % |
| lean | per-page gate | 32 | 55.05 s | 103 % | 141 % |
| lean | native only | 32 | 24.43 s | 96 % | 100 % |

Character counts and serialized sizes are identical across every arm, so these are the same extraction done at different thread budgets.

Two separate observations, and the second is the actionable one.

**Native text extraction is entirely single-threaded.** 96 % mean and 100 % peak is one core, exactly. 24.6 seconds of the 55 second total is spent there.

**OCR is barely parallel.** The gate sends 124 of the 731 pages to OCR, which accounts for the remaining 30.5 seconds. Across that phase the process reaches 141 % at peak, so roughly one and a half cores on a machine with 32. At 32-way parallelism those 124 pages would cost about one second instead of thirty.

GH#1666 reported that OCR throughput does not scale with the thread budget and was closed as fixed in this release. On this measurement it still does not scale: 4 threads and 32 threads differ by 1.4 %, which is inside the run-to-run spread.

One candidate mechanism, not confirmed: the PDF OCR pipeline spawns per page and then serialises on the engine mutex, so pages queue for a single engine regardless of the configured budget (`extractors/pdf/ocr/pipeline.rs`, around the two spawn sites). That would produce exactly this shape. It is worth checking before any other hypothesis, and it is cheap to check.

### Why this is P0

[...]

PR #1724, opened 2026-09-21
fix(pdf): size the per-page OCR batch by free memory, not the content limit

tobocop2:
The per-page OCR route sized its render batch from the configured limit on returned content size. That value bounds the text an extraction returns and says nothing about the rasters a batch holds while it works. On an ordinary 150 DPI document the arithmetic pinned the batch at four pages, so rendering, encoding and recognition all ran four wide whatever the thread budget said.

## The fix

The batch is now bounded by what a page actually costs. The per-page render-and-encode cost is derived from the page box and the effective render resolution, and the batch is sized against free memory less the document and a reserve. A fixed estimate remains only where the caller has no page box to measure, and that call site names it.

## Reading free memory no longer blocks the executor

That reading blocks: on Linux it opens `/proc/meminfo` and the cgroup files, and on macOS it spawns a `sysctl` child process and waits for it. Both call sites are async, and this change puts one of them on the primary mixed-OCR route, so the read now goes to the blocking pool instead of a runtime worker. A read that cannot be joined reports unknown, the same answer a failed query gives, which leaves the configured batch size alone; a measured zero narrows the batch to one page.

## Measured

Cold, with the recognition cache cleared before every run and entries written asserted each time, on a 731-page document at a thread budget of 32: 167.2 s to 144.5 s, a 13.6 percent reduction. The base does not scale across the same range. Peak resident memory rises from 2,234 MB to 3,583 MB, which is the cost of holding more rasters in flight. Output digest is identical across all twenty runs.

## What this does not do

It does not make extraction use the machine. Peak CPU reaches about two cores of thirty-two, so the work is still close to serial and a larger machine is not yet a remedy. Recognition concurrency and the serial native parse are being looked at separately.

Issue #1666 stays open after this merges. It is about wall time not scaling with the thread budget across 4, 8 and 16, and one measured point at a budget of 32 does not establish that its symptom is gone; the section above says as much. The batch-width defect is real and the in-flight probe test pins it, but the scaling #1666 describes has still to be measured.

The measurement in #1716 was taken on a warm cache and is corrected in a comment there.


Goldziher (changes requested):
Two things to address before merge:

1. `extract_mixed_ocr_native` now calls `adapt_batch_size_to_memory`, which reaches `get_available_memory()` — a blocking `/proc/meminfo` read on Linux and a synchronous `sysctl` subprocess spawn on macOS — directly inside the `async fn`, with no `spawn_blocking`. This is now on the primary mixed-OCR route (three call sites in `mod.rs`). Wrap it in `spawn_blocking` rather than widening the pattern that already exists in `extract_with_ocr_for_page`.

2. "Closes #1666" does not match the PR's own "What this does not do" section. #1666 is specifically about wall time not scaling with the thread budget (its table is 4/8/16). This PR measures one point at budget=32 (167.2 s → 144.5 s) and says peak CPU stays at ~2/32 cores, "still close to serial". The batch-width fix is real and the in-flight probe test proves it, but it does not establish that #1666's symptom is resolved. Either drop the auto-close or re-scope the claim.

Minor:
- No `CHANGELOG.md` entry; this changes batch sizing and raises peak RSS (2.45 → 3.84 GB per the body), which is user-facing.
- `mixed_ocr_keeps_as_many_pages_in_flight_as_the_thread_budget` asserts exact equality against real host free memory (~700 MB needed) instead of going through the `adapt_batch_size_to_memory_inner` seam this PR added for determinism — a flake risk on constrained runners.

Heads up: #1733 edits the `ConcurrencyConfig` literal in the same test function this PR's doc-comment update touches (`mod.rs` ~5185-5199); whichever lands second needs a small manual rebase.

The Rust/E2E CI failures on this PR (jupyter/ODT image tests, tessdata probe, model registry count, node plugin_api, kotlin `supportsLanguage`) are pre-existing on `main` and not caused by this change.

tobocop2:
All four addressed in c26d21490.

**The blocking read is off the executor.** Rather than wrapping the new call site and leaving the old one, the `spawn_blocking` sits inside the batch-sizing helper itself, so both call sites are covered by one change and a third caller cannot reintroduce the pattern. A read that cannot be joined reports zero, which is the same "could not tell" answer a failed query already gives, so the configured batch size stands. The non-tokio and wasm builds call through directly, matching the gating the rest of the file uses.

**`Closes #1666` dropped.** You are right that one point at budget=32 does not establish that #1666's scaling symptom is resolved, and the body already said the route stays close to serial. The body now states plainly that the issue remains open until the scaling it describes is shown.

**Changelog entry added**, naming both the batch-width change and the peak RSS rise.

**The in-flight test no longer reads host memory.** It goes through the same injection seam the arithmetic tests use, pinned to a figure well above the bound, restored on drop, and the test was already `#[serial]`. The page box stays Letter, so the test is still red against the original four-page ceiling — shrinking the page to dodge the memory bound would have made it green on the unfixed tree, which is why I pinned the memory instead.

On the two CI notes: agreed the Rust and E2E failures here are pre-existing on `main`. Locally 312 of this module's tests pass and 7 fail only because the `test_documents` submodule is not initialised in my checkout, which is environmental.

On #1733 touching the same test function: understood, whichever lands second takes the rebase. That branch also adds a second free-memory reader, and I have raised there that it should reuse this helper rather than diverge from it — worth settling before either merges, since a third reader is how the three drift apart.

tobocop2:
## Fixed

**Blocking read on the async executor.** The read moves to the blocking pool inside the sizing helper, so both call sites are covered and a third cannot reintroduce it.

**Host-dependent test.** The in-flight test pins free memory through the seam this PR adds, rather than needing ~700 MB spare on the runner.

## Gate

Merged tree: fmt and clippy clean, 291 OCR tests, 118 pdf tests.

## Worth knowing

An earlier arm reported zero failures while passing nothing. `--lib` builds default features, which exclude `pdf` and `ocr`, so a filter over them matches nothing and exits 0. That arm needs `--features full` and the fetched corpus.

Goldziher (changes requested):
The `spawn_blocking` move and the pinned-memory test are both done right (both call sites await the helper; pinning 600 MB instead of 8 GiB makes the in-flight test fail with `saw 4`, the width the unfixed tree produces), and the 5 new/changed tests pass under `--features pdf,ocr`. Three things before this can merge.

**Regression on the pipeline + `include_page_rasters` branch.** `pipeline.rs:618-625` still validates the whole batch against `max_content_size`, and the content-limit pre-shrink that kept that sum small is gone. With `ocr.pipeline` (or `vlm_fallback`, or an active layout pass) and `images.include_page_rasters = true`, any thread budget ≥ 2 on Letter pages now fails hard: 2 × (25.2 MB + 33.9 MB) + 25.2 MB = 143 MB > 100 MiB at the 300 DPI that `images: Some(..)` implies. Reproduced with a 16-page Letter PDF, a one-stage pipeline over a no-op custom backend, `include_page_rasters: true`, `max_threads: 8`, memory pinned to 8 GiB: `Validation error: Image dimensions 2550x3300 require 565902152 live image-processing bytes, exceeding ... (104857600 bytes)`. The same probe passes on main (batch 1). Apply the per-page loop you added at :660-662 here as well, and add that case as a regression test.

**#1666 is still auto-closed.** `closingIssuesReferences` on this PR lists #1666: "This no longer auto-closes #1666" contains `closes #1666`, which GitHub matches. Reword so no closing keyword is adjacent to the ref, and check the Development sidebar.

**Measured zero and unknown are the same value.** `get_available_memory()` returns 0 both when it cannot read and when the cgroup has no headroom (`memory.current` includes page cache, so that is routine in a warm container), and `adapt_batch_size_to_memory_inner` turns 0 into the full configured width — 1 byte of headroom gives a 1-page batch, 0 bytes gives 8. Main also had the content-limit cap on this route; now the memory bound is the only one. Return `Option<usize>` (None = unknown → configured) and treat `Some(0)` as a one-page batch.

Smaller:
- `TEST_AVAILABLE_MEMORY` (:2610) is `#[cfg(test)]` only; `cargo clippy -p xberg --no-default-features --features "pdf,tokio-runtime" --tests -- -D warnings` fails on it as dead. Gate it `all(test, any(feature = "ocr", feature = "ocr-pipeline"))`.
- The in-flight probe backend (`tests.rs:3749`) is never unregistered; every other route test here unregisters.
- CHANGELOG/PR body say peak RSS 2.45 → 3.84 GB; commit f685f70e3d says 2,234 → 3,583 MB. Quote one run. The executor sentence is an implementation detail and can go.

Merge order with #1733: this merges clean against it; its `ConcurrencyConfig` field will need `..Default::default()` in the new test's literal (`tests.rs:3773-3775`) once both land.

Goldziher (approved):
All three items from the last round are in: per-page validation on the raster branch with a regression test that fails when the batch is summed again (checked), `Option` free-memory reading with `Some(0)` narrowing to one page and `None` leaving the configured width, and #1666 no longer in the closing references. 9 targeted tests pass under `--features pdf,ocr`; clippy `-D warnings` clean on both the `pdf,ocr` and `pdf,tokio-runtime` legs. I edited one stale body sentence (a failed join now reports unknown, not zero).

Issue #1725, opened 2026-09-21
bug: native page text depends on font load order, which blocks any page-parallel parse


tobocop2:
## Problem

Native page text depends on the order in which pages are loaded. Font sets and TrueType character maps are shared across pages, and where two pages carry conflicting subsets of the same font, whichever loaded first wins. The engine documents this behaviour in its own source.

Sequentially this is invisible, because the order is always the same. It stops being invisible the moment any page-level work runs concurrently.

## Measurement

A page-parallel read of the same 731-page document, four interleaved repetitions on an idle 32-core machine, page cache dropped between runs:

    arm             wall       peak CPU    output
    sequential      26.60 s    105.5 %     962,806 B, one digest across four runs
    page-parallel   25.67 s    315.7 %     952,808 to 960,483 B, four digests across four runs

Every parallel run produced a different result. Two threads are enough to diverge, and pinning to a single thread reproduces the sequential output byte for byte. The variation is up to about 10 KB of text on a 962 KB extraction.

## Why this matters beyond one experiment

It blocks page-level parallelism in the parse path generally. Extraction currently uses about one core of thirty-two on this document, and the obvious remedy is to read pages concurrently. That remedy is unavailable while the result depends on load order, so this is a prerequisite rather than an optimisation.

It is also a correctness question on its own terms. Two callers extracting the same document should not be able to get different text, and today concurrency is the only thing preventing that.

## What would fix it

Resolve font and character-map conflicts deterministically rather than by arrival order, so the winner is a property of the document rather than of the schedule. A page-parallel read should then be byte-identical to a sequential one, which is the test worth writing first.

## Version

v1.2.6.

Issue #1723, opened 2026-09-20
bug: native extraction is serial, and the cost is in scan detection and provenance routing rather than the text loop


tobocop2:
## Problem

Native text extraction runs on one core. On a 731-page mixed native and scanned PDF of 95.3 MB it takes 24.6 seconds at 100 % of a single core on a 32-core machine, with no stage of it spreading across the other 31.

This was previously hidden behind the OCR path. Once the per-page OCR route stops clamping its batch width (#1666), native extraction becomes roughly **half the total wall clock** and the binding floor on the document: no amount of additional OCR concurrency moves it.

## Why this is not #1666

#1666 is a batching defect: a route computed its width from the wrong configured limit, so every stage below it ran four wide. Removing that clamp lets recognition widen, and measured peak CPU rises from 140 % to 197 %.

This is a different shape. The parse is page-parallel work being done in sequence. There is no width to unclamp, because nothing is batching in the first place. Fixing #1666 does not touch it, and the 197 % ceiling above is partly this: one core of the two is the serial parse.

## Measurement

    stage                     wall      CPU
    native extraction         24.6 s    100 % of 3200 %
    whole document, head      48.2 s    197 % peak

Machine: 32 vCPU, idle, run verified idle beforehand. Three repetitions, content digest identical across all of them.

## What would fix it

Pages parse independently. The work is per-page structure extraction over a shared document handle, which is the shape that parallelises without changing output, provided the page order of the result is preserved.

Raising the thread budget is not the fix and does not change this number, which is the same thing #1666 records for the OCR route: the budget is not reaching the code that would use it.

## Scope note

This should be measured cold as well as warm. A related measurement in #1716 was taken on a warm result cache and attributed time to a stage that was not running. The figure above is a native-only extraction with no OCR requested, so no result cache applies to it, but anyone extending this measurement to whole-document runs should clear the cache between repetitions and say so.

## Version

v1.2.6, and present on the branch that fixes #1666.

tobocop2:
## My framing in this issue was wrong. Re-scoping it, with the measurements.

The 24.6 seconds is the whole native extraction path, not the per-page text loop I pointed at. Stage timings from a 731-page document:

    open                                        1.5 s
    per-page text                               2.7 s
    scan detection and provenance routing      14.2 s
    table extraction                            6.8 s

[...]

PR #1726, opened 2026-09-21
refactor(pdf): read the native page texts through one ordered helper

tobocop2:
Both native text paths built their page list separately, and one of them reserved capacity through a heuristic that guessed five pages. They now share one helper that reserves exactly what the page range needs and returns the pages in order. The page separator was a duplicated literal in two places; it is named and shared.

Output is byte-identical, but allocation is not. The fast path now forces every page's cleaned text through an owned copy before concatenation, so a page that was previously borrowed takes one extra allocation. That is the cost of reserving exactly the right capacity instead of guessing. A test pins page order, the per-page byte counts and the boundary slices, so a later change to either path fails loudly rather than drifting.

## What this does not do

It does not make the parse concurrent, and the investigation behind it found that it cannot be made concurrent yet. A page-parallel read returns different text on every run, because page text depends on the order fonts are loaded. That is filed as #1725 and has to land first.

The investigation also corrected the premise of #1723. On a 731-page document the per-page text loop costs 2.7 seconds of a 26.6 second native extraction; scan detection and provenance routing cost 14.2 seconds. The remaining work on serial extraction belongs there, not here.

## Provenance of the measurements

The timings above were taken on a machine that has since been deleted. This branch was rebuilt from the patches kept off that machine, and it compiles and formats clean, but the full gate has not been re-run on the rebuilt tree.

Refs #1723.


Goldziher:
Draft review — no blockers. Verified against GH#1723: this does not introduce parallelism (no rayon/spawn_blocking/join_all in the diff), page order is preserved deterministically in both paths via sequential `(0..page_count).map(..).collect()`, the per-page abort-on-first-error semantics are unchanged, and the `&mut NativeDocument` → `&NativeDocument` change is sound since `PdfDocument`'s caches are all `Mutex`-wrapped. The exact-capacity reservation also incidentally fixes the old 5-page sampling heuristic's wrong per-separator byte count.

Two minor notes:
- The fast path now forces every page's cleaned `Cow<str>` through `into_owned()` before concatenation, so already-clean pages that previously went `Cow::Borrowed` straight into `content` now take an extra allocation. Reasonable trade for the exact reservation, but "byte-identical output" doesn't capture the allocation change; one line in the body.
- `extraction.rs`'s doc comment above `join_pages_with_boundaries` still says `otherwise "\n\n" separators` in prose while the code now uses `PAGE_SEPARATOR`.

Goldziher (approved):
Traced both paths: output bytes, page order, capacity arithmetic (`(n-1) * PAGE_SEPARATOR.len()`, marker lengths on the tracking path) and abort-on-first-error are unchanged, and `native_page_text_preserves_page_order` fails when the page range is reversed. 1206 `pdf::` and 65 `extractors::pdf::` tests pass with `--features pdf`. I pushed d45611eddf to the branch for three stale comments: the `~keep` rationale cited GH#1723 for the font-load-order dependence (that is #1725), `build_paged_text_pdf`'s doc referred to a concurrency test that is not in this PR, and `join_pages_with_boundaries` still said `"\n\n"` in prose. For the record: the fast path now holds every page's `String` before joining (transient ~2x text size, one extra allocation per already-clean page) — fine, just not captured by "byte-identical".

PR #1729, opened 2026-09-21
ci(rust): cover xberg-candle-ocr in the workspace test and lint jobs

### LINKED ISSUE #1728: ci: the vision-OCR crate is excluded from the workspace test and lint jobs, so nothing there is covered
## Problem

The workspace-wide test and lint jobs exclude `xberg-candle-ocr`, so nothing in that crate is covered by CI. Two changes currently in review land in it, and neither was tested by any automated check before review caught up with them.

An integration build across five branches ran that crate explicitly and it passed, 26 tests. CI would not have run them.

## Why it matters beyond one crate

The crate holds the local vision-model backends. Recent work there included a device-selection change that, in an earlier revision, silently loaded a second copy of a multi-gigabyte model, and a per-element device loop that cost six times throughput. Both were found by review rather than by a test run, because no test run covered them.

An exclusion that was cheap when the crate was small is now hiding a meaningful surface.

## What to check before changing anything

The exclusion probably exists for a reason. Likely candidates are build time, a model download the runner cannot do, or a dependency that needs a GPU or a large native toolchain. Establish which before proposing a fix, because the right answer may be to run a subset rather than the whole crate.

If the blocker is model downloads, the tests that do not need a model can still run. If it is build time, that is a scheduling question rather than a coverage question.

## Suggested shape

Include the crate in the workspace jobs, or add a dedicated job for the parts that can run without special hardware, so a change there fails before a human reads it. State in the workflow why anything that remains excluded is excluded, so the next person does not have to guess.

## Version

Observed on the current default branch.

tobocop2:
The workspace test leg excluded this crate entirely, so none of its per-model code was compiled or run by any pull-request job, and the workspace lint only ever saw the feature-independent baseline.

## Why the exclusion existed, and the part that still holds

The crate's device-accelerator features are mutually platform-hostile under `--all-features`: the Metal feature pulls an Apple-only dependency that fails to build on Linux, and the CUDA feature needs a compiler that is absent on macOS. That was reproduced rather than taken from the history, and it remains true. The crate still cannot be built with every feature at once on any single runner.

## What did not hold

Excluding the whole crate was broader than that problem required. Its default feature set is empty, so the four per-model backends were never built anywhere in push or pull-request CI. A GPU workflow does exercise them, but it runs only on manual dispatch and goes through a different crate with real model downloads, so it never gates a change.

## The change

A curated test leg and lint leg that enable every per-model feature and no device backend, following the pattern this script already uses for another crate with the same accelerator conflict. Both skip on 64-bit Arm Linux, where a transitive dependency needs a target feature that runner does not have, matching the existing gating for the same reason.

Verified on a 32-core Linux host: the new test leg passes 76 tests, two ignored because they need real model weights, and the matching lint invocation is clean.

Closes #1728.


Goldziher (approved):
Feature list matches xberg-candle-ocr's four per-model features with no device backend, the aarch64 skip targets the only Linux/aarch64 matrix leg, the clippy leg runs -D warnings, and the new test leg feeds the existing zero-pass guard. The pending Rust matrix jobs will hit the 9 pre-existing main failures (jupyter/ODT images, tessdata probe, registry count), which are unrelated to this change.

PR #1730, opened 2026-09-21
fix(layout): charge each detection batch against the budget, not the whole document

### LINKED ISSUE #1721: bug: layout detection retains every page raster against one budget, so it fails past ~12 pages and reports success
## Problem

Layout detection accumulates every page raster for a document in one retained collection, and charges the running total against a single content-size budget. The budget is never released between pages, so a long document exhausts it and layout detection fails — while extraction continues and the result is reported as successful.

A caller who requested layout hints gets a result with none, a success status, and a warning that a client library may not surface.

## The failure looks like a page-size limit and is not

The warning names the page's pixel dimensions, which reads as though the page is too large. Measured at the default 150 dpi render resolution:

    one US Letter page   30,887,250 bytes
    one A4 page          31,100,880 bytes

Both clear the 100 MiB default comfortably. A single page of ordinary size never trips this.

What trips it is the sum. The figure in a real warning decomposes exactly, with zero remainder:

    16 × (1275 × 1650 × 3)   = 100,980,000   sixteen retained RGB8 page rasters
    1280 × 1280 × (12 + 3)   =  24,576,000   the model workspace
                               -----------
                               125,556,000

So the failure is a function of document length. At the 100 MiB default it appears after about twelve retained pages, at any standard page size. A document of twelve pages or fewer will not reproduce it.

## Why a larger budget is not the fix

Raising the limit moves the cliff rather than removing it — roughly 81 pages at 512 MiB instead of 12 at 100 MiB. The shape of the defect is accumulation, so the fix is to release each raster once it has been consumed, or to charge the budget per batch rather than per document.

## Two smaller things in the same path

The constants behind the model workspace figure are bare values with no doc comment, no benchmark and no reference, so the workspace estimate cannot be checked against anything.

And a failure here is reported as a warning rather than an error. If the budget is genuinely intended to stop the work, the request should be refused with a message naming the setting to raise. As it stands the caller cannot distinguish "layout detection ran and found nothing" from "layout detection never ran".

## Reproduction

Extract a PDF of more than twelve standard-size pages with layout detection requested and OCR forced, then run the same document again with layout detection off. The two outputs are identical in nodes, bounding boxes, text, tables, metadata and confidence; they differ by one warning.

## Version

Observed on **v1.2.6**, the current release.

`crates/xberg/src/layout/engine.rs` and `crates/xberg/src/extractors/pdf/layout_runner.rs` are byte-identical at v1.2.3 and v1.2.6 — the same blob on each tag — so this has been present unchanged across at least three releases.

tobocop2:
Layout detection accumulated every page raster for a document in one retained collection and charged the running total against a single content-size budget that was never released between pages. Past about twelve standard pages the budget was exhausted, layout detection failed, extraction carried on, and the result was reported as successful with only a warning.

The failure reads like a page-size limit and is not one. The warning names the page's pixel dimensions, but a single Letter page is about 31 MB against a 100 MiB default. What exhausts the budget is the sum.

## The fix

Each batch is charged alone. The parameter that carried the document-wide total and the per-chunk refold over every retained raster are both deleted, so there is no running total to release.

Raising the limit was considered and rejected: it moves the cliff from about twelve pages to about eighty-one at 512 MiB rather than removing it, because the shape of the defect is accumulation.

## Model workspace constants

The three constants behind the workspace estimate were bare values with no derivation. They now carry one, and a test ties them to the real backend input sizes so they cannot drift from the models they describe.

## Still open

Whether a budget failure should refuse the request rather than warn is a question about a public signal, and it is not decided here. A caller today cannot distinguish "layout ran and found nothing" from "layout never ran". The recommendation is a field beside the existing gated-pages signal rather than an error, following the precedent that budget skips warn and name the page.

Closes #1721.


Goldziher:
Draft review. The per-batch accounting is the right shape and the regression test's arithmetic checks out (13 Letter pages / chunk 8: old cumulative second chunk 106,622,250 B > 104,857,600 B default; new per-chunk 73,301,350 B available → capacity 2). Before it leaves draft:

1. `CHANGELOG.md` and the comment above `detect_layout_chunk`'s `current_live_bytes` both say page count "is bounded by `security_limits.max_pages`" — that field defaults to `None` (unlimited). After this change `max_content_size` no longer bounds the layout pass's total retained rasters across a long document (`all_images.push(page.image)` keeps every chunk alive), only `max_pages` does, and it is off by default. Say that explicitly rather than implying a ceiling exists out of the box.
2. The new rationale comment at `layout_runner.rs` ~564-567 has no `~keep` marker; the uncomment pass will strip it. Siblings in the file (402, 728, 1790) carry it.
3. `layout_runner.rs`'s `current_live_bytes` fold (~568-574) is now the same computation as `engine::validate_layout_batch_peak`'s `current` fold (`engine.rs` ~499-505). Factor it into one helper in `engine.rs`.

Also noted: the `LAYOUT_MODEL_MAX_INPUT_SIDE` doc and test correctly say custom Yolox exports aren't bounded by the 1280 px ceiling, but nothing enforces it at runtime, so the workspace estimate silently under-counts for them (pre-existing gap, first written down here). And the `wip:` commit subject needs a conventional-commits rewrite before ready.

tobocop2:
## Fixed

**Duplicated fold.** The live-raster accounting is now one helper in the layout engine, called by both sites.

**Stripped comment.** The rationale comment carries `~keep`.

**Overstated bound.** The changelog and the comment now say the content-size limit bounds only the work in flight. Rasters retained across a document are bounded by the page limit, which is unset by default.

**`wip:` subject.** Rewritten. The tree hash is unchanged by the rewrite.

## Gate

Merged tree: fmt and clippy clean, 392 layout tests pass.

## Not a branch defect

A layout arm went red on a missing fixture. The corpus under `test_documents` is fetched by a script and is separate from the submodule checkout.

Ready for review.

Goldziher (approved):
All three earlier items and the commit subject are addressed at 257d02ecb6: `live_raster_bytes` is the one fold in `layout/engine.rs`, the rationale comment carries `~keep`, and the CHANGELOG/comment state that only `max_pages` (unset by default) bounds the retained rasters. The regression test fails on main with the predicted 106,622,250-byte validation error and passes here; 27 layout_runner tests and both peak tests pass, clippy `-D warnings` clean. I pushed 107ef8bcb3 to reword `SecurityLimits::max_content_size`'s doc, which still said "per document", and removed the stale Draft "Status" section from the body.

PR #1733, opened 2026-09-21
feat(ocr): let recognition concurrency follow the thread budget and free memory

### LINKED ISSUE #1727: bug: recognition concurrency is fixed at four by a compile-time constant, leaving most cores idle
## Problem

The number of concurrent Tesseract recognition operations is fixed at 4 by a compile-time constant. No configuration field and no environment variable reaches it, and the thread budget does not affect it. On a 32-core machine the recognition phase runs 4 wide and leaves 28 cores idle.

The constant is applied twice, so relaxing either site alone changes nothing: once as a semaphore acquired before the blocking call, and once as the capacity of the API pool itself.

## Measurement

A 731-page mixed document, cold recognition cache asserted empty before every run, 32 vCPU, same build and same settings apart from the constant:

    constant    wall clock                  recognition-phase runnable threads
    4           148.1 s                     4.06 mean, never above 4
    32          76.9 / 77.1 / 77.2 s        19.86 mean

**Halving the wall clock, with byte-identical output.** Both arms produced 2,091 cache entries and the same character count.

Thread sampling shows exactly 4 runnable threads for 92.3 seconds of the 148.1 second run.

## Phase breakdown

Symbolized profile of the same cold run:

    prologue   49.5 s   1 runnable    JPEG2000 28 %, resampling 13 %, text 12 %, PNG 9 %, JBIG2 5 %
    OCR        92.3 s   4 runnable    tesseract 61 %, leptonica 17 %

Once the recognition bound is lifted, the serial prologue becomes the remaining ceiling. That is image decode, and it is a separate piece of work.

## Suggested shape

Make the pool size configurable and default it to something derived from the available parallelism rather than to 4. Both call sites take it from the same place, so they cannot drift apart. A caller that wants the current behaviour can still ask for 4.

The two sites are in the OCR API pool and the Tesseract backend.

## Why this matters more than it looks

Extraction is widely reported as not scaling with the thread budget. This is why. Raising the thread budget cannot move recognition throughput, because recognition concurrency is not derived from it, and recognition is the majority of the run on a scanned document.

## Version

v1.2.6, and present on the current fix branch for #1666.

tobocop2:
Concurrent Tesseract recognition was fixed at four by a compile-time constant. Two limiters enforced it, an admission semaphore before the blocking call and the capacity of the handle pool, and each read the constant separately. No configuration reached either, so the thread budget could not move recognition throughput, and recognition is the majority of the run on a scanned document.

## The change

Both limiters now read one latched accessor, so no construction order can give them different values. It resolves to the thread budget, reduced to the number of sessions the host has memory for. A new setting gives recognition a tighter limit of its own, for a host with cores to spare but not memory; setting it to 4 keeps the previous behaviour.

## The memory bound is measured

A session owns a recognition handle, the image it reads and the intermediate page data, so the cost is per session rather than per document. Measured on an 84-page scanned document: about 68 MiB for each session added, linear, with the default resolving to 8 sessions at 714 MiB peak. The reservation is set well above that because the working set scales with the page raster, and a large page at a high scan resolution costs several times a letter-sized one. Cores the memory cannot feed do not become sessions, which is what stops a container with a wide processor quota and a narrow memory limit being killed rather than slowed.

## What the default does, and does not, change

The default inherits the existing documented ceiling of eight. On a 32-core host that takes recognition from four sessions to eight, not to thirty-two. A host that wants more sets the thread budget explicitly, which now reaches recognition for the first time; before this change, setting it did nothing at all to recognition throughput.

An explicit limit is applied as the caller set it, and is not reduced to the thread budget. The sibling thread-budget setting already wins outright over its own default ceiling, and silently lowering a number somebody set is the defect this change exists to remove. The CLI help, both guides and the struct documentation say so.

## Verification

Red on the base tip, with the assertions expressed against the base API: the handle pool admits 4 sessions on a host budgeted 8 threads. Green on the head, 41 tests passing. Formatting clean, clippy clean on the full feature set and on both per-feature legs, and no new test failures against the baseline.

## Not covered

The generated bindings and reference pages are regenerated here, so nothing is left stale.
`alef verify` checks each file's stamp rather than re running the generator, so it cannot see a
binding that was left behind; that gap is not this change's to close.

An explicit `max_concurrent_ocr` is fixed by the first extraction in the process. A later
extraction that names a different value keeps the first one and now logs a warning naming both
numbers, and the documentation says so. Making it follow the later caller means rebuilding the
admission semaphore and the handle pool while recognition sessions may be in flight, which is a
larger change than this one.

## Landing order against #1724

#1724 rewrites the same OCR batch sizer. At `32dba7ec` against its head `fb31dffe0` the two now
conflict in `crates/xberg/src/extractors/pdf/ocr/pipeline.rs` and
`crates/xberg/src/extractors/pdf/ocr/tests.rs` as well as in the changelog, so the earlier note
that only the changelog conflicts is out of date. Land #1724 first.

Resolve toward #1724, not toward this branch. #1724 makes the memory read return an optional
where absent means "could not tell" and a measured zero means a one page batch; spelling both
as zero handed the widest batch to the host with the least memory, and that is the defect it
was blocked on. This branch hands the same value back through `unwrap_or(0)`, which collapses
them again. The shared reader this branch introduces already returns the optional with the right
meaning, so the resolution is to delete the discard at the module boundary and return it
unchanged. That keeps this branch's per call freshness and its removal of the duplicate reader
together with #1724's three way meaning and its off executor read. In the test file, keep
#1724's batch width cases and this branch's deletion of the parser tests, whose parsers no
longer exist here.

Closes #1727.


Goldziher:
Draft review — not merge-ready yet:

- `resolve_batch_input_config`/`resolve_batch_base_config` (`engine/extract_impl.rs` ~1092 and ~1108) rebuild `ConcurrencyConfig { max_threads: Some(thread_budget), max_concurrent_ocr: None }` whenever the batch divides the thread budget, so batch extraction silently drops any `max_concurrent_ocr` the caller set. Carry the original through. No test covers this path with the field set.
- `read_available_memory_bytes` in `core/config/concurrency.rs` duplicates the existing `get_available_memory`/`cgroup_headroom` in `extractors/pdf/ocr/pipeline.rs`, with weaker semantics: it reads `memory.max` without subtracting current usage (the existing helper does `limit.saturating_sub(usage)`), and returns `None` on every non-Linux platform where the existing one has a `sysctl` path — so on macOS the memory cap is a no-op. Reuse or extend the existing helper; #1724 sits on the same helper, so a shared one avoids a three-way divergence.
- `ConcurrencyConfig` gains a public field and is not `#[non_exhaustive]`; any external Rust caller constructing it by struct literal breaks. Decide: mark it `#[non_exhaustive]` or call the break out for the version bump.
- No `CHANGELOG.md` entry for the new `max_concurrent_ocr` field or the changed default (fixed 4 → follows the thread budget).
- When the memory bound clamps recognition concurrency below the thread budget, nothing is logged; the sibling `resolve_thread_budget` warns once when its ceiling is hit.
- Nit: `tesseract_api_capacity()` wraps an already-latched `ocr_concurrency()` in a second `OnceLock`.
- Nit: `resolve_ocr_concurrency` reuses the name of the VLM-concurrency function removed under GH#1465 that `image_ocr.rs:384`'s doc comment still references; a more specific name avoids the collision.

No textual conflict with #1724 in `extractors/pdf/mod.rs`, but both edit the same test function within a few lines.

tobocop2:
## Fixed

**Batch extraction dropped a caller's limit.** Both batch resolvers rebuilt the concurrency config and lost the field. They carry the original through, and a test now covers that path with the field set.

**Duplicate memory reader.** Deleted rather than synced, since a second reader would become a three-way divergence with #1724. The pipeline delegates to the one helper in core, which also fixes the two defects in the copy: it subtracts current usage from the cgroup limit, and macOS is no longer a no-op.

**Silent clamp.** A memory-bound clamp below the thread budget now warns once, matching the sibling resolver.

**Both nits.** The redundant latch is gone and the resolver is renamed off the collision.

**Changelog.** Already on the branch.

## Not applied

`#[non_exhaustive]`. No config struct in this area uses it, so adding it here alone makes one struct behave differently from its siblings for external callers.

## Open

The field is not reachable from any binding. Every generated binding builds the config with defaults, so Python, Node, PHP, WASM and FFI callers cannot set it. This stays a draft until they can.

## Gate

Merged tree: fmt and clippy clean, 284 OCR tests, 118 pdf tests.

Goldziher:
Re-reviewed at c1126c9cb5. Still not merge-ready; two things need to move before the binding work.

**Compile break on the `ocr`-only legs.** `recognition_concurrency()` is gated `#[cfg(sceptre_ocr)]` (`core/config/concurrency.rs:382`). That cfg is set by `build.rs` only for the `sceptre-ocr-*` features; the Tesseract `ocr` feature does not enable any of them. `ocr/processor/api_pool.rs:23` calls it under plain `feature = "ocr"`, so `cargo clippy -p xberg --no-default-features --features "ocr,auto-rotate-tract"` (ci-rust "Clippy no-ORT tract engine") and the Windows `ocr,static-embeddings` test leg fail with an unresolved function. A `--workspace` clippy passes only because xberg-cli's defaults unify `ocr` and `sceptre-ocr` into xberg. The gate should be `#[cfg(feature = "ocr")]`.

**The shared memory reader is now latched.** `available_memory_bytes()` sits in a `OnceLock`, and `get_available_memory()` in the pipeline delegates to it. On main the batch sizer re-reads free memory per document; after this PR it is sized from whatever was free at the first extraction, for the life of the process. That is a regression for the server, and it changes #1724's bound after the merge (its `spawn_blocking` ends up wrapping an atomic load, and the one cold read moves onto the executor inside `init_thread_pools`). Keep the latch on `ACTIVE_OCR_CONCURRENCY` only and let the reader itself stay fresh.

Smaller:
- The body says recognition never exceeds the thread budget and the CLI help says "below the thread budget", but an explicit `max_concurrent_ocr` returns `requested.max(1)` with no budget clamp (`concurrency.rs:362`, asserted at `:1077`). Pick one and align the docs.
- The batch-path test covers `resolve_batch_base_config` only; `resolve_batch_input_config` still has no test with the field set.
- `test_ocr_concurrency_is_not_pinned_to_four_on_a_many_core_host` reads the real host's free memory; on a >4-core runner with under ~2.5 GiB free it fails. Skip when the memory bound binds, or inject.
- `ConcurrencyConfig` gains a public field without `non_exhaustive`; the struct-literal break should at least be named in the CHANGELOG entry.

Earlier items verified as addressed: batch resolvers carry the field (`extract_impl.rs:1098`, `:1115`), the duplicate reader is deleted and the core one subtracts usage and has a macOS path, warn-once on the memory clamp, the redundant `OnceLock` in `api_pool.rs` is gone, the resolver is renamed. Merge check against #1724: only `CHANGELOG.md` conflicts; the code files auto-merge and nothing dangles.

tobocop2:
## Measured end to end

A 731-page mixed native-and-scanned PDF, per-page OCR gate, cold, `--max-threads 32`. One dedicated
idle box, three repetitions per arm, median reported, result cache cleared and asserted empty before
every run. Spread 0.10 percent and 0.71 percent.

| arm | wall clock | sessions | CPU peak | peak RSS |
|---|---|---|---|---|
| v1.2.6 as released | 82.595 s | 4 | 404 % | 1.8 GB |
| v1.2.6 plus the open fixes | 39.853 s | 32 | 3196 % | 5.5 GB |

**2.07 times faster, and the extracted content is byte-identical.** The same SHA-256 over every run
of both arms, and the same 79 processing warnings. The time did not come from doing less work.

Peak memory rises 3.2 times, because 32 pages are in flight instead of 4. That is the sizing
consequence, and it is why the resolved session count is bounded by host memory.

Two things this table does not say.

It was taken with the thread budget set explicitly. A default deployment resolves to eight sessions,
not thirty-two, so it sees a smaller gain.

**It needs #1724 as well as this change, and on the gate path this change alone does almost
nothing.** Measured on this branch with only the session limit reverted: 85.1 s against 87.0 s, and
the run stays at 4 concurrent sessions and 1.8 GB whichever arm runs. The gate feeds recognition
through the per-page batch, and until #1724 widens that batch it holds recognition at four whatever
this change resolves the limit to. Lifting the limit is necessary and not sufficient.

Where this change is measurable on its own is a forced-OCR run, which does not go through that
batch: 13.6 s to 6.5 s on a 22-page document, 4 concurrent sessions to 22. That box was shared at
the time, so treat 2.08x as a floor rather than a figure.

tobocop2:
## Five things are being fixed before this leaves draft

**A second extraction ignores the setting.** Measured: two extractions in one process, the second asking for 16, both resolve to 3. The value latches on the first extraction and later callers are discarded. The field docs, both guides and the changelog all say it is applied as given. Silently ignoring a setting is the defect this change exists to remove, so it cannot ship that way.

**Nothing guards the un-latching of the memory reader.** Re-latching it leaves the whole suite green with the same failing names. The regression you blocked on is fixed and nothing stops it returning. A test is being added that goes red if the latch comes back.

**The conflict with #1724 is semantic, not textual.** Your 18:40 note said only the changelog conflicts. At the current heads `pipeline.rs` and `tests.rs` conflict too, and the resolution matters: this branch rewrites the memory read to a plain integer with a zero fallback, which collapses "unknown" and "no headroom" back together. That is the defect #1724 was blocked on. The resolution keeps #1724 optional return and deletes only the discard at the module boundary. A full integration build verified it.

**The version annotation names a version that has not shipped.** `alef verify` differs from the base by exactly that one line.

**The "Not covered" section is stale.** It says the bindings still need regenerating. This diff regenerates all of them.

Goldziher (approved):
Both earlier blockers are fixed at 5835dccaf4 (`recognition_concurrency` is gated on `feature = "ocr"`; the memory reader no longer latches) and the field is now reachable from every binding, which was the reason for the draft. I pushed three commits: `then_some` for the macOS-only `clippy::unnecessary_lazy_evaluations` that fails the `macos-latest` workspace clippy leg; the alef regen of 21 doc-only binding files the branch's later doc edits had left stale (`alef verify` does not catch that — it hashes inputs); and the merge of main, which keeps #1724's `Option`-returning batch-sizer reader delegating to the single core reader and adds `max_concurrent_ocr: None` to the two new route-test literals. On the merged tree: clippy `-D warnings` clean on the `pdf,ocr` and `ocr,auto-rotate-tract` legs, 258 concurrency/OCR/extract_impl/batch tests pass, `alef verify` clean. Noted, not blocking: a caller driving `TesseractBackend` directly before any `extract_*` builds the pool at the automatic limit, and a later explicit `max_concurrent_ocr` then installs without the latch warning.

Issue #1732, opened 2026-09-21
bug: enabling OCR turns on a single-threaded embedded-image pass that decodes twice and discards its output


tobocop2:
## Problem

Configuring OCR switches on whole-document embedded-image extraction, and that extraction is a single-threaded loop over every page. On a 731-page document it costs about 17 seconds on one core, decodes images a second time that the native path has already decoded, and re-encodes every non-JPEG image to PNG only to discard the result.

The extraction config treats "OCR is configured" as "the caller needs image data", so the image pass is enabled by default whenever OCR is on. The pass itself is a plain `for page_idx in 0..page_count` in the native image extractor.

## Measurement

731-page mixed document, 32 cores, OCR and extraction caches cleared and confirmed empty before every run, sampled for runnable threads:

    arm                          wall clock    prologue    runnable threads in prologue
    OCR on, 32 threads             168.0 s       42 s                  1
    OCR on, 8 threads              168.9 s       42 s                  1
    OCR on, 4 threads              168.5 s       42 s                  1
    OCR off                         24.8 s       25 s                  1

The thread budget makes no difference at all. The prologue is 25 seconds of native extraction plus about 17 seconds that exists only because OCR is configured.

Attributed by profile over the whole run: JPEG2000 decode 14.1 s, unattributed glue 13.7 s, resample and colour 4.2 s, JBIG2 3.0 s, PNG encode 2.6 s, PDF text 2.5 s. The OCR-off control does the same native work with JPEG2000 at 5.7 s, JBIG2 at 1.5 s, and no PNG encode at all.

## Three separate costs

**Serial by construction, not by necessity.** The document handle is documented as safe to share across threads, with its interior mutability behind a mutex. Nothing prevents the loop running wide.

**The same images are decoded twice.** The native path already decodes those objects; the image pass decodes them again.

**The PNG encode is thrown away.** The result is re-encoded and the returned image list is empty.

## Related work that stopped short

An earlier change parallelised one page-rendering function and closed the throughput issue it was filed against. Its sibling batch function stayed serial, and this image loop was never in scope. The guard test added alongside it covers only the function that was fixed.

## Do not "fix" this by turning the pass off

Disabling the embedded-image route makes the same document take 347 seconds, because the pages then go through a far more expensive path. The image route is the cheap route. It needs to be parallel and to stop duplicating work, not to be skipped.

## What this is worth

[...]

PR #1734, opened 2026-09-21
fix(pdf): extract embedded images in parallel across the thread budget

tobocop2:
Enabling OCR turns on a pass that extracts every embedded image in the document, and that pass ran on one thread. On a 731-page file with 2,881 images it took 20 seconds at exactly one runnable thread, whatever the thread budget was set to.

## The change

The pass now runs across the thread budget.

## Memory is not unchanged

Up to a budget's worth of raw pixel buffers and re-encodes are live at once instead of one page's, and unlike the OCR render route this pass has no memory-budgeted cap. The ceiling is the thread budget, which is the lower of the core count and eight until it is raised explicitly.

Measured on the same 731-page file, 2,881 images, with the pass's own peak isolated:

| arm | pool | pass | peak |
|---|---|---|---|
| sequential | | 624.5 s | 1,133 MB |
| parallel | 1 | 613.4 s | 1,136 MB |
| parallel | 8 | 94.0 s | 1,463 MB |
| parallel | 32 | 55.7 s | 2,186 MB |

About 35 to 48 MB per concurrent page. The pool-of-one arm is a control: it is within 0.3 percent of sequential, so the rise comes from concurrency rather than from the new collection shape. Returned bytes are identical across all four.

These were taken on a shared box in a debug build, so read them as bounds rather than figures.

## Measured

731 pages, 2,881 images, the image cache cleared and confirmed empty before all seven runs, thread budget 32:

    arm                      pass duration              runnable threads, peak
    serial                   20.098 / 20.037 / 20.149 s        1
    parallel                  2.211 / 2.153 / 2.271 s         32
    parallel, one thread     19.760 s                          1

**9.1 times faster, 17.9 seconds removed.** The single-thread arm is a control: it shows the rewrite adds no per-page overhead, so the gain is parallelism rather than a change in the work.

End to end on the same file, image extraction on, the whole extraction goes from 57.314 s to 41.308 s. That is 16 seconds of the 17.9 showing up in wall clock, measured separately on the same box.

## Output is unchanged

A hash over page number, image index, format, dimensions, byte length and the image bytes themselves, in returned order, is identical across all seven runs of both binaries. 2,881 images, no warnings, in every run. Returned order matters as much as content here, so it is part of what the hash covers.

## Verification

The new test asserts the pass uses more than one thread, and fails on a tree with the parallel iterator swapped back for a sequential one: "got 1 distinct thread(s)". That failure is a runtime one, not a compile error, so the test genuinely exercises the behaviour rather than the type signature.

## Not in this change

Two further costs in the same pass are deferred: the images are decoded a second time here after the native path has already decoded them, and every non-JPEG image is re-encoded and the result discarded. Both change the returned bytes, and byte-identity was the constraint on this change.

The sibling batched force-OCR route is also still serial. Its batch size is already capped against a memory budget, so widening it multiplies peak raster retention and needs its own measurement. It belongs on its own branch.

Refs #1732. This fixes the first of the three costs the issue lists. The duplicate decode and the discarded re-encode stay open, for the reasons under "Not in this change".


tobocop2:
## Measured, end to end

A 731-page PDF, image extraction on, thread budget 32. One box, one session, two repetitions per binary, cache cleared and asserted empty before each run.

| whole extraction | wall clock | threads | CPU peak |
|---|---|---|---|
| before | 57.314 s | 1 | 106 % |
| after | 41.308 s | 32 | 3295 % |

These are whole-extraction wall clock, not the pass in isolation. The pass itself is in the description: 20.1 s to 2.2 s. The 16 seconds that reach wall clock here are the pass's 17.9 second saving less the prologue.

## Scope

This path runs only when image extraction is on, which is off by default. A run using the defaults shows no difference.

Goldziher (changes requested):
The parallel pass itself checks out: page order and `image_index` assignment match the sequential loop, per-image skip semantics are unchanged, `PdfDocument.load_lock` only covers cold object loads so decode and re-encode genuinely run wide, and the guard test fails with "got 1 distinct thread(s)" when the parallel iterator is swapped back for a sequential map (checked locally; clippy `-D warnings` and the 2 new tests clean under `--features pdf,tokio-runtime`). Three things before merge:

1. **No `CHANGELOG.md` entry.** This is a user-visible performance change on every OCR-configured PDF extraction, the same class as the GH#1666 render-pass entry under 1.2.6. Add a `### Changed` bullet under `[Unreleased]` with the measured numbers and the memory bound below.

2. **"Nothing else about it changes" is not true for memory.** Up to `budget` pages' raw pixel buffers and PNG re-encodes are now live at once instead of one page's (`images.rs:377-383`); the commit message says so, the PR body does not, and there is no peak-RSS figure for the 731-page run. Report peak RSS before/after, or state the bound explicitly — unlike the OCR render route this pass has no memory-budgeted batch cap.

3. **`Closes #1732` should be `Refs #1732`.** The issue enumerates three costs and says the pass must "stop duplicating work"; this PR fixes the serial pass and defers the duplicate decode and the discarded re-encode by its own "Not in this change" section. Leave the issue open or re-scope it first.

Minor:
- `PAGE_CALL_THREAD_IDS` is process-global and `#[serial]` only excludes other serial tests, so any concurrently running test that reaches `extract_page_images` (the fixture-backed tests in `images.rs::tests`, OCR-configured PDF extractions) can inflate the set and let a sequential regression pass. Naming the pool threads and counting only those, or recording `page_idx` with the id, closes that.
- The numbers disagree across the commit message (17 core-seconds), the body (20.1 s → 2.2 s for the pass) and the follow-up table (57.3 s → 41.3 s, labelled as the pass). Label which is pass time and which is whole-extraction wall clock.
- `images.rs:322-325`, `:331-333`, `:465` carry rationale without `~keep`.

Goldziher (approved):
All three requested items are in place and verified: changelog entry with the memory bound, the peak-RSS table in the body, and `Refs #1732` with no closing reference. The thread guard now scopes to its own named pool and still fails ("got 1 of the pool's threads") when the parallel iterator is made sequential; 19 `pdf::native::images` tests pass on the branch merged with current main, clippy `-D warnings` clean. I pushed two small commits (changelog wording: each pool thread holds a page's buffers, not one buffer; a `~keep` on a test rationale comment) plus the merge of main.

PR #1736, opened 2026-09-21
fix(pdf): reject a row-padded pixel buffer instead of panicking

### LINKED ISSUE #1735: bug: a row-padded image buffer panics in the PNG encoder instead of returning an error
## Problem

A PDF whose embedded image carries a row-padded pixel buffer makes extraction **panic** instead of
returning an error. The panic is inside the `image` crate's PNG encoder, reached from
`raw_pixels_to_png` in `crates/xberg/src/pdf/native/images.rs`.

    thread 'probe' panicked at image-0.25.10/src/io/encoder.rs:115:17:
    assertion `left == right` failed: Invalid buffer length: expected 182055 got 182320
      for 229x265 image

## Cause

`229 x 265 x 3` is 182,055 bytes. The decoder produced 182,320 — exactly one extra byte per row, which
is what a 4-byte-aligned row stride gives. The buffer is not corrupt; it is padded.

`ImageBuffer::from_raw` only rejects a buffer that is too **small**. A buffer that is too **large** is
accepted and the extra bytes are kept silently. The mismatch then surfaces much later, inside
`DynamicImage::write_to`, as the encoder's own exact-size assertion — a panic rather than a recoverable
error.

## This hazard is already known and already fixed elsewhere in the tree

`to_png_bytes` in `crates/xberg-native-pdf/src/extractors/images.rs` carries this exact guard, with a
comment describing the same failure almost word for word: `from_raw` accepts a buffer at least as long
as the image needs, so it rejects the mismatch there rather than handing the encoder a bad buffer.

That comment predates this report. The guard was simply never applied to this second call site.

## Impact

A panic is not a failed extraction. It takes the process down rather than returning something the
caller can act on, so a caller sees a dropped job instead of a message naming what went wrong. All
three pixel formats in this function are affected, because all three build an image from a raw buffer
the same way.

## Reproduction

Extract a PDF containing an embedded raw image whose decoded row stride exceeds width times channels.
Measured on two documents in a private corpus, which turned out to carry the same image under two
names — one image, one defect. Twenty-four other documents in the same corpus extract normally.

The property is the size mismatch, not the document: any row-padded buffer reproduces it.

## Fix

Validate the exact buffer length before construction and return the recoverable error this function
already returns for an undersized buffer. A pull request follows.

tobocop2:
Closes #1735

## Problem

A PDF whose embedded image carries a row-padded pixel buffer panics instead of returning an error:

    assertion `left == right` failed: Invalid buffer length: expected 182055 got 182320
      for 229x265 image

`229 x 265 x 3` is 182,055 bytes. The decoder produced 182,320 — one extra byte per row, which is what
a 4-byte-aligned row stride gives. The buffer is padded, not corrupt.

`ImageBuffer::from_raw` only rejects a buffer that is too small. One that is too large is accepted with
the extra bytes kept, and the mismatch surfaces much later inside the PNG encoder as its own assertion.
A panic takes the process down rather than returning something the caller can act on.

## Fix

Check the exact length before construction and return the recoverable error this function already
returns for an undersized buffer. Applied to all three pixel formats, because all three build an image
from a raw buffer the same way.

`to_png_bytes` in `xberg-native-pdf` already carries this guard, with a comment describing the same
hazard. It was never applied to this second call site.

## Tests

Three regressions, one per pixel format, each building a buffer with exactly one padding byte per row.
They are driven by the size property rather than by any document, so they do not depend on a corpus.

## Verified

`cargo test --release --lib` on this branch, filtered to the new cases:

    test_raw_pixels_to_png_row_padded_rgb_returns_error_not_panic ... ok
    test_raw_pixels_to_png_row_padded_grayscale_returns_error_not_panic ... ok
    test_raw_pixels_to_png_row_padded_cmyk_returns_error_not_panic ... ok

    test result: ok. 17 passed; 0 failed

The before-state was reproduced directly: the panic, its backtrace and the byte counts above all come
from a real run rather than from reading.

Six other tests in this module fail in a bare checkout for a reason unrelated to this change — they
need `test_documents/` fixtures that are not present there, and they fail identically without this
diff.

## Scope

The two documents that exposed this turned out to carry the same image under two names — one image,
one defect. Twenty-four others in the same corpus extract normally.


Goldziher (approved):
Reproduced the 229×265 / 182,320-byte panic at `image-0.25.10 encoder.rs:115` with the guard removed and confirmed the branch turns it into `ExtractionFailed` with the image skipped and a `ProcessingWarning` recorded; the three new tests panic without the fix. Optional follow-up: the three `from_raw(...).ok_or_else` branches are now unreachable, and the tests could assert the byte counts in the message rather than `is_err()`.

PR #1737, opened 2026-09-21
fix(pdf): make a page's text independent of the order the pages are read

### LINKED ISSUE #1725: bug: native page text depends on font load order, which blocks any page-parallel parse

tobocop2:
Page text depended on the order pages were read, so any page-parallel parse returned different text on different runs.

## The issue names the wrong cause

Read single-threaded in different orders, 396 of 397 corpus documents returned identical text. Load order alone is almost never the problem. Read page-parallel, 26 of 397 diverged, each with object-parse warnings that never appear sequentially. That is a data race. The font behaviour the issue describes is real and accounts for one of the 26.

## The race

The document kept one seekable reader behind a mutex. An object's header and its body were read under separate guards, so another thread's seek in between fed the body from a different object.

That same split lock had already been patched twice in that file. The third time, the reader is deleted rather than guarded again: reads address the bytes by absolute offset with a private cursor per call. The serialising lock and a full file copy per open go with it.

## The font half

The font-set caches stored the whole extractor, so an entry belonged to whichever page reached a font dictionary first. Each cache now stores that dictionary's own fonts, taken before character-map donation.

## Evidence

Base fails at repetition 3: two digests, 19,656 bytes against 16,309. Head passes twice.

On a 731-page document, base produces four distinct parallel digests between 943,045 and 944,130 bytes. Head produces one, identical to the sequential result, with no change in wall clock.

Across the corpus the output changes on 1 document of 397, by 1 byte.

## Gate

289 test binaries, 6,188 passed, 0 failed. Formatting and clippy clean.

## Not covered

The character-map merge path has a remaining gap of the same shape, not addressed here.

The test for the font half is statistical rather than deterministic: it detects divergence by repetition. On the base commit both tests caught the race early, within the first two repetitions against budgets of 80 and 60, so a passing run is not a lucky interleaving.

Main is merged in and the gate has been re-run on the merged tree.

Closes #1725.


Goldziher (approved):
Verified the race removal (a private cursor per object load; `load_lock` and the whole-file copy gone) and the font-set caches now storing each dictionary's own pre-donation fonts. Full `xberg-native-pdf` suite 6189/0 and `xberg` `pdf::native` 251/0, clippy clean. I pushed one deterministic single-threaded forward/backward test that fails on main (two pages with inline `/Font` dicts sharing a resource name decode `["a","a"]` there, `["a","b"]` here) and widened the changelog entry to name that sequential case, plus the merge of main.

Issue #1701, opened 2026-09-20
bug: candle-deepseek-ocr repeats until its token limit at F32


tobocop2:
### Deployment

Other

### Version

Branch `fix/1674-deepseek-token-cap`, measured at `82d3a42159` and `7eaf9fb975`. Also reproduces on the same branch before the shared selector landed, so the defect is not introduced by it.

### Description

At `DType::F32`, `candle-deepseek-ocr` does not terminate on its own. Its greedy decoder enters a repeating state and emits the same short window until it reaches the token limit. The other two candle VLM backends on the same pages, in the same build, on the same machine, do not.

Measured on one A100-SXM4-40GB, three 150 dpi page images of one document, F32, one run each:

| page | characters | ms | distinct 24-character windows / total |
|---|---|---|---|
| p2 | 17870 | 115007 | 0.015 |
| p3 | 5637 | 35871 | 0.346 |
| p4 | 22274 | 114555 | 0.063 |

A ratio of 0.015 means 98.5 percent of the output is the same window repeating. For comparison, on the same three pages GLM-OCR returns 2440, 1490 and 2392 characters at ratios of 0.952, 0.930 and 0.989, and PaddleOCR-VL returns 2110, 2414 and 2731 at 0.990, 0.984 and 0.973.

Two things this is NOT:

- It is not the token limit. That limit was a hard-coded 128 and is now configurable, which is what made the looping visible: at 128 the output was truncated at roughly 400 to 650 characters before the repetition could be seen.
- It is not the absence of a repetition penalty. The backend now applies one. A presence-style penalty, which suppresses a token once however often it has appeared, does not break the state.

A frequency-scaled penalty, which suppresses a token once per occurrence, was measured as a candidate and rejected: p2 recovered to 1713 characters at 0.911, but p3 went from 5637 to 23590 characters and its ratio fell from 0.346 to 0.117, and p4 did not recover. It trades one absorbing state for another and is worse on two pages of three.

Note the dtype. At BF16 the same pages give 1797 characters at 0.950 on p2, so the backend looks healthy there. F32 is the dtype every shipping backend passes, so the BF16 figure does not describe production.

### Steps to reproduce

1. Build with `--features candle-deepseek-ocr,cuda`.
2. Construct the engine with `DType::F32` and a token limit at its current default.
3. Run it on a 150 dpi page image of ordinary printed text.
4. Measure the ratio of distinct 24-character windows to total windows in the output, and the wall clock.

### Logs and configuration

[...]

PR #1739, opened 2026-09-21
perf(candle-ocr): remove two per-element device loops from DeepSeek-OCR

### LINKED ISSUE #1719: bug: two more per-element device loops remain in the DeepSeek-OCR vision path
### Deployment

Other

### Version

v1.2.6. Found by the independent review of #1715, which fixed two instances of this pattern.

### Description

Two more per-element device operations remain in the DeepSeek-OCR vision path, both on the per-page hot path. They are the same class as the two removed in #1715, where the first cost six times throughput and the second a further 1.08x.

**`interpolate_linear_1d`** runs `target_size` iterations of four device operations each, then concatenates `target_size` separate tensors. It is called four lines above the function #1715 fixed, on the identical chain: twice per `add_decomposed_rel_pos`, per attention layer, per crop. It is guarded by a dimension check, so it fires only when the tile size differs from the pretrained table, which makes it configuration-dependent rather than always-on.

This one is a kernel-launch storm rather than a device-to-host synchronisation storm, so the fix and the expected return differ from #1715. Worth measuring before assuming it behaves the same way.

**`masked_scatter_dim0`** gathers rows one at a time over the sequence length, then stacks thousands of tensors. It runs once per page during prefill. The mask is already read in bulk; the rows are not. `index_select` followed by `where_cond` replaces the loop.

Two sites were checked and are NOT this pattern, recorded so nobody re-checks them. `interpolate_bicubic` is a nested per-pixel loop but runs at position-embedding resize rather than per page, so it is cold. The seven `to_scalar` calls in the PaddleOCR-VL path are all a single `max_all().to_scalar()` per image, which is one transfer, not a loop.

### Steps to reproduce

Profile a page through `candle-deepseek-ocr` on CUDA and attribute time to kernel launches in the SAM attention path. The guarded function requires a tile size that differs from the pretrained relative-position table, so the reproduction must use a configuration that triggers the resize.

### Logs and configuration

A100-SXM4-40GB, CUDA 12.9, v1.2.6. Sites identified by reading the crate during the review of #1715; the two removed there measured 62.71 to 8.09 seconds a page in total. These two are not yet measured, and this issue does not claim a figure for them.

tobocop2:
Two per-element device loops remained in the DeepSeek-OCR vision path: the rel-pos resize spent four device operations per output position plus a concatenation of one tensor per position, and the embedding scatter narrowed one row per sequence position and stacked them.

## The loop is deleted, not vectorised

Candle owns this. `upsample_bilinear2d` is the identical half-pixel formula on CPU and CUDA, so the hand-rolled resize goes rather than being made wider. The same resize was already patched once before under #1701 and #1702, which makes this the second defect in one hand-rolled primitive.

An in-crate sibling was rejected: it round-trips through the host.

## It is not a speedup, and the issue's premise is wrong

Measured on one L4, BF16, five repetitions per arm: 47.610 s before, 47.546 s after. The base spread alone is 0.147 s, so the difference is inside the noise.

The reason is that the resize fires 8 times per page, not once per layer per crop as the issue assumes. That is roughly 5,000 kernel launches out of 47.6 seconds. The premise is off by about two orders of magnitude.

Peak GPU memory moves from 8,818 MiB to 8,722 MiB of 23,034 available.

## Output is unchanged

Every repetition of both arms returned 2,052 characters with the same hash.

That is a measured result on one document, not a bit-exactness guarantee. The sampling schedule is
identical, verified against the library's own CPU and CUDA implementations, but the arithmetic is
not: the library accumulates in double precision and rounds once, where the deleted loop
accumulated in the tensor's own type. An oracle comparing the two over 27 cases, including the real
model shapes and strided inputs, puts the worst difference at 7.6e-6. The new path is the more
accurate of the two.

## Not covered

The workspace lint could not run on the measuring host: its system image carries a libheif too old for the `full` feature set. The per-crate arms ran clean, and a planted unused variable drove the lint to a non-zero exit, so the arms that did run have proven power.

Closes #1719.


Goldziher (approved):
Traced candle 0.11's CPU/CUDA/Metal bilinear kernels against the deleted loop: same half-pixel schedule (`src = max(scale*(i+0.5)-0.5, 0)`, floor, `min(w0+1, in-1)`, clamped weight), so with a one-row image it reduces to the old linear blend; the scatter is an exact gather. Measured worst |diff| on (1,64,127)→79 random data: 1.19e-7. I pushed two commits: equivalence tests that pin both vectorised paths to the deleted loops (verbatim reference impls; planting align_corners or swapping the where_cond args fails them), and a CHANGELOG correction (only the four global-attention layers hold the 127-position table). Retitled to match the commit: it removes two loops, not one.

PR #1741, opened 2026-09-21
fix(ocr): charge each page against the image limit, not the whole batch

### LINKED ISSUE #1731: bug: the OCR fallback charges a whole batch against a per-image size limit, so pages are silently skipped
## Problem

The OCR fallback path charges the total size of a whole batch of rendered pages against a limit that is documented and named as a per-image bound. An eight-page batch sums to 170,397,152 bytes against a 104,857,600 byte limit, so the batch is refused and both OCR backends are skipped.

This is the same shape as #1721, in a different file, and it becomes reachable once #1721 stops the earlier accumulation from failing first.

## Measurement

Extracting a 22-page document with OCR forced and layout detection requested, on a build carrying the #1721 fix:

    arm                     wall clock    characters
    OCR, layout off           12.88 s        41,568
    OCR, layout on            14.68 s        13,000

The second run returns less than a third of the text. The pages whose batch exceeds the limit produce nothing, and the result is still reported as successful.

The exact figure decomposes with no remainder: eight rendered pages summed against a limit that applies to one image.

## Why the limit is the wrong one

The bound exists to stop a single oversized image being decoded. Summing a batch against it makes the effective per-page allowance a function of how many pages happen to be batched together, which is not a property of any page. A caller cannot predict it and cannot configure around it.

## Suggested shape

Charge each image against the per-image limit, as its name says, or introduce a separate batch bound that is explicitly a batch bound. Releasing each page once consumed would also work; what must not survive is a running total measured against a per-item ceiling.

## Scope note

This was found while verifying the fix for #1721 and is not fixed by it. Any measurement of layout detection on a forced-OCR path is unreliable until this lands, because the text those pages should contribute is silently missing.

## Version

Present on the current default branch and on the #1721 branch.

tobocop2:
Layout detection renders its own pages and passes them in. On that route the batch width comes from the thread budget, and the whole slice was charged once against a limit whose own error text calls it a per-image bound. Past a handful of pages the check trips, and the automatic OCR fallback turns the resulting error into native text plus a warning. The run reports success and every OCR'd page is missing.

## Why layout off worked and layout on did not

The two routes differ. The route that renders internally never had the batch-wide check; only the pre-rendered route did. That is the split the issue's own table shows.

## The fix

Each page is charged on its own, matching the shape #1724 applies to the sibling route. No new helper: the same pattern, so the two routes cannot drift apart again.

## Evidence

Base fails with `Image dimensions 1275x1650 require 170397152 live image-processing bytes, exceeding the configured maximum (104857600 bytes)`, which is the figure in the issue.

Red on the base, 1 failed. Green on the head, 1 passed.

The test pins the thread budget rather than inheriting the core count. Inheriting it is why this looks fixed on a small host: at four cores the batch is four pages and fits under the limit.

## Not covered

Running the whole library target in one process fails intermittently in the pipeline tests, a different test each time, each passing in isolation, none touched by this change.

That is not this branch's. It reproduces on main with this branch absent, and again with this branch's own new test deleted, naming a third distinct test. An inert probe that only lengthens the run changes the outcome, and one test thread suppresses it. Filed as #1749.

An earlier draft of this section said CI runs one process per case under a different harness. That was wrong: there is no such harness in this repository, and CI runs the whole lib target in one process, which is the shape that floats.

A second site sums a batch against the same ceiling inside the function #1724 rewrites. Left alone deliberately rather than resolved twice.

Closes #1731.


Goldziher (approved):
Per-page charge on the pre-rendered route is correct and the regression test fails on main with the issue's exact 170,397,152-byte figure; it merges cleanly with #1724 and the two routes' tests pass together on the merged tree (6/6). Nothing in the body is open, so I marked it ready and pushed the merge of main.

PR #1743, opened 2026-09-21
perf(pdf): spread scan detection and provenance routing across the thread budget

### LINKED ISSUE #1723: bug: native extraction is serial, and the cost is in scan detection and provenance routing rather than the text loop

tobocop2:
Native text extraction ran on one core. On a 731-page mixed document that is 24.6 seconds at 100 percent of a single core on a 32-core machine, and once the OCR route stops clamping its batch width it becomes the floor on the whole extraction.

## The cost is not where the issue first said

The issue's own comment corrects it: the per-page text loop is 2.7 seconds. Scan detection and provenance routing are 14.2. This change parallelises those two loops and leaves the text loop alone, because the correcting comment rejects parallelising it.

## No new primitive

Rayon is already the page-parallel mechanism at three other sites in this repository, and the document type already carries the thread-safety assertion for it. One private helper in the scan-detection module maps the page range with the same platform gating and the same failure semantics as its siblings.

## Measured

| document | before | after | peak CPU | runnable threads |
|---|---|---|---|---|
| 676 pages | 24.109 s | 21.919 s | 130 % to 2630 % | 1 to 14 |
| 4,778 pages | 66.916 s | 63.311 s | 120 % to 1230 % | |

A single-thread arm is the control: 24.078 s against 24.210 s, so the change costs nothing where it cannot help.

Output is identical. One normalised digest across all eight runs of the first document and all six of the second, single-threaded runs included.

Those two documents are almost entirely born-digital, so nearly every page short-circuits the raster pass and only the provenance pass has work. The document the issue describes is mixed native and scanned, and it was measured separately.

## On the document the issue describes

731 pages, mixed, with the gate routing 127 pages to OCR. Dedicated idle box, three repetitions per cell, median reported, cache cleared and asserted empty before every run.

| run | before | after | max threads | CPU peak |
|---|---|---|---|---|
| no OCR | 24.638 s | 12.324 s | 1 to 32 | 106 % to 3194 % |
| the per-page gate, cold | 83.104 s | 69.384 s | 4 to 32 | 404 % to 3195 % |

**Native extraction halves.** Character counts are identical in both cells, 959,352 and 1,124,154.

The absolute saving is the number that transfers: about 12.3 seconds of serial work removed, which appears in both cells. The gate cell shows a smaller percentage because that same saving is diluted by recognition time, not because the change does less there.

One caveat on the gate row. That arm reaches only four concurrent recognition sessions, because #1733 is not an ancestor of this branch. On a tree carrying #1733 the gate baseline is far lower and the percentage will not transfer. The no-OCR row is unaffected by that and lands within 4.3 percent of this round's independent reference for it.

## Landing order

This is stacked on #1737 and targets that branch, so the diff here is this change alone. It cannot merge before #1737: the provenance pass reads page text, and without that fix a concurrent read is order-dependent. GitHub retargets this to main when #1737 merges.

## Gate

The measurements above were taken on the pre-rebase tree. The rebase onto #1737's current head is two files and conflict-free, but the gate has not been re-run on it. That happens before this leaves draft.

## Not in this change

The provenance pass re-reads every page's text the main pass already extracted. Deleting that second read would beat parallelising it, because it removes the work rather than distributing it. That needs the extractor to retain spans into metadata extraction, which is a larger change than this one. Filed as #1744.

Table extraction, 6.8 seconds, is out of scope.

Closes #1723.


Goldziher (approved):
The parallel pass checks out: bounded by the global pool `init_thread_pools` builds, positional output via the indexed range iterator, infallible closures with panics caught inside `guard_native_panic` as before, wasm32 fallback. Two test defects fixed on the branch: `both_page_passes_match_a_page_by_page_run` ran on `docling.pdf`, where every page scores 0.0 and nothing is fabricated, so it passed on any permutation (reversing `map_pages`' output left it green) — it now uses `non_ascii_text.pdf` (heterogeneous scores, 15 fabricated pages), asserts that non-degeneracy, and opens a fresh handle for the parallel pass; `map_pages_dispatches_pages_across_the_pool` failed 22 of 60 runs under a 12-core load because no page did any work — each page now holds its thread briefly, 60/60 under load. The changelog entry drops the peak-CPU percentages since the body and the entry quoted different figures for the same runs; the wall-clock numbers agree and stay. Retargeted to main after #1737 landed (GitHub closed it when the base branch was deleted). 21/21 scan_detect tests on the merged tree, clippy `-D warnings` clean.

